PR #2, opened 2023-09-20
feat: adds example structure for api intergration departure

mikaelbr:
Adding fundamental structure for how to integrate and get server data for pages and starting to establish overall architecture for modules.

Open for feedback on this. Somethings might seem more advanced types and structure, but trying to clean up consumption API for pages

Untitled-2023-02-06-1529

## For later PRs

- [ ] Add documentation in text form that explains how it all is connected.
- [ ] Add example of client side APIs

### DIFF
--- a/src/modules/api-client/utils.ts
+++ b/src/modules/api-client/utils.ts
@@ -111,7 +111,7 @@ export function resT(
 export function resD(
   res: NextApiResponse,
   code: number,
-  data: ServerErrormessage
+  data: ServerErrorMessage
 ) {
   return res.status(code).json(data);
 }
@@ -181,7 +181,7 @@ export async function throwErrorFromResponse(result: Response) {
   }
 }
 
-function mapServerToMessage(e: any): ServerErrormessage {
+function mapServerToMessage(e: any): ServerErrorMessage {
   if (typeof e === "string") {
     return { message: translation(e, e, e) };
   }
@@ -240,12 +240,12 @@ function isInternalUpstreamServerError(
 }
 
 export class ApplicationError extends Error {
-  data!: ServerErrormessage;
+  data!: ServerErrorMessage;
   status: number;
   upstreamResponse?: Response | NextApiResponse<any>;
 
   constructor(
-    error: ServerErrormessage,
+    error: ServerErrorMessage,
     status: number = 500,
     upstreamResponse?: Response | NextApiResponse<any>
   ) {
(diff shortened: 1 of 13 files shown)


src/modules/api-client/utils.ts
```diff
@@ -0,0 +1,257 @@
...
+
+export function resD(
+  res: NextApiResponse,
+  code: number,
+  data: ServerErrormessage
+) {
+  return res.status(code).json(data);
+}
+
+export type ServerErrormessage = {
```
adriansberg, line 119:
```suggestion
export type ServerErrorMessage = {
```

src/modules/api-client/utils.ts
```diff
@@ -0,0 +1,257 @@
...
+}
+
+function isInternalUpstreamServerError(
+  e: any
+): e is InternalUpstreamServerError {
+  return "errorCode" in e && "shortNorwegian" in e;
+}
+
+export class ApplicationError extends Error {
+  data!: ServerErrormessage;
```
adriansberg, line 243: Is the exclamation mark needed?

src/modules/api-client/utils.ts
```diff
@@ -0,0 +1,257 @@
...
+  ) {
+    const baseUrl = externalHttpUrls[baseUrlKey];
+    const actualUrl = `${baseUrl}${url}`;
+
+    try {
+      const data = await fetch(actualUrl, {
+        ...init,
+        headers: {
+          ...init?.headers,
+          "ET-Client-Name": "FOO",
```
adriansberg, line 87: Should the client name be updated at a later point?

mikaelbr: Yep. Just an example. need to plug in env vars etc. Doesn't require a lot though, the point is just that we need a configurable et client. Might construct it from selected org though 🤔

adriansberg: That would work. E.g. `<orgId>-planner-web`.

src/modules/api-client/utils.ts
```diff
@@ -0,0 +1,257 @@
...
+        req,
+        err: e,
+      },
+      e.message
+    );
+  }
+}
+
+const externalHttpUrls = {
+  entur: "https://api.entur.io",
```
adriansberg, line 61: How might i.e. a staging environment tie into this? Just expand on this list as needed?

mikaelbr: Yes, thinking this will be used as an env var in that case. But usually, you won't have the need to go to staging for these endpoints.

adriansberg (approved):

PR #4, opened 2023-09-20
ci: adds tests and CI workflow

mikaelbr:
Adds running tsc and tests on PRs and changes, and add similar release drafter as with webshop.


mikaelbr:
Failing as it looks in main I think

mortennordseth (approved):
👍 Do we want an ESLint workflow as well?

PR #5, opened 2023-09-20
feat: adds example API endpoint for autocomplete

mikaelbr:
Uses same structure as server side props client, but add specific API endpoint for it. Point being that you get injected ready made client with type specification inside HTTP handler.


mortennordseth (approved):
I think this structure looks really good and will make it clean and easy to create new API endpoints 👏

PR #6, opened 2023-09-21
refactor: adds ability to support multiple http methods for endpoints

mikaelbr:
Changes to be able to use:

```ts
export default handlerWithDepartureClient<AutocompleteApiReturnType>({
  async GET(req, res, { client, ok }) { },
  async POST(req, res, { client, ok }) { },
  // ... etc
});
```

And have it automatically return invalid method if nothing is matched. Simplifies HTTP verb handling for each API endpoint.


adriansberg (approved):
🚀 🤯

PR #9, opened 2023-09-22
build: add build and deployment automation

mortennordseth:
- Temporary Error Page: 
I've added a temporary error page so that the application can be built without errors
- Rename env variable: 
I've renamed the environment variable from `NEXT_PUBLIC_WEBSHOP_ORG_ID` to `NEXT_PUBLIC_PLANNER_ORG_ID`
- Build: 
New script in the package.json file: build:all. This runs the new build.sh script that creates a standalone build for all organizations and stores it in the dist directory.
- Docker Image: 
I've added a Dockerfile that creates a single image that can be used to serve all organizations. The new server.js file now determines the decision on which organization to run.

Note: I've tested that each organization handles the assets correctly. However, the theme is not functioning as expected. It's worth noting that this issue persists even when using `yarn dev`, suggesting that the current setup is not the root cause. Please let me know if you see what is missing/wrong about the theme setup so that I can test that as well 👍

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,2 @@
 # What ORG to activate (nfk | atb | fram)
-NEXT_PUBLIC_WEBSHOP_ORG_ID=atb
+NEXT_PUBLIC_PLANNER_ORG_ID=atb
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,7 +4,7 @@ on:
   release:
     types: [published]
   push:
-    branches: ['morten/build-docker']
+    branches: ['main']
 
 jobs:
   build:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,15 @@
+name: Docker Build, Push & Deploy
+
+on:
+  release:
+    types: [published]
+  push:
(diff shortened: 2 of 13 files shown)


adriansberg (approved):
Not seeing why the theme would not work. Is none of the variables in the theme being loaded?

The rest looks good. Did you figure out how we will have to handle build time environment variables?

mortennordseth:
The theme is loaded, but most often for the incorrect org. It seems that the output CSS is the same for all environments. It looks like the webpack alias configuration isn't working properly, but can't say that for sure. Also, this works in the webshop and I can't find any differences so I find it strange 🤔 

To handle build-time environment variables we would have to make some adjustments. I think one solution is to add them as secrets to the repository, suffixed with the org-name. These can be passed as build-args in the common docker workflow. The Dockerfile can then look something like this: 

```Dockerfile
# Build AtB
ARG BUILD_DEPENDENT_VARIABLE_ATB
ENV BUILD_DEPENDENT_VARIABLE=$BUILD_DEPENDENT_VARIABLE_ATB

yarn setup atb
yarn build
# move to dist directory etc..

# Build NFK
...
```

PR #10, opened 2023-09-22
fix: correct env variable name for test setup

mortennordseth:
This updates the env variable name for test setup after updating this in #9 .


mikaelbr (approved):
😂

PR #11, opened 2023-09-22
feat: add initial version of search bar

adriansberg:
This is the initial version of the search bar. It's hooked up to the autocomplete client and gets data from there to show in the menu. 

<details>

image

image

image

image

</details>

## TODO:

- [x] Debounce data fetching
- [x] Add "My position" functionality
- [x] Add icons
- [x] Make mobile friendly
- [x] Verify a11y
- [x] Fix keys with unique IDs
- [x] Add component tests


src/components/search/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { split } from 'lodash';
+
+type SearchProps = {
+  label: string;
+  onChange: (selection: any) => void;
+};
+
+export default function Search({ label, onChange }: SearchProps) {
+  const [query, setQuery] = useState('');
+  const { data } = useAutocomplete(query);
```
mikaelbr, line 14: This was just a simple example. I think for this to be used in a text input like this we should introduce some sort of debouncing to reduce the number of requests 🤔

mikaelbr: (inside the `useAutocomplete` hook)

adriansberg: Agree, was just thinking about that. Adding it to the TODO list.

src/pages/departures.tsx
```diff
@@ -23,17 +24,23 @@ function DeparturesContent({ autocompleteFeatures }: DeparturesContentProps) {
       <input type="text" onChange={(e) => setQuery(e.currentTarget.value)} />
 
       {autocompleteFeatures.map((f, i) => (
-        <div key={i}>{f.name}</div>
+        <div key={i}>
```
mikaelbr, line 27: this should really be a unique ID. Using the index here isn't actually recommended. I can talk at length about why in a call if you want 😅

mikaelbr: It was `i` in the example just to get eslint stop complaining but it shouldn't been in the first place. Just me being lazy (sorry)

adriansberg: I'm aware 😛 Was going to see if the result from the endpoint had an ID, which I assume it has.

adriansberg: Same with the key for the result list.

mikaelbr: It does!

src/components/search/index.tsx
```diff
@@ -0,0 +1,96 @@
+import Downshift from 'downshift';
```
mikaelbr, line 1: I'm trying to be better at being consistent at testing (😅), so it would be interesting to see if we can add some tests to this component

adriansberg: 😳

I should be too. Will get on it and add some. Adding it to the list.

adriansberg: MIght need some inspiration on how to test.

mikaelbr: I think the cookie context tests can be used as inspiration of rendering and user interactions.

mikaelbr: Maybe also with SWR fallback provider (look at swr documentation) to return a result for `useAutocomplete`.

adriansberg: Will have a look at that. I assume we want this isolated and not calling the API.

mikaelbr: Yeah. I think for the actual component test we can use fallback for swr to avoid doing actual request. I haven't actually tried it, but I'm pretty sure swr has some solutions for mocking data. If not it is possible to use some sort of HTTP mocking service like mwn (or what it is called)

adriansberg: Looks like vitest has some mocking: https://vitest.dev/guide/mocking.html

mikaelbr: Yeah, but this is mocking of modules, not of HTTP results. Both are fine, but it depends on of what degree you want to test (unit vs integration)

mikaelbr: mwn (or something like that, I don't actually remember the module name) instead monkey patches `fetch` and returns faux results on HTTP requests. Same, same, but different.

adriansberg: I see they mention MSW: https://vitest.dev/guide/mocking.html#requests

mikaelbr: that's it!

src/components/search/index.tsx
```diff
@@ -25,7 +25,9 @@ export default function Search({ label, onChange }: SearchProps) {
   return (
     <Downshift
       initialInputValue={query}
-      onInputValueChange={(inputValue) => setQuery(inputValue || '')}
+      onInputValueChange={(inputValue) =>
+        debounce(setQuery, 500)(inputValue || '')
```
mikaelbr, line 29: Could see the `useDebounce` hook in webshop. That is quite handy to use the debounce concept in React context

mikaelbr: Oh dang. I see both my examples are the same (taken from webshop and app). And didn't have an example of the useState semantics. But that is OK enough I think :D might as well just keep it consistent and use the same concept here. At some point also maybe even reusable 😱😱

adriansberg: Used that hook instead and moved the debouncing to the `useAutocomplete` hook.

src/pages/departures.tsx
adriansberg: https://github.com/AtB-AS/planner-web/blob/fec409dddaa0a5d2cff39e60ba4ff9b4fd306e37/src/components/search/index.tsx#L81-L83

How should these keys be handled optimally?

mikaelbr: Hmm. I think in this case index is the best we got. But you could wrap the entire thing in a span and let that have the key also 🤔

src/components/search/index.tsx
mikaelbr: Looks good!

mikaelbr (approved):

PR #12, opened 2023-09-25
feat: Mapbox

mortennordseth:
This PR introduces Mapbox integration
- Center position is adjustable with props, default is set by environment variables
- The Mapbox access token is public to the client and we must use access tokens with read-only style rights. We can consider enabling URL restrictions on the token (https://docs.mapbox.com/help/troubleshooting/how-to-use-mapbox-securely/#access-tokens)
- I've added coordinates to the autoCompleteFeature and an onClick function for map-testing purposes. This should be revisited when we start developing the departures page. 

Next up is adding interactive map features, such as geolocation and stop-place interaction, and making a full-screen version for mobile view.

(partially) Fixes https://github.com/AtB-AS/kundevendt/issues/9147

<details>

image
</details>


mortennordseth:
I'm not sure what's causing the error `TypeError: The "obj" argument must be an instance of Blob. Received an instance of Blob` and tests to fail, but I'll investigate more. Let me know if you have any ideas. The Mapbox implementation is ready for review.

src/components/map/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const mapContainer = useRef<HTMLDivElement>(null);
+  const map = useRef<mapboxgl.Map>();
+  const [position, setPosition] = useState<Position>(initialPosition);
+
+  useEffect(() => {
+    if (!mapContainer.current) return;
+    map.current = new mapboxgl.Map({
+      container: mapContainer.current,
+      accessToken: process.env.NEXT_PUBLIC_MAPBOX_API_TOKEN,
+      style: process.env.NEXT_PUBLIC_MAPBOX_STOP_PLACES_STYLE_URL,
```
mikaelbr, line 25: What happens if these aren't added? Should we show some error? build error?

src/components/map/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  useEffect(() => {
+    if (!initialPosition) return;
+    setPosition(initialPosition);
+  }, [initialPosition]);
+
+  useEffect(() => {
+    if (map.current) {
+      map.current.setCenter([position.lng, position.lat]);
+    }
+  }, [position.lat, position.lng]);
```
mikaelbr, line 42: Do we need to store position in separate state here or could we just use position directly in the same way?

mikaelbr: Hmm. I actually think it is OK to create your own type "boundary" here instead of using common type. I don't think we should propagate geojson position around and tightly couple that type with this API

mortennordseth: Okay, I'll revert that change then 👍 What is the reason for not doing so? I guess this applies to the Feature type from geojson as well?

mikaelbr: It's about handling API change. `Position` might not change, but generally, what happens if geojson changes it's API, can we easily map over to be backward compatible or do we have to do an all-out change in the entire codebase? If you use the type directly you are more tightly coupling external dependencies and internal code.

src/components/map/index.tsx
mortennordseth: I agree, we should show an error, and it's probably best to do so at build time. Since we are building one image for all organizations I think the best way to do so is to add them to GitHub secrets with the org name suffixed. This way we can pass them as build arguments to docker build. It would require us to update the environment variables in two places though. What do you think? It could look something like this in the Dockerfile: 

```Dockerfile
#Dockerfile
...
ARG ATB_NEXT_PUBLIC_MAPBOX_API_TOKEN
ARG ATB_NEXT_PUBLIC_MAPBOX_STOP_PLACES_STYLE_URL
ARG NFK_NEXT_PUBLIC_MAPBOX_API_TOKEN
ARG NFK_NEXT_PUBLIC_MAPBOX_STOP_PLACES_STYLE_URL
...
# Build AtB
ENV NEXT_PUBLIC_MAPBOX_API_TOKEN=$ATB_NEXT_PUBLIC_MAPBOX_API_TOKEN
ENV NEXT_PUBLIC_MAPBOX_STOP_PLACES_STYLE_URL=$ATB_NEXT_PUBLIC_MAPBOX_STOP_PLACES_STYLE_URL

yarn setup atb
yarn build
...

# Build NFK
ENV NEXT_PUBLIC_MAPBOX_API_TOKEN=$NFK_NEXT_PUBLIC_MAPBOX_API_TOKEN
ENV NEXT_PUBLIC_MAPBOX_STOP_PLACES_STYLE_URL=$NFK_NEXT_PUBLIC_MAPBOX_STOP_PLACES_STYLE_URL

yarn setup atb
yarn build
...

```

mikaelbr: Maybe it's possible to inject all environments from Github secrets into a separate file that can be included 🤔 As to prevent us from having to add all variables explicitly.

mikaelbr: But I think your solution is a start and we can work on a more "elegant" solution later.

src/components/map/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+    });
+
+    return () => map.current?.remove();
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  useEffect(() => {
+    if (map.current) {
+      map.current.setCenter([initialPosition.lng, initialPosition.lat]);
+    }
+  }, [initialPosition]);
```
mikaelbr, line 37: I think it is still wise to use explicit scalar values here. As this can cause many updates if the object changes reference.

```suggestion
  }, [initialPosition.lng, initialPosition.lat]);
```

src/modules/org-data/index.tsx
```diff
@@ -14,3 +21,21 @@ function getCurrentOrg(): WEBSHOP_ORGS {
...
+
+  if (!accessToken || !style || !defaultLat || !defaultLng) {
+    throw new Error('Mapbox data required but missing');
+  }
+
+  return {
+    accessToken,
+    style,
+    defaultLat: Number(defaultLat),
+    defaultLng: Number(defaultLng),
```
mikaelbr, line 39: We have kind of control of this as it is part of env setup, but I think still this can be more proper parsing:
```suggestion
    defaultLat: parseFloat(defaultLat, 10),
    defaultLng: parseFloat(defaultLng, 10),
```

src/pages/departures.tsx
```diff
@@ -16,25 +17,52 @@ function DeparturesContent({ autocompleteFeatures }: DeparturesContentProps) {
   const { t } = useTranslation();
   const [query, setQuery] = useState('');
   const result = useAutocomplete(query);
+  const [selectedPosition, setSelectedPosition] = useState<
+    Position | undefined
+  >();
```
mikaelbr, line 22: This might be temporary as we don't have selected stop place, but for future reference I think this should come from some selected venue/stop place/address/position with something like:

```ts
const selectedPlace = useSelectedPlaceByUrl();

//...

<Map position={selectedPlace.position} />
```

Where `useSelectedPlaceByUrl` gets ID from URL and fetches specific place by id.

src/components/map/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+import 'mapbox-gl/dist/mapbox-gl.css';
+import { mapboxData } from '@atb/modules/org-data';
+
+export type Position = {
+  lat: number;
+  lng: number;
+};
+
+export type MapProps = {
+  initialPosition?: Position;
```
mikaelbr, line 13: Why is this `initialPosition` and not just `position`? `initial` implies to me that it is only set on initialized component, but the semantics of the component here listens for change on position. So it's not just initial position but just "position"
```suggestion
  position?: Position;
```

mikaelbr:
Might just want to mock mapbox out of the tests for now. Something like

```ts
vi.mock('mapbox-gl/dist/mapbox-gl.js', () => {
  return {
    default: {
      Map: class Map {
        setCenter() {}
        remove() {}
      },
    },
  };
});
```

inside `src/tests/setup.ts`. This will override module and just mock out the central API for Mapbox, causing it to do absolutely nothing.

mikaelbr:
Also I added a task for this in the sprint backlog, so might reference that issue here.

src/components/map/index.tsx
mortennordseth: I agree. I forgot to update this when reverting the type change, my bad.

src/components/map/index.tsx
mortennordseth: You are correct. I initially thought the component would update the position, but I can see that listening for position change is probably better.

src/pages/departures.tsx
mortennordseth: Yes, I was thinking that this was just temporary to have something to test the map with. Thinking about it, it might not be the best to have such code merged into the main branch 🤔 I'm trying to divide the Mapbox issue into multiple PRs and make this only the initial integration of Mapbox, but might want to increase the scope. Not sure

mikaelbr: I think this is task is more relevant for the issue with searching for actual places.

mikaelbr:
What is missing before merge here now, @mortennordseth ?

mortennordseth:
I'm working on handling the required build time environment variables. The solution I suggested isn't a good solution as build-args stores the variables on the image, making them accessible and not secure. It's better to use "secrets" because those are wiped when the docker builder is done. 

I've extended the build script and dockerfile so that it can handle build-time environment variables passed with Docker secrets. What's missing is to generate a file with all variables in GHA and pass that to the workflow. Haven't quite figured out how this can be solved in a way that is maintainable over time.

mortennordseth:
I've added a workaround that can be used with the current build time required env variables, @mikaelbr . In the future, if we need secrets at build time, we should use `secrets` instead of `build-args`. If this looks good to you I can merge this and create an issue that we can solve at a later time.

mikaelbr (approved):
Yep. It's a good start. We can improve later. These are public keys in any case so not too important to have them hidden for now

PR #13, opened 2023-09-26
build: specify overlays for each org

mortennordseth:
We need to specify which overlays we push to, to ensure that the system will pick up new versions. 
Ref: https://github.com/AtB-AS/gcp-infra/pull/415#pullrequestreview-1643544450

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -7,9 +7,30 @@ on:
     branches: ['main']
 
 jobs:
-  build:
+  build-atb:
     uses: atb-as/workflows/.github/workflows/cluster-docker-build-tag-push.yaml@v2
     with:
       image: gcr.io/atb-mobility-platform/planner-web
+      push_overlay: atb-staging-c420
+      release_overlay: atb-prod-2e7e
+      overlays_base: manifests/amp/overlays
+    secrets:
+      github_pat: ${{ secrets.GH_PAT }}
+  build-nfk:
+    uses: atb-as/workflows/.github/workflows/cluster-docker-build-tag-push.yaml@v2
+    with:
+      image: gcr.io/atb-mobility-platform/planner-web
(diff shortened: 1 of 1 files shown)


mikaelbr (commented):
Could use of matrixes be more scalable for multiple orgs here? 🤔

Example:


```
jobs:
  build:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        include:
          - push_overlay: "atb-staging-c420"
             release_overlay: "atb-prod-2e7e"
          - push_overlay: "nfk-staging-dac7"
             release_overlay: "nfk-prod-ce66"
          # - ... osv for fram
  steps:
    uses: atb-as/workflows/.github/workflows/cluster-docker-build-tag-push.yaml@v2
    with:
      image: gcr.io/atb-mobility-platform/planner-web
      push_overlay: {{ matrix.push_overlay }}
      release_overlay: {{ matrix.release_overlay }}
      overlays_base: manifests/amp/overlays
    secrets:
      github_pat: ${{ secrets.GH_PAT }}

```

mortennordseth:
Indeed! I didn't think of using multiple variables for each element in the matrix

mikaelbr:
magic

mikaelbr:
Might be something wrong with my example 🤔 😅 https://github.com/AtB-AS/planner-web/actions/runs/6310892180

mortennordseth:
Yup! Seems like we can't use common workflows in a step 😅 Looking into it!

mortennordseth:
I think this should do the trick and that the job `build` will run three times with the correct overlay values 👍 From the documentation on GHA matrix: 
> A job will run for each possible combination of the variables.

mikaelbr (approved):
Didn't find out how to trigger it manually using `gh workflow` 😅 More research required, but it looks sound

PR #15, opened 2023-09-27
feat: pass on correlation and install id from incoming request for client

mikaelbr:
If client passes correlation id or install id, pass it along to the server (having transient communication of correlation ids)

### DIFF
--- a/src/modules/api-server/__tests__/utils.test.ts
+++ b/src/modules/api-server/__tests__/utils.test.ts
@@ -1,0 +1,157 @@
+import { Mock, afterAll, describe, beforeEach, expect, it, vi } from 'vitest';
+import { createRequester } from '../utils';
+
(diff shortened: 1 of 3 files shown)


src/modules/api-server/__tests__/utils.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+const initialFetch = global.fetch;
+let mockFetch: Mock;
+
+beforeEach(() => {
+  mockFetch = vi.fn();
+  global.fetch = mockFetch;
+});
+afterAll(() => {
+  global.fetch = initialFetch;
```
adriansberg, line 12: What happens if the fetch isn't "reset"?

mikaelbr: It could cause other integration tests to fail

PR #17, opened 2023-09-27
test: add search box tests

adriansberg:


src/tests/setup.ts
```diff
@@ -1,3 +1,4 @@
 import { vi } from 'vitest';
+import '@testing-library/jest-dom/vitest';
```
adriansberg, line 2: Had to add this to be able to use `toBeInDocument()`/`toHaveFocus()` etc. Not quite sure if this is the correct place though.

mikaelbr (approved):
Fantastic!

PR #18, opened 2023-09-28
fix: docker build push deploy workflow

mortennordseth:
This fixes the docker-workflow that is currently on the main branch (after https://github.com/AtB-AS/planner-web/pull/12). It's not an optimal solution, but fixes the current problem at least. We'll work on a better approach later.


mikaelbr (approved):
Havent checked it out too much but can test in ci

mortennordseth:
I've tested it on this branch. The logs from `yarn build` look good as well. See https://github.com/AtB-AS/planner-web/actions/runs/6343296944

PR #19, opened 2023-09-29
feat: add geolocation

adriansberg:
Added functionality to the geolocation button. When a user clicks the geolocation button for the first time it gets prompted to allow geolocation from the browser. If the user denies it, nothing happens. We should probably handle that in some way, have not added that as of now. If the user approves, the navigator.geolocation is called and the coordinates are used to fetch the reversed position/address from the Entur API. That result is then selected in the search bar.

I have not added handling to what happens if the reverse search returns nothing, other than nothing happens. Should possibly show a toast or something both when that happens and if the user denies geolocating.

Closes https://github.com/AtB-AS/kundevendt/issues/9631


src/page-modules/departures/client/geocoder/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  const debouncedQuery = useDebounce(q, 500);
+
+  return useSWR<AutocompleteApiReturnType>(
+    `/api/departures/autocomplete?q=${debouncedQuery}`,
+    swrFetcher,
+  );
+}
+
+export async function reverse(coords: GeolocationCoordinates) {
+  const result = await fetch(
```
adriansberg, line 19: Should we call the API differently from the client, @mikaelbr? Didn't want to use `useSWR` in this case, just a regular fetch.

src/components/__tests__/search.test.tsx
```diff
@@ -22,6 +22,15 @@ const result = [
   },
 ];
 
+const mockGeolocation = {
+  getCurrentPosition: vi.fn(),
+};
+
+Object.defineProperty(global.navigator, 'geolocation', {
+  writable: true,
+  value: mockGeolocation,
```
adriansberg, line 31: Can this be mocked globally in this file?

src/page-modules/departures/client/geocoder/index.ts
mikaelbr: I think for GET requests it's good to use `useSWR` it has SSR support, does caching and can do some polling/mutations if we want. And it's easily composable with other hooks.

src/components/__tests__/search.test.tsx
mikaelbr: Yes, but could be nice to do some proper cleaning afterwords (e.g. resetting it)

src/components/__tests__/search.test.tsx
mikaelbr, line 1: Haven't caught this earlier, but this test folder should be inside the search folder

src/pages/api/departures/reverse.ts
```diff
@@ -0,0 +1,32 @@
...
+        constants.HTTP_STATUS_BAD_REQUEST,
+        ServerText.Endpoints.invalidMethod,
+      );
+    }
+
+    const lat = parseFloat(latQuery.data);
+    const lon = parseFloat(lonQuery.data);
+
+    if (lat === 0 || lon === 0) {
+      return ok(undefined);
```
mikaelbr, line 25: Is this actually OK? Should this rather be bad request? Or isn't this possible? Maybe the parsing above should be directly to float so to avoid this double casing?

```
z.number().safeParse(req.query.lat);
```

or whatever

adriansberg: Remnants of when I used useSWR and it kept revalidating on focus with 0, 0 (fallback values). 

Wasn't able to safeParse the query string directly to a number when I tried 🤔

src/components/__tests__/search.test.tsx
adriansberg: Setting it in `beforeAll` and resetting it in `afterAll` now.

src/page-modules/departures/client/geocoder/index.ts
adriansberg: Tried using it to begin with, but couldn't quite achieve what I wanted. The flow became too asynchronous for me and I felt I didn't have too much control of when I got a position. 

Had something like this. 
```tsx
export default function GeolocationButton({
  onGeolocate,
  className,
}: GeolocationButtonProps) {
  const { position, getPosition, isLoading, isUnavailable } =
    useGeolocation();
  const { data} = useReverse(position?.coords);

  useEffect(() => {
    if (data) {
      onGeolocate(data);
    }
  }, [data]);

  if (isUnavailable) return null;

  return ...
}

function useGeolocation() {
  const [position, setPosition] = useState<GeolocationPosition | null>(null);
  const [error, setError] = useState<GeolocationPositionError | null>(null);
  const [isLoading, setIsLoading] = useState(false);
  const [isUnavailable, setIsUnavailable] = useState(false);

  useEffect(() => {
    if (!navigator.geolocation) setIsUnavailable(true);
  }, []);

  const getPosition = () => {
    setIsLoading(true);

    navigator.geolocation.getCurrentPosition(
      (position) => {
        setIsLoading(false);
        setPosition(position);
      },
      (error) => {
        setIsLoading(false);
        setError(error);
      },
      { enableHighAccuracy: true, timeout: 10000 },
    );
  };

  return {
    position,
    setPosition,
    getPosition,
    getPositionAsync,
    error,
    isLoading,
    isUnavailable,
  };
}
```

mikaelbr: Yep. https://swr.vercel.app/docs/conditional-fetching

adriansberg: Right. I can try that approach some more. The problem I had (one of them at least) was that if the geolocation had been called once, it wouldn't update the data and thus not call `onGeolocate`. Couldn't figure out an elegant way to trigger the call of `onGeolocate` whilst assuring that the fetching of data has succeeded (if it's cached it might not be too big of an issue).

adriansberg: Do you have any tips/recommendation for how to model the data fetching and updating the selected item in the search box when using `useSWR` and relying on state changes?

src/components/search/geolocation-button/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return isLoading ? (
+    <div className={className}>
+      <LoadingIcon />
+    </div>
+  ) : (
+    <button
+      className={className}
+      onClick={() => {
+        getPosition();
+      }}
```
mikaelbr, line 29:
```suggestion
      onClick={getPosition}
```

src/components/search/geolocation-button/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  onGeolocate: (feature: GeocoderFeature) => void;
+  className?: string;
+};
+function GeolocationButton({ onGeolocate, className }: GeolocationButtonProps) {
+  const { t } = useTranslation();
+  const { getPosition, isLoading, isUnavailable } = useGeolocation(onGeolocate);
+
+  if (isUnavailable) return null;
+
+  // TODO: Hide or disable the button if user blocks geolocation?
```
mikaelbr, line 18: I think it is fine having it there so people know it is an option

src/page-modules/departures/client/geocoder/index.ts
mikaelbr: Decided not to useSWR here. Resolving comment thread

mikaelbr (approved):

src/pages/api/departures/reverse.ts
adriansberg: Cleaned it up some, decided to remove the check for 0 since we have control over what's sent in and not.

PR #20, opened 2023-09-29
feat: button components

mortennordseth:
Needed for https://github.com/AtB-AS/kundevendt/issues/9147

I've added the button components from the webshop. From my understanding, the buttons will be the same in the planner-web as in the webshop, so it's nice to reuse them. 

There could be some good tests I'm not seeing that should be included. Let me know if you have any thoughts.


mikaelbr:
Could at some point do some testing for API and behavior here, but it's not as critical I think

mikaelbr (approved):

PR #21, opened 2023-09-29
feat: adds type gen for graphql and departure example

mikaelbr:
Adds integration to JourneyPlanner for getting stop places, nearest, and initial departures.

## Todo

- [x] Map types properly for external use

Related to https://github.com/AtB-AS/kundevendt/issues/9148

---

Showing how to fetch data from GraphQL with code generation

```ts
export function createJourneyApi(
  client: GraphQlRequester<'graphql-journeyPlanner3'>,
): JourneyPlannerApi {
  const api: JourneyPlannerApi = {
    async departures(input) {
      const result = await client.query<
        StopPlaceQuayDeparturesQuery,
        StopPlaceQuayDeparturesQueryVariables
      >({
        query: StopPlaceQuayDeparturesDocument,
        variables: {
          id: input.id,
          startTime: new Date(),
          numberOfDepartures: 10,
        },
      });

      if (result.error) {
        throw result.error;
      }

      return result.data;
    },
  };

  return api;
}
```

### DIFF
--- a/graphql.config.ts
+++ b/graphql.config.ts
@@ -1,9 +1,0 @@
-import type { IGraphQLConfig } from 'graphql-config';
-
-const config: IGraphQLConfig = {
-  schema: 'https://api.entur.io/journey-planner/v3/graphql',
-  documents: 'src/**/*.gql',
-  exclude: ['src/**/*.generated.ts'],
-};
-
-export default config;
--- a/graphql.config.ts
+++ b/graphql.config.ts
@@ -1,0 +1,9 @@
+import type { IGraphQLConfig } from 'graphql-config';
+
+const config: IGraphQLConfig = {
+  schema: 'https://api.entur.io/journey-planner/v3/graphql',
+  documents: 'src/**/*.gql',
+  exclude: ['src/**/*.generated.ts'],
+};
+
+export default config;
--- a/package.json
+++ b/package.json
@@ -12,7 +12,10 @@
     "start": "next start",
     "lint": "next lint",
     "clean-assets": "rimraf ./public/assets/",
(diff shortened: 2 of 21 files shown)


graphql.config.ts
adriansberg, line 1: Could you convert this to yaml/.graphqlrc/json? My IDE didn't like TypeScript.

mikaelbr: Yep! Forgot about that

package.json
```diff
@@ -50,9 +59,11 @@
     "@vitejs/plugin-react": "^4.0.4",
     "eslint": "8.49.0",
     "eslint-config-next": "13.5.1",
+    "graphql-tag": "^2.12.6",
     "happy-dom": "^12.1.5",
     "prettier": "^3.0.3",
     "rimraf": "^5.0.1",
+    "vite": "^4.4.9",
```
adriansberg, line 66: What's `vite` used for?

src/modules/api-server/index.ts
```diff
@@ -13,12 +13,11 @@ import { ExternalClient, ExternalClientFactory } from './external-client';
 export type {
   AllEndpoints,
   HttpRequester,
-  GraphQlRequester,
   ExternalClient,
   ExternalClientFactory,
 };
 
-export { ApplicationError };
+export { GraphQlRequester, ApplicationError };
```
adriansberg, line 20: Why was this moved?

package.json
mikaelbr: It was a warning for a peer dependency of something I can't remember right now.

src/modules/api-server/index.ts
mikaelbr: It's changed from not only being a type that is exported but also the class / constructor

adriansberg (approved):
Looking good 👍

PR #22, opened 2023-10-02
feat: add header and footer

adriansberg:
Added header and footer to the base layout. Reused the components from the online store and removed most of the irrelevant features that is not applicable to the travel planner.

<p>
Full page:
image1

</p>
</details>


mikaelbr (approved):
Should make a issue to make environments configurable

src/layouts/shared/footer.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+        <section className={style.footer__bottomLinks}>
+          <div className={style.footer__some}>
+            {facebookLink && (
+              <ButtonLink
+                mode={isDarkMode ? 'interactive_2' : 'interactive_1'}
+                radius="top-bottom"
+                display="inline"
+                radiusSize="circular"
+                icon={{ left: <img src="/fb.svg" alt="Facebook" /> }}
```
mortennordseth, line 144: Looks like these social media icons are missing

adriansberg: True! Do we want them so I should add them now? Probably?

src/layouts/shared/page-header.module.css
```diff
@@ -0,0 +1,273 @@
+.pageHeader {
+  background-color: var(--static-background-background_0-background);
+}
+
+.pageHeader__content {
+  width: 100%;
+  max-width: var(--maxPageWidth);
+  margin: 0 auto;
```
mortennordseth, line 8: Looking in Figma I think this should be `margin: 0`

adriansberg: You sure? The sketches don't show many examples of wide screens.

src/layouts/shared/footer.tsx
mortennordseth: They are at least in the Figma sketches, so my guess is that we should have them. If not, we should probably remove the links altogether 🤔

adriansberg: I can add them 👍

mortennordseth: It seems like it's copy paste from the webshop though (we are upgrading link etc.), so not sure how thought-through it is 🤔

adriansberg: Faster to remove at a later point than add them now that it is being worked on.

mortennordseth: True 👍

src/layouts/shared/page-header.module.css
mortennordseth: It looks so in this frame 🤔 

image

adriansberg: Let's have a design review meeting at the end of this week to clean up stuff like this. Leaving it as is for now to keep it similar to the webshop's.

mortennordseth (approved):

PR #23, opened 2023-10-02
feat: add routing on search

adriansberg:
Restructured the pages and added a button to route when searching for a stop place or address etc.

Not sure if this is the way we want to do it, but it's something to use to test when implementing the departures and map. 

The way it is set up now with the layout keeps the search box populated between renders.


src/pages/departures/[[...id]].tsx
```diff
@@ -0,0 +1,30 @@
+import DefaultLayout from '@atb/layouts/default';
```
Comment (mikaelbr) [inline src/pages/departures/[[...id]].tsx:1]: shouldn't the file just be named `[...id]`?

src/layouts/departures/index.tsx
mikaelbr, line 1: I think this should be placed in `page-module` instead of layout. As it is only relevant for departures

src/layouts/departures/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+          <Search
+            label={t(PageText.Departures.search.input.from)}
+            onChange={setSelectedFeature}
+          />
+
+          <Button
+            title={t(PageText.Departures.search.button.title)}
+            className={style.searchButton}
+            mode={'interactive_0'}
```
mikaelbr, line 33:
```suggestion
            mode="interactive_0"
```

src/pages/departures/[[...id]].tsx
Comment (adriansberg) [reply src/pages/departures/[[...id]].tsx:1]: https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes#optional-catch-all-segments

Comment (mikaelbr) [reply src/pages/departures/[[...id]].tsx:1]: Ohh. TIL! Makes sense!

src/layouts/departures/index.tsx
adriansberg: True! Will move it.

src/page-modules/departures/__tests__/departure.test.tsx
```diff
@@ -47,67 +28,6 @@ describe('departure page', function () {
     };
 
     const output = render(<DeparturesPage {...initialProps} />);
-    expect(await output.findAllByText(/^Result/)).toHaveLength(2);
-  });
-
-  it('Should return props from getServerSideProps', async () => {
```
adriansberg, line 53: I removed this test for now, and didn't replace it, since I removed all of the data fetching. We can add it later when we fetch data for the departures list/ map.

mortennordseth: Instead of removing the data fetching and tests, maybe we could include the `getServerSideProps` function from earlier using the query parameter? 👍

src/pages/departures/[[...id]].tsx
```diff
@@ -0,0 +1,30 @@
...
+import { withGlobalData } from '@atb/layouts/global-data';
+import { useRouter } from 'next/router';
+import DeparturesLayout, {
+  DeparturesLayoutProps,
+} from '@atb/page-modules/departures/layout';
+
+function DeparturesIdContent() {
+  const router = useRouter();
+
+  console.log(router);
```
Comment (mortennordseth) [inline src/pages/departures/[[...id]].tsx:13]:
```suggestion
```

mortennordseth (commented):
I'm not sure if this is the way we would like to handle routing, but I think this would help with testing as you say 👍 I'll let Mikael have the final say, but it looks good to me

adriansberg:
Can do that.

src/page-modules/departures/__tests__/departure.test.tsx
adriansberg: Can do that.

src/pages/departures/[[...id]].tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+export default DeparturesPage;
+
+export const getServerSideProps = withGlobalData<DeparturesContentProps>(
+  withDepartureClient<DeparturesContentProps>(async function ({
+    client,
+    query,
+  }) {
+    const id = _.first(query.id);
```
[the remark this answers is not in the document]
Comment (adriansberg) [reply src/pages/departures/[[...id]].tsx:61]: Nice, adding that suggestion right away when you add it. Could always do it manually but was lazy in the moment 😅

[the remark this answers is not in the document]
Comment (mikaelbr) [reply src/pages/departures/[[...id]].tsx:61]: See https://github.com/AtB-AS/planner-web/pull/26

I restructured some also to be able to provide correct typings.

mikaelbr (approved):

PR #24, opened 2023-10-02
feat: interactive map

mortennordseth:
This PR should include the necessary functionality to have an interactive map that can be integrated on the departures page. 

## TODO
- [x] Use GeocoderFeature type
- [x] Add map buttons
- [x] Mobile view
- [x] Integrate with departures page (nearest-stop-places)
- [x] Handle screen reader for map
- [ ] ~~Fix map layer warnings or suppress them~~ (Should be fixed in MapBox studio)
- [x] Tests

<details>

image

image

image

image

image

image

</details>

Solves https://github.com/AtB-AS/kundevendt/issues/9147


adriansberg:
2023-10-03 at 14 41 53

The map needs to be position relative to the screen and not the container it looks like. When I clicked to show the map it appears at the top of the screen.

src/components/map/map.module.css
```diff
@@ -1,4 +1,117 @@
...
+  display: flex;
+  gap: var(--spacings-medium);
+}
+
+.header__icons div {
+  background-color: var(--static-background-background_accent_0-background);
+  color: var(--static-background-background_accent_0-text);
+  padding: 6px;
   border-radius: var(--border-radius-regular);
+  width: 40px;
```
adriansberg, line 27: Should we stick to rem?

src/components/map/map.module.css
```diff
@@ -1,4 +1,117 @@
...
+.map__buttons button {
+  margin: var(--spacings-medium);
+}
+
+.map__closeButton {
+  display: none;
+}
+
+.map__showButton {
+  display: none !important;
```
adriansberg, line 77: Is it not possible to get away without using `!important`?

src/translations/components/map.ts
```diff
@@ -0,0 +1,20 @@
+import { translation as _ } from '@atb/translations/commons';
+
+export const Map = {
+  header: {
+    address: _(
+      'Holdeplasser i nærheten av',
+      'Stops near',
+      'Stoppestader i nærleiken av',
```
adriansberg, line 8: Have we used "haldeplasser" previously? Maybe not in this app though.

src/utils/colors.ts
```diff
@@ -0,0 +1,66 @@
+import { MonoIconProps } from '@atb/assets/mono-icon';
+
+export function colorToOverrideColor(
```
adriansberg, line 3: Don't think we need both this file and `utils/color.ts`.

src/components/map/map.tsx
```diff
@@ -0,0 +1,223 @@
...
+    ) => {
+      if (!map) return;
+
+      const features = map.queryRenderedFeatures(e.point);
+      if (
+        features.length &&
+        isFeaturePoint(features[0]) &&
+        isStopPlace(features[0])
+      ) {
+        onSelectStopPlace(features[0].properties?.id);
```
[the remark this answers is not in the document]
adriansberg: Was able to trigger it now, but the hitboxes are a little bit weird. I guess that's out of our control.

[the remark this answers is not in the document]
mortennordseth: I introduced a bug in the last commit, working on a fix as we speak. Having multiple elements in the mapContainer causes issues with the map handlers.

src/utils/colors.ts
mortennordseth: My bad, didn't notice after rebase. Fixed now 👍

src/components/map/map.tsx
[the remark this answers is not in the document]
mortennordseth: The bug should be fixed now. It still needs some refactoring, but it should work as intended at least. Sorry about that!

src/components/map/map.module.css
mortennordseth: Yes, definitely. Bad habits...

src/components/map/map.module.css
mortennordseth: It probably is, but I didn't find out why. I'll look into it when I continue refactoring 👍

src/components/map/map.tsx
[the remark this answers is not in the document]
adriansberg: Works perfectly now 👏

adriansberg:
The icon seems to be uncentered
image
image

src/components/map/map-header.tsx
```diff
@@ -0,0 +1,85 @@
+import style from './map.module.css';
+
+import { ButtonLink } from '../button';
```
mikaelbr, line 3: I think on shared components it is good to use absolute path here
```suggestion
import { ButtonLink } from '@atb/components/button';
```

src/components/map/map.tsx
```diff
@@ -0,0 +1,234 @@
...
+      });
+    }
+  }, [position]);
+
+  return (
+    <div className={style.map}>
+      <Button
+        className={style.fullscreenButton}
+        title={t(ComponentText.Map.map.openFullscreenButton)}
+        icon={{ left: <MonoIcon src="map/Map.svg" /> }}
```
mikaelbr, line 81: I think we should have used the other Icon component from Webshop. But I can fix that in another PR

mikaelbr: (meaning the `@atb/component/icon` one)

src/components/map/map.tsx
```diff
@@ -0,0 +1,234 @@
...
+          icon={{ left: <MonoIcon src="places/City.svg" /> }}
+          onClick={() => centerMap(position)}
+        />
+        <div ref={mapContainer} className={style.mapContainer} />
+      </div>
+    </div>
+  );
+}
+
+const defaultPosition: LngLatPosition = [
```
mikaelbr, line 109: This should probably be defined further up to where it is used. Also, personally I think it is difficult to use arrays for position because I can never remember the order 😓

src/components/map/map.tsx
```diff
@@ -0,0 +1,234 @@
...
+import style from './map.module.css';
+import 'mapbox-gl/dist/mapbox-gl.css';
+import { mapboxData } from '@atb/modules/org-data';
+import { isFeaturePoint, isStopPlace } from './utils';
+import { ColorIcon } from '@atb/assets/color-icon';
+import Button from '../button/button';
+import { MonoIcon } from '@atb/assets/mono-icon';
+import { ComponentText, useTranslation } from '@atb/translations';
+
+export type LngLatPosition = [number, number];
```
mikaelbr, line 12: I think this should be a record with named parameters. Or at the very least named tuples.
```suggestion
export type LngLatPosition = [lng: number, lat: number]
```

src/components/map/map.tsx
```diff
@@ -0,0 +1,234 @@
...
+  onSelectStopPlace: (id: string) => void;
+};
+
+type Marker = HTMLElement &
+  React.DetailedReactHTMLElement<
+    {
+      className: string;
+    },
+    HTMLElement
+  >;
```
mikaelbr, line 25: Hmm. Why is this so advanced? Could it be as simple as `JSX.IntrinsicElements['span']`?

src/components/map/map.tsx
```diff
@@ -0,0 +1,234 @@
...
+      }
+
+      if (marker) marker.setLngLat(position).addTo(map);
+    },
+    [createMapPin, mapRef, marker],
+  );
+
+  const removeMarkerFromMap = useCallback(() => {
+    if (!marker) return;
+    marker.remove();
```
mikaelbr, line 147: This can also be simplified to this

```suggestion
    marker?.remove();
```

mikaelbr:
This happens as image inside button isn't `display: block`

Screenshot 2023-10-03 at 21 29 13

mikaelbr:
Feeling like the zoom level should be closer?

src/components/map/map.tsx
mortennordseth: You're not alone 😄 I'll fix that

src/components/map/map.tsx
mortennordseth: I meant to ask about this, but forgot. The constructor for the mapboxgl Marker expects an element of type `HTMLElement` and I'll get a ts error using `JSX.IntrinsicElements['span']` as there are a lot more properties to that type 🤔

mikaelbr: Are you sure this even works? 😅 I have a hard time seeing how this would render properly. CreateElement here returns an JSON structure which react in turns renders as DOM operations at some point. But passing this to setMarker API of mapbox has no idea what to do with that JSON structure. Also the Icon path looks to be wrong?

mortennordseth: You are absolutely correct, this isn't working 🤦‍♂️ I'll fix it asap.

mortennordseth: I've solved this using React createRoot. I also added a new component, `PinSvg`, so that I can use the CSS fill property to set the correct color on the SVG.

mikaelbr: You could also do this the old fashioned way:

```ts
const pin: HTMLElement = document.createElement('img');
pin.src = '/assets/map/Pin.svg';
setMarker(pin);
```

mortennordseth: That's neat! The design says that we should use the `interactive_0` color for the pin and that's not straightforward to use when using an `` though, is it? 🤔

mikaelbr: Hm.. No, but I think maybe the color of the pin isn't as important and might not be a hard require. Often in these cases I'd verify and think it over with the designer. Because it's not really interacteble and interactive_0 implies primary function. Also black is more contrast then blue interactive on some blue water for instance. So I think using map/Pin.svg is a better choice here, but like I said, I think verifying/discussing this is a good way to go

mortennordseth: That's true. Good point. I'll have a chat with the designer 👍

mortennordseth: I've talked with the designer and we'll go for the black SVG. Updated the code to use `document.createElement` 👍

src/components/map/map.module.css
mortennordseth: Fixed 👍

mortennordseth:
I've looked into and resolved the comments you've made. What's not done now is creating tests for the actual map and handle the Mapbox errors. The tests in not so straightforward and I'll create an issue so that can be handled in another PR. The Mapbox errors should be fixed in Mapbox Studio and since these errors already exists on the main branch I don't think that should block this from being merged.

adriansberg (approved):
Great work, @mortennordseth! 👏🚀🗺️

PR #25, opened 2023-10-03
feat: adds page for nearest stop places

mikaelbr:
- Adds navigation from search input to location based and stop place based content.
- Adds listing stop places near a location

Splitting this task for collaboration. Need to follow up on things like proper design as a separate task.

## Todo
- [ ] Fix design for stop place listing.

<details>
Screenshot 2023-10-04 at 22 28 37
Screenshot 2023-10-04 at 22 28 29

</details>

### DIFF
--- a/src/components/search/geolocation-button/index.tsx
+++ b/src/components/search/geolocation-button/index.tsx
@@ -25,6 +25,7 @@ function GeolocationButton({ onGeolocate, className }: GeolocationButtonProps) {
       onClick={getPosition}
       title={t(ComponentText.GeolocationButton.alt)}
       aria-label={t(ComponentText.GeolocationButton.alt)}
+      type="button"
     >
       <MonoIcon icon="places/City" />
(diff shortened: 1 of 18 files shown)


src/components/search/geolocation-button/index.tsx
```diff
@@ -25,6 +25,7 @@ function GeolocationButton({ onGeolocate, className }: GeolocationButtonProps) {
       onClick={getPosition}
       title={t(ComponentText.GeolocationButton.alt)}
       aria-label={t(ComponentText.GeolocationButton.alt)}
+      type="button"
```
mikaelbr, line 28: This is to avoid triggering on enter as buttons default to submit

mortennordseth (approved):
This is looking good! I think it will be nice to have this now for further collaboration as you say. Also, I gotta say, seeing and learning something new from all of these PRs is great 😄

PR #26, opened 2023-10-03
refactor: change to use params instead of query

mikaelbr:


adriansberg (approved):
Looks good. So `params` is similar to `query` and the slug gets added to that property as well?

mikaelbr:
Yep, and in this case I think it's always an array: https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes#optional-catch-all-segments

PR #29, opened 2023-10-05
fix: map container max height on departures page

mortennordseth:
This makes it so that the map container is max 600px. 

image


adriansberg (approved):
👌

PR #33, opened 2023-10-09
refactor: map header icons and text

mortennordseth:
This PR updates the MapHeader icons from FeatureCategory to TransportMode to fit better with the StopPlace type. I also removed the street property from the MapHeader as this doesn't exist on the StopPlace object. I'll ask the designer what we should do about the subtitle in the header, but for now, I think it's better to remove it and don't use the question mark as a placeholder. 

Closes https://github.com/AtB-AS/kundevendt/issues/11279

<details>

image

image

</details>

### DIFF
--- a/src/components/map/map-header.tsx
+++ b/src/components/map/map-header.tsx
@@ -50,7 +50,7 @@ export function MapHeader({ id, name, layer, transportModes }: MapHeaderProps) {
               {t(ComponentText.Map.header.address)}
             </p>
           )}
-          <h3 className={and('typo-heading--medium')}>{name}</h3>
+          <h2 className={and('typo-heading--medium')}>{name}</h2>
         </div>
       </div>
 
--- a/src/components/map/map-header.tsx
+++ b/src/components/map/map-header.tsx
@@ -27,7 +27,7 @@ export function MapHeader({ id, name, layer, transportModes }: MapHeaderProps) {
             </div>
           ) : (
             transportModes.map((mode) => (
-              <div key={[mode, 'icon'].join('-')}>
+              <div key={`${mode}-icon`}>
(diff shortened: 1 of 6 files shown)


src/components/map/map-header.tsx
```diff
@@ -2,61 +2,55 @@ import style from './map.module.css';
...
+            <p
+              className={and(
+                'typo-body__secondary',
+                style['header__info__secondary'],
+              )}
+            >
+              {t(ComponentText.Map.header.address)}
+            </p>
+          )}
+          <h3 className={and('typo-heading--medium')}>{name}</h3>
```
mikaelbr, line 53: I think these should be h2 🤔

src/components/map/map-header.tsx
```diff
@@ -2,61 +2,55 @@ import style from './map.module.css';
...
+          {layer === 'address' || !transportModes ? (
             <div>
               <MonoIcon size="large" icon="map/Pin" overrideMode="dark" />
             </div>
           ) : (
-            category.map((type) => (
-              <div key={[type, 'icon'].join('-')}>
-                <VenueIcon category={[type]} size="large" overrideMode="dark" />
+            transportModes.map((mode) => (
+              <div key={[mode, 'icon'].join('-')}>
```
mikaelbr, line 30: Is this shorter and more intuitive than this?
```suggestion
              <div key={`${mode}-icon`}>
```

mikaelbr: (he asks almost passive-aggressively and rhetorically)

mortennordseth: 😄 I agree though

mikaelbr (approved):

PR #34, opened 2023-10-09
feat: add filter for transport mode

adriansberg:


src/components/travel-search-filter/index.tsx
```diff
@@ -0,0 +1,255 @@
+import { getTextForLanguage } from '@atb/translations/utils';
+import { useTranslation } from '@atb/translations';
+import { ChangeEvent, useState } from 'react';
+import style from './travel-search-filter.module.css';
+import { getTransportModeIconPath } from '@atb/components/travel-search-filter/utils';
+import { MonoIcon } from '@atb/components/icon';
+import { TransportModeFilterOptionWithSelectionType } from '@atb/components/travel-search-filter/types';
+import { Typo } from '@atb/components/typography';
+
+const defaultTravelSearchFilter: TransportModeFilterOptionWithSelectionType[] =
```
adriansberg, line 10: How should we handle this initial state? I assume it needs to be passed as prop, but should we have it hard coded or set up firestore for this and other stuff?

src/components/travel-search-filter/types.ts
```diff
@@ -0,0 +1,1643 @@
+import { z } from 'zod';
+
+export declare const TransportModeFilterOptionType: z.ZodObject<
```
adriansberg, line 3: I reused this from config-specs which is used in the app. I don't have access to that repo, so this is taken from code inspection. Are there any simpler versions of this somewhere?

src/components/travel-search-filter/types.ts
```diff
@@ -0,0 +1,1643 @@
+import { z } from 'zod';
```
adriansberg, line 1: Should zod objects/definitions be placed in separate files or is it okay to have them with the types?

src/components/travel-search-filter/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+  METRO = 'metro',
+  RAIL = 'rail',
+  TRAM = 'tram',
+  WATER = 'water',
+}
+
+export type AnyMode = LegMode | Mode | TransportModeType | 'flex';
+export type AnySubMode = TransportSubmodeType;
+
+const TRANSPORT_SUB_MODES_BOAT: AnySubMode[] = [
```
adriansberg, line 29: @mortennordseth, some of this might overlap with the implementation you did with the TransportMode component. How should we handle duplicates?

src/components/travel-search-filter/types.ts
mortennordseth: Any reason to not use the package instead? 🤔  `yarn add @atb-as/config-specs`

adriansberg: I guess we can. Not familiar with it. Is it public?

src/components/travel-search-filter/index.tsx
mortennordseth: I'm not sure if we should use Firestore or not, but I just want to comment that there is some infra work needed if we want to use Firestore. Maybe we could hard code it for now and potentially connect to Firestore in another PR?

adriansberg: The Firestore addition would happen in a later PR regardless. More so if the placement of the state should be temporary or not.

src/components/travel-search-filter/utils.ts
mortennordseth: Yup, there's some overlap. For example `TRANSPORT_SUB_MODES_BOAT` and `getTransportModeIconPath` already exist in `@atb/components/transport-mode/transport-icon/index.tsx`. I think we should coordinate these types and functions to avoid duplicates 🤔 Maybe we should add `@atb-as/config-specs` and use that as the standard?

src/components/travel-search-filter/types.ts
mortennordseth: It's not public. I think @mikaelbr is the owner and can add you (or the AtB Utvikling team)

adriansberg: The repo isn't public, that I knew. But the package itself. Do we need to set up some auth to be able to use the package or is it public?

src/components/travel-search-filter/utils.ts
mortennordseth: The transport mode types at least. If that makes them too connected we probably need duplicates, then? 🤔

adriansberg: It might not. If we expand on the transport-mode to work with the travel search filter that would work. Then we can contain those types in the transport-mode folder and only use them in the travel search filter.

src/components/travel-search-filter/types.ts
adriansberg: `[x] Travel Search Filters`, perfect.

@mikaelbr gets to be the decision maker for using this package or not.

mikaelbr: This package is intended to ensure contracts between firestore and clients. So should use it here! It is public and you can use it as any other package.

src/components/travel-search-filter/types.ts
mikaelbr: For cases where you have both zod and definitions I try to use zod and `z.infer` to create types so we only have one source of truth.

src/components/travel-search-filter/types.ts
adriansberg: Ok! Adding the package and using it then.

src/page-modules/departures/layout.tsx
```diff
@@ -40,39 +73,69 @@ function DeparturesLayout({ children }: DeparturesLayoutProps) {
...
               onChange={setDepartureDate}
             />
           </div>
+        </div>
+
+        <div className={style.buttons}>
+          <Button
+            title={t(PageText.Departures.search.buttons.alternatives.title)}
+            className={style.button}
+            mode={showAlternatives ? 'interactive_3' : 'interactive_2'}
```
adriansberg, line 105: The Button component doesn't support the mode used in Figma. Should we add it, and if so what should the mode be called?

2023-10-12 at 23 13 53

src/page-modules/departures/layout.tsx
```diff
@@ -40,39 +73,69 @@ function DeparturesLayout({ children }: DeparturesLayoutProps) {
...
+            </Typo.p>
 
             <DepartureDateSelector
               initialState={departureDate}
               onChange={setDepartureDate}
             />
           </div>
+        </div>
+
+        <div className={style.buttons}>
```
adriansberg, line 101: Is it possible to make these buttons be selected before the alternatives even though it precedes in order?

adriansberg:
Let's try to get this PR merged before any other changes are made. Minor things are okay I think, but to avoid this getting even more crazy in size I think larger changes should be done in subsequent PRs. Everything should work as expected. My brain haven't been online the past couple of days, so please point out things that should not be done the way they are.

I decided to get rid of the enums for transport modes and use the type from `@atb-as/config-specs`. That way we can abide by a single source of truth instead of maintaining several enums in different places. Consequences of that were 1) we have to use strings instead of enums, but type safety should be in place still, 2) we have to convert strings to enums when passing it to queries. 

Having the object `filterOptionsWithTransportModes` limits the configurability of the filters, so we probably want to rethink that part.

Filtering in the queries was a bit tricky because it's only possible to filter on main transport modes and not submodes. Some of the filter options available have multiple transport modes, but the transport modes have relevant submodes (e.g. `rail` has the transport mode `bus`, but submode `replacementBus`). So if only "Train" is selected we still get regular bus departures since it's only filtering on `bus` and not the submode.

Haven't added any animation to the appearance of the alternatives section yet, but I think we can add a nice one here.

src/page-modules/departures/layout.tsx
```diff
@@ -10,27 +10,60 @@ import DepartureDateSelector, {
   DepartureDate,
   DepartureDateState,
 } from '@atb/components/departure-date-selector';
+import TransportModeFilter from '../../components/transport-mode-filter';
```
mortennordseth, line 13:
```suggestion
import TransportModeFilter from '@atb/components/transport-mode-filter';
```

src/page-modules/departures/layout.tsx
mortennordseth: Maybe you could use the FocusScope from react-aria to set focus when the showAlternatives button is clicked? This way the user doesn't have to tab through the "find departures" button?

src/page-modules/departures/server/journey-planner/journey-gql/stop-place.gql
```diff
@@ -2,6 +2,7 @@ query getStopPlace($id: String!) {
   stopPlace(id: $id) {
     name
     id
+    description
```
mortennordseth, line 5: Is this needed? I couldn't find any use of it

mortennordseth (commented):
Great work! I think we could improve the keyboard navigation using react-aria, but other than that I think this looks great 🎉

src/page-modules/departures/layout.tsx
adriansberg: But I don't want to trap the focus within the alternatives, the user should be able to keep the alternatives open.

src/page-modules/departures/server/journey-planner/journey-gql/stop-place.gql
adriansberg: The parsing threw an error without it since the type says it's nullable but not optional. Could always have changed the type though, if it's never going to be used.

src/page-modules/departures/layout.tsx
mortennordseth: That's configurable. I think we could do something like this (probably need some other tweaks but you get the idea)
```tsx
<FocusScope contain={false} restoreFocus autoFocus={showAlternatives}>
{showAlternatives && (...)}
</FocusScope>
```

src/page-modules/departures/server/journey-planner/journey-gql/stop-place.gql
mortennordseth: I think this is my fault from a previous PR. If you want you can remove the description from stopPlace as that is not in use or else we can do that in a separate PR 👍

adriansberg: Turns out it is used in some places, e.g.
image

I say we can leave it.

mortennordseth: Alright, leave it 👍

src/page-modules/departures/layout.tsx
adriansberg: Added FocusScope and reordered the elements, so the order should now be correct.

src/page-modules/departures/layout.tsx
adriansberg: Yes, let's add it later.

mortennordseth (approved):
It works great now 👍 Regarding the button I think it's all right to create a new issue for adding the new buttons (primary, secondary and tertiary), but you decide.

PR #35, opened 2023-10-11
feat: load more departures for a specific quay

mortennordseth:
This PR includes new functionality for loading more departures for a specific quay. 

Closes https://github.com/AtB-AS/kundevendt/issues/9149

### DIFF
--- a/src/page-modules/departures/server/journey-planner/index.ts
+++ b/src/page-modules/departures/server/journey-planner/index.ts
@@ -29,6 +29,8 @@ import {
   estimatedCallsSchema,
   nearestStopPlaces,
   stopPlaceSchema,
+  Quay,
+  Departure,
 } from './validators';
 import {
   TransportMode,
@@ -43,7 +45,7 @@ export type DepartureInput = {
   id: string;
   transportModes: TransportModeFilterOption[] | null;
 };
-export type { DepartureData };
+export type { DepartureData, Quay, Departure };
 
 export type StopPlaceInput = {
   id: string;
--- a/src/page-modules/departures/server/journey-planner/index.ts
+++ b/src/page-modules/departures/server/journey-planner/index.ts
@@ -229,7 +229,7 @@ export function createJourneyApi(
         query: QuayEstimatedCallsDocument,
         variables: {
           id: input.quayId,
-          numberOfDepartures: 10,
+          numberOfDepartures: 5,
           startTime: new Date(input.startTime),
         },
       });
--- a/src/page-modules/departures/server/journey-planner/index.ts
+++ b/src/page-modules/departures/server/journey-planner/index.ts
@@ -248,7 +248,9 @@ export function createJourneyApi(
           date: e.date,
           expectedDepartureTime: e.expectedDepartureTime,
           aimedDepartureTime: e.aimedDepartureTime,
-          transportMode: isTransportMode(e.serviceJourney.line.transportMode)
+          transportMode: isTransportModeType(
+            e.serviceJourney.line.transportMode,
+          )
             ? e.serviceJourney.line.transportMode
             : undefined,
           transportSubmode: e.serviceJourney.line.transportSubmode,
--- a/src/page-modules/departures/server/journey-planner/index.ts
+++ b/src/page-modules/departures/server/journey-planner/index.ts
@@ -15,11 +15,18 @@ import {
   GetStopPlaceQueryVariables,
 } from './journey-gql/stop-place.generated';
 import {
+  QuayEstimatedCallsDocument,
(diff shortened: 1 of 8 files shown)


src/page-modules/departures/stop-place/index.tsx
```diff
@@ -115,7 +137,8 @@ export function EstimatedCallList({ quay }: EstimatedCallListProps) {
             <button
               className={and(style.listItem, style.listItem__last)}
               aria-label={t(Departures.stopPlace.quaySection.a11yToQuayHint)}
-              onClick={() => alert('Not implemented yet')} // @TODO: implement function
+              onClick={getMoreDepartures}
+              disabled={isFetchingDepartures}
```
adriansberg, line 141: Should it possibly be a loading state as well, so it shows that more departures are being fetched?

adriansberg: Let me rephrase, since the answer to the previous question is of course that's what `isFetchingDepartures` is.

Rephrased: should we show a spinner or something that indicates that departures are being fetched? Just disabling the button might not be indicative enough. Also a11y labels.

src/page-modules/departures/server/journey-planner/index.ts
```diff
@@ -194,6 +207,48 @@ export function createJourneyApi(
...
+
+    async estimatedCalls(input) {
+      const result = await client.query<
+        QuayEstimatedCallsQuery,
+        QuayEstimatedCallsQueryVariables
+      >({
+        query: QuayEstimatedCallsDocument,
+        variables: {
+          id: input.quayId,
+          numberOfDepartures: 10,
```
adriansberg, line 219: Should we get as much as 10 or is 5 enough?

src/page-modules/departures/stop-place/index.tsx
mortennordseth: > should we show a spinner or something that indicates that departures are being fetched? 

Yes, I think that's better. I've updated to show a spinner while fetching.

> Also a11y labels.

Do you mean a11y label when the button is disabled? As in changing the current `aria-label` when `isFetchingDepartures` is true?

src/page-modules/departures/server/journey-planner/index.ts
mortennordseth: I changed it to 5. I think that's a better experience for the user 👍

src/page-modules/departures/stop-place/index.tsx
mortennordseth: Ah, you were probably thinking of `aria-disabled`. That's also fixed now.

adriansberg: `aria-disabled` should be set to either `true` or `false` from what I can see. With a11y labels I meant that it should be clear to the user when the loading spinner is showing (and the departures are being fetched), and by only disabling the button it might not be obvious.

mortennordseth: With the current change the `aria-disabled` is set to `true` when the state is `loading` (done by the Button component).

adriansberg: Isn't it set to `aria-disabled={'DISABLED'}`? That's what I see. 🤔

mortennordseth: Sorry about the confusion! I pushed some changes that I was unaware of.

src/pages/api/departures/journey-planner.ts
```diff
@@ -0,0 +1,31 @@
...
+        res,
+        constants.HTTP_STATUS_BAD_REQUEST,
+        ServerText.Endpoints.invalidMethod,
+      );
+    }
+
+    return tryResult(req, res, async () => {
+      return ok(
+        await client.estimatedCalls({
+          quayId: String(req.query.quayId),
```
adriansberg, line 25: Can use `query.data` here, no?

src/page-modules/departures/stop-place/index.tsx
```diff
@@ -55,6 +57,27 @@ type EstimatedCallListProps = {
 export function EstimatedCallList({ quay }: EstimatedCallListProps) {
   const { t } = useTranslation();
   const [isCollapsed, setIsCollapsed] = useState<boolean>(false);
+  const [departures, setDepartures] = useState<
+    DepartureData['quays'][0]['departures'][0][]
```
adriansberg, line 61: There is no better type for this?

src/pages/api/departures/journey-planner.ts
mortennordseth: True!

adriansberg (approved):

PR #36, opened 2023-10-13
feat: button for refreshing the estimated calls list

mortennordseth:
This introduces a button to refresh the estimated calls list.

<details>

image

image

</details>

### TODO
- [x] Add correct icon (waiting on https://github.com/AtB-AS/design-system/pull/145)

Closes https://github.com/AtB-AS/kundevendt/issues/9149

### DIFF
--- a/src/components/icon/mono-icon.tsx
+++ b/src/components/icon/mono-icon.tsx
@@ -1,7 +1,8 @@
-import { TextColorType } from '@atb-as/theme';
+import { InteractiveColor, TextColorType, Theme } from '@atb-as/theme';
 import { MonoIcons, icons } from './generated-icons';
 import { SizeProps, useSize } from './utils';
-import { useDarkMode } from '@atb/modules/theme';
+import { useDarkMode, useTheme } from '@atb/modules/theme';
+import { colorToOverrideMode } from '@atb/utils/color';
 
 export type { SizeProps };
 export type MonoIconProps = Omit<JSX.IntrinsicElements['img'], 'src'> & {
@@ -19,18 +20,31 @@ export type MonoIconProps = Omit<JSX.IntrinsicElements['img'], 'src'> & {
    * @default none
    */
   overrideMode?: 'none' | 'dark' | 'light';
+  /**
+   * Used to override light- or dark mode icon color when MonoIcon is used with an interactive component
+   */
+  interactiveColor?: keyof Theme['interactive'];
+  interactiveState?: keyof InteractiveColor;
 };
 
 export function MonoIcon({
   icon,
   size = 'normal',
   overrideMode = 'none',
+  interactiveColor,
+  interactiveState,
   ...props
 }: MonoIconProps) {
   const [isDarkMode] = useDarkMode();
   const currentIcon = icons['mono'][icon];
+  const interactiveOverrideColor = useInteractiveThemeColor(
+    interactiveColor,
+    interactiveState,
+  );
   const darkModeString =
-    colorToMode(overrideMode) ?? (isDarkMode ? 'dark' : 'light');
+    colorToMode(overrideMode) ??
+    colorToMode(interactiveOverrideColor) ??
+    (isDarkMode ? 'dark' : 'light');
   const totalPath = `/assets/${insertMode(
     currentIcon.relative,
     darkModeString,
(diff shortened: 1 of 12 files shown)


mortennordseth:
FYI @mikaelbr, I misread the Figma sketches when I started working on the refresh button for the estimated calls list. As I'd already started working on the functionality needed for the nearest-stop-places page I chose to include it in this PR, making this PR cover both refresh buttons. This is still WIP and I'll continue this work after I've looked at the other open PRs. I just wanted to mention it since you've been working on the nearest-stop-places page

mikaelbr:
I don't think refresh on stop page makes sense. So I didn't include it in my original page. There is nothing to refresh, so I think it's just an error in the design

mortennordseth:
@mikaelbr or @adriansberg , do you see a better way to handle the icon color when hovering over the link? With the current solution the text changes color before the icon and makes it a bit weird. 

(also ignore the icon type, need to merge this before I can add the correct one: https://github.com/AtB-AS/design-system/pull/145)

https://github.com/AtB-AS/planner-web/assets/43166974/f814660a-a5ab-45f5-abf8-bc56b0859f2e

adriansberg:
> @mikaelbr or @adriansberg , do you see a better way to handle the icon color when hovering over the link? With the current solution the text changes color before the icon and makes it a bit weird.
> 
> (also ignore the icon type, need to merge this before I can add the correct one: [AtB-AS/design-system#145](https://github.com/AtB-AS/design-system/pull/145))
> 
>  Skjermopptak.2023-10-16.kl.13.34.02.mov

If I understood it correctly, https://github.com/AtB-AS/kundevendt/issues/9140 sort of fixes this by enabling the icons to be manipulated by css, thus being able to add the wanted transition of color to match the transition of the text.

I can't think of any solution to this with the current setup for icons.

A work-around might be to remove the color transition of the text so that it's less weird for now, but not sure how it will be with this button (and buttons with icon) being different than others.

mortennordseth:
> A work-around might be to remove the color transition of the text so that it's less weird for now

This makes it less weird for sure. Not the best solution, but as you say, the other issue (https://github.com/AtB-AS/kundevendt/issues/9140) might fix this.

src/components/icon/mono-icon.tsx
```diff
@@ -19,18 +20,31 @@ export type MonoIconProps = Omit<JSX.IntrinsicElements['img'], 'src'> & {
    * @default none
    */
   overrideMode?: 'none' | 'dark' | 'light';
+  /**
+   * Used to override light- or dark mode icon color when MonoIcon is used with an interactive component
+   */
+  interactiveColor?: keyof Theme['interactive'];
```
adriansberg, line 26: Are `interactiveColor` and `interactiveState` both going to be present at the same time?

src/styles/globals.css
```diff
@@ -99,8 +99,7 @@ textarea {
   --contentPadding: var(--spacings-xLarge);
 }
 
-a,
-a:visited {
```
adriansberg, line 103: Why was this removed?

mikaelbr:
Icon color here is one of the reasons why I want https://github.com/AtB-AS/kundevendt/issues/9140 as you've said. I think this technically could be solved for mono icons using CSS filters, but I think it's fine as is for now. But I'd want to verify that the contrast is actually correct there though. Even though it uses theme colors, there might be a contrast issue in the theme design.

src/components/icon/mono-icon.tsx
mortennordseth: It can be. There is no need for the `interactiveState` if you don't use the `interactiveColor`, but you can use the `interactiveColor` without the `interactiveState`. Most likely you would use both at the same time to handle hover/disabled colors etc.

src/styles/globals.css
mortennordseth: The reason for this was that this overrides the interactive color on the refresh button. From what I can see, this is necessary for the links in the footer, but that is already covered here: https://github.com/AtB-AS/planner-web/blob/main/src/layouts/shared/footer.module.css#L46. 

Let me know if you see any reason why we need this 👍

mortennordseth:
AtBs interactive 1 color for hover passes WCAG AA for normal text.

src/styles/globals.css
adriansberg: I guess it's wanted in cases we have links in other places than e.g. the footer where it's covered. But not too big of a deal.

adriansberg (approved):

PR #39, opened 2023-10-23
feat: assistant page header

mortennordseth:
This includes the assistant page header and the required functionality to get a trip with trip patterns. We might want to fine-tune the data included in a trip, but that might be easier to do when we start to work on the assistant page content. The search is stored in the URL. 

### TODO
- [x] Update search bar text on swap
- [x] Tests

Closes https://github.com/AtB-AS/kundevendt/issues/11317

<details>

image

image

image

</details>

### DIFF
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -11,6 +11,7 @@ type SearchProps = {
   onChange: (selection: any) => void;
   button?: ReactNode;
   initialFeature?: GeocoderFeature;
+  selectedItem?: GeocoderFeature;
 };
 
 export default function Search({
@@ -18,16 +19,17 @@ export default function Search({
   onChange,
   button,
   initialFeature,
+  selectedItem,
 }: SearchProps) {
   const [query, setQuery] = useState('');
   const { data } = useAutocomplete(query);
 
   return (
     <Downshift<GeocoderFeature>
-      initialSelectedItem={initialFeature}
       onInputValueChange={(inputValue) => setQuery(inputValue || '')}
       onChange={onChange}
       itemToString={geocoderFeatureToString}
+      selectedItem={selectedItem || initialFeature || null}
     >
       {({
         getInputProps,
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -25,7 +25,6 @@ export default function Search({
   return (
     <Downshift<GeocoderFeature>
       initialSelectedItem={initialFeature}
-      initialInputValue={geocoderFeatureToString(initialFeature)}
       onInputValueChange={(inputValue) => setQuery(inputValue || '')}
       onChange={onChange}
       itemToString={geocoderFeatureToString}
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -6,17 +6,20 @@ import VenueIcon from '@atb/components/venue-icon';
 import { andIf } from '@atb/utils/css';
 import GeolocationButton from '@atb/components/search/geolocation-button';
 import { GeocoderFeature } from '@atb/page-modules/departures';
+import SwapButton from '@atb/components/search//swap-button';
 
 type SearchProps = {
   label: string;
   onChange: (selection: any) => void;
   initialQuery?: string;
+  onSwap?: () => void;
 };
 
 export default function Search({
   label,
   onChange,
   initialQuery = '',
+  onSwap,
 }: SearchProps) {
   const [query, setQuery] = useState(initialQuery);
   const { data } = useAutocomplete(query);
@@ -51,10 +54,14 @@ export default function Search({
             <input className={style.input} {...getInputProps()} />
           </div>
 
-          <GeolocationButton
-            className={style.geolocationButton}
-            onGeolocate={selectItem}
-          />
+          {onSwap ? (
+            <SwapButton className={style.geolocationButton} onSwap={onSwap} />
+          ) : (
+            <GeolocationButton
+              className={style.geolocationButton}
+              onGeolocate={selectItem}
+            />
+          )}
 
           <ul className={style.menu} {...getMenuProps()}>
             {isOpen &&
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -1,35 +1,34 @@
 import Downshift from 'downshift';
-import { useState } from 'react';
+import { ReactNode, useState } from 'react';
 import { useAutocomplete } from '@atb/page-modules/departures/client';
 import style from './search.module.css';
 import VenueIcon from '@atb/components/venue-icon';
 import { andIf } from '@atb/utils/css';
-import GeolocationButton from '@atb/components/search/geolocation-button';
 import { GeocoderFeature } from '@atb/page-modules/departures';
-import SwapButton from '@atb/components/search//swap-button';
 
 type SearchProps = {
   label: string;
   onChange: (selection: any) => void;
-  initialQuery?: string;
-  onSwap?: () => void;
+  button?: ReactNode;
+  initialFeature?: GeocoderFeature;
 };
 
 export default function Search({
   label,
   onChange,
-  initialQuery = '',
-  onSwap,
+  button,
+  initialFeature,
 }: SearchProps) {
-  const [query, setQuery] = useState(initialQuery);
+  const [query, setQuery] = useState('');
   const { data } = useAutocomplete(query);
 
   return (
     <Downshift<GeocoderFeature>
-      initialInputValue={query}
+      initialSelectedItem={initialFeature}
+      initialInputValue={geocoderFeatureToString(initialFeature)}
       onInputValueChange={(inputValue) => setQuery(inputValue || '')}
       onChange={onChange}
-      itemToString={(item) => (item ? `${item.name}, ${item.locality}` : '')}
+      itemToString={geocoderFeatureToString}
     >
       {({
         getInputProps,
@@ -40,7 +39,6 @@ export default function Search({
         inputValue,
         highlightedIndex,
         getRootProps,
-        selectItem,
       }) => (
         <div className={style.container}>
           <label className={style.label} {...getLabelProps()}>
@@ -54,14 +52,7 @@ export default function Search({
             <input className={style.input} {...getInputProps()} />
           </div>
 
-          {onSwap ? (
-            <SwapButton className={style.geolocationButton} onSwap={onSwap} />
-          ) : (
-            <GeolocationButton
-              className={style.geolocationButton}
-              onGeolocate={selectItem}
-            />
-          )}
+          {button ?? null}
 
           <ul className={style.menu} {...getMenuProps()}>
             {isOpen &&
@@ -125,4 +116,9 @@ function highlightSearchText(input: string | null, name: string) {
   } else {
     return [{ part: name, highlight: false }];
   }
+}
+function geocoderFeatureToString(
+  feature: GeocoderFeature | null | undefined,
+): string {
+  return feature ? `${feature.name}, ${feature.locality}` : '';
 }
--- a/src/page-modules/assistant/assistant.module.css
+++ b/src/page-modules/assistant/assistant.module.css
@@ -66,6 +66,18 @@
   z-index: 10;
 }
 
+.searchInputButton {
+  height: var(--height);
+  background-color: var(--static-background-background_0-background);
+  border: none;
+  padding: var(--spacings-small);
+  border-bottom-right-radius: 0.75rem;
+  border-top-right-radius: 0.75rem;
+  cursor: pointer;
+  display: flex;
+  align-items: center;
+}
+
 @media (max-width: 650px) {
   .container {
     grid-template-areas:
--- a/src/page-modules/assistant/assistant.module.css
+++ b/src/page-modules/assistant/assistant.module.css
@@ -26,7 +26,7 @@
 .input {
   display: flex;
   flex-direction: column;
-  gap: 6px;
+  gap: 0.375rem;
 }
 
 .button {
@@ -35,7 +35,7 @@
 
 .heading {
   color: var(--static-background-background_accent_0-text);
-  margin: 10px;
+  margin: 0.625rem;
 }
 
 .alternativesWrapper {
--- a/src/page-modules/assistant/assistant.module.css
+++ b/src/page-modules/assistant/assistant.module.css
@@ -1,0 +1,96 @@
+.wrapper {
+  background-color: var(--static-background-background_accent_0-background);
+}
+
+.container {
+  background-color: var(--static-background-background_accent_0-background);
+  height: 100%;
+  position: relative;
+  display: grid;
+  grid-template-areas:
+    'main'
+    'alternatives';
+}
+
+.main {
+  grid-area: main;
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  grid-gap: var(--spacings-xLarge);
+  width: 100%;
+  max-width: var(--maxPageWidth);
+  padding: var(--spacings-xLarge);
+  margin: 0 auto;
+}
+
+.input {
+  display: flex;
+  flex-direction: column;
+  gap: 6px;
+}
+
+.button {
+  text-align: center;
+}
+
+.heading {
+  color: var(--static-background-background_accent_0-text);
+  margin: 10px;
+}
+
+.alternativesWrapper {
+  grid-area: alternatives;
+  width: 100%;
+  background-color: var(--static-background-background_accent_1-background);
+  padding-bottom: 5.25rem;
+}
+
+.alternatives {
+  padding: var(--spacings-xLarge);
+  width: 100%;
+  max-width: var(--maxPageWidth);
+  margin: 0 auto;
+}
+
+.buttons {
+  grid-area: alternatives;
+  align-self: end;
+  display: flex;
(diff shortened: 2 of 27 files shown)


src/page-modules/assistant/assistant.module.css
```diff
@@ -0,0 +1,96 @@
...
+  gap: 6px;
+}
+
+.button {
+  text-align: center;
+}
+
+.heading {
+  color: var(--static-background-background_accent_0-text);
+  margin: 10px;
```
adriansberg, line 38: `rem` or the spacings variables. `10px` isn't one of the set variables IIRC, so not sure how we should handle that.

src/page-modules/assistant/layout.tsx
```diff
@@ -0,0 +1,222 @@
...
+        ...query,
+        arriveBy,
+      };
+    }
+
+    const when = departureDate.type !== DepartureDateState.Now;
+    if (when) {
+      query = {
+        ...query,
+        when: departureDate.dateTime,
```
adriansberg, line 74: Is both `when` and `arriveBy` needed if e.g. `departureDate.type === DepartureDateState.Arrival`? Instead of `arriveBy` _or_ `departBy` that is.

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -0,0 +1,265 @@
...
+          numTripPatterns: 10,
+          modes,
+        },
+      });
+
+      if (result.error) {
+        throw result.error;
+      }
+
+      const data: RecursivePartial<TripData> = {
```
adriansberg, line 72: I assume this kind of parsing can't be done with zod?

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -0,0 +1,265 @@
...
+type RecursivePartial<T> = {
+  [P in keyof T]?: T[P] extends (infer U)[]
+    ? RecursivePartial<U>[]
+    : T[P] extends object | undefined
+    ? RecursivePartial<T[P]>
+    : T[P];
+};
+
+function mapSituations(
+  situations: TripsQuery['trip']['tripPatterns'][0]['legs'][0]['situations'][0][],
```
adriansberg, line 222: Got no type for `situations`?

adriansberg:
```ts
export type Situation = z.infer<typeof situationSchema>;
```

This is not the same Situation?

src/pages/assistant/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+          >
+            {tripPattern.legs.map((leg, i) => (
+              <div
+                key={`leg-${leg.expectedStartTime}-${i}`}
+                className={style.leg}
+              >
+                <span>{leg.mode ? leg.mode : 'foot'}</span>
+                <span>{leg.line?.publicCode}</span>
+                <span>
+                  {leg.fromPlace.name} til {leg.toPlace.name}
```
adriansberg, line 39: Needs to be translated. Specifically the "til".

src/page-modules/assistant/layout.tsx
adriansberg: Ah, see that `arriveBy` is a boolean now. But the question sort of still stands.

src/pages/assistant/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+};
+
+export default AssistantPage;
+
+export const getServerSideProps = withGlobalData(
+  withAssistantClient<AssistantLayoutProps & AssistantContentProps>(
+    async function ({ client, query }) {
+      const transportModeFilter = parseFilterQuery(query.filter);
+
+      let initialFromFeature = undefined;
```
adriansberg, line 75:
```suggestion
      let initialFromFeature;
```

src/pages/assistant/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+      if (query.fromLat && query.fromLon && query.fromLayer) {
+        const fromLat = parseFloat(query.fromLat.toString());
+        const fromLon = parseFloat(query.fromLon.toString());
+        const fromLayer = parseLayerQueryString(query.fromLayer.toString());
+
+        initialFromFeature = await client.reverse(fromLat, fromLon, fromLayer);
+      }
+
+      let initialToFeature = undefined;
```
adriansberg, line 84:
```suggestion
      let initialToFeature;
```

src/pages/assistant/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+          <div
+            key={`tripPattern-${tripPattern.expectedStartTime}-${i}`}
+            className={style.tripPattern}
+          >
+            {tripPattern.legs.map((leg, i) => (
+              <div
+                key={`leg-${leg.expectedStartTime}-${i}`}
+                className={style.leg}
+              >
+                <span>{leg.mode ? leg.mode : 'foot'}</span>
```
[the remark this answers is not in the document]
adriansberg: This is probably temporary.

src/pages/assistant/index.tsx
adriansberg: This is probably temporary.

adriansberg:
Is there a limit to how far into the future trips can be looked up? For instance if I search from my current location to e.g. Trondheim S, I don't get any results because the next bus departure isn't until 3-4 hours or so from now. Wouldn't it be better to get the trips either way?

src/translations/pages/assistant.ts
```diff
@@ -0,0 +1,39 @@
+import { translation as _ } from '@atb/translations/commons';
+
+export const Assistant = {
+  title: _('Planlegg reisen', 'Plan travel', 'Planlegg reisen'),
+  shortTitle: _('Reiseplanlegger', 'Assistant', 'Reiseplanlegger'),
```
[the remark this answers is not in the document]
mortennordseth: I can't open the image 🤔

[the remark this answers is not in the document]
adriansberg: image

src/pages/assistant/assistant.module.css
```diff
@@ -0,0 +1,11 @@
+.tripPattern {
+  display: flex;
+  flex-direction: column;
+  padding: var(--spacings-medium);
+  border-bottom: 1px solid black;
```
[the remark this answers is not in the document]
mortennordseth: This is temporary until we start working on the trip data listing. I can see that it's confusing and might not have been pushed, but I thought it was nice for testing the trip search

src/page-modules/assistant/server/journey-planner/index.ts
mortennordseth: The difference is that some of the values in `TripsQuery['trip']['tripPatterns'][0]['legs'][0]['situations'][0][]` can be `undefined` while the same values in `Situation` cannot (`null` instead). I thought about creating a second type, e.g. `GraphQLSituation`, but if the auto-generated type changes we would have to update that one as well. I don't think that is a good approach either 🤔

src/components/search/index.tsx
```diff
@@ -51,10 +54,14 @@ export default function Search({
...
-            onGeolocate={selectItem}
-          />
+          {onSwap ? (
+            <SwapButton className={style.geolocationButton} onSwap={onSwap} />
+          ) : (
+            <GeolocationButton
+              className={style.geolocationButton}
+              onGeolocate={selectItem}
+            />
+          )}
```
mikaelbr, line 64: I think maybe a better approach here is to be more generic in the API. Pass inn a button component from parent for instance. 🤔

src/page-modules/assistant/layout.tsx
```diff
@@ -0,0 +1,222 @@
...
+      <form className={style.container} onSubmit={onSubmitHandler}>
+        <div className={style.main}>
+          <div className={style.input}>
+            <Typo.p textType="body__primary--bold" className={style.heading}>
+              {t(PageText.Assistant.search.input.label)}
+            </Typo.p>
+
+            <Search
+              label={t(PageText.Assistant.search.input.from)}
+              onChange={setSelectedFromFeature}
```
mikaelbr, line 145: Should this instead change the URL? Or doesn't that make sense in this case?

src/page-modules/assistant/layout.tsx
```diff
@@ -0,0 +1,222 @@
...
+              initialQuery={
+                initialFromFeature?.name
+                  ? `${initialFromFeature.name}, ${initialFromFeature.locality}`
+                  : ''
```
mikaelbr, line 149: Could also be a named function for easier ways to read

src/page-modules/assistant/server/journey-planner/index.ts
adriansberg: Hmm, okay.

src/page-modules/assistant/layout.tsx
adriansberg: Is it necessary to change the URL before the "Find departures" button is clicked (or form submitted)?

mikaelbr: No, that's a good point!

src/page-modules/assistant/server/journey-planner/index.ts
mortennordseth: I haven't found anything from their documentation that suggests so 🤔

adriansberg: I figured. 👍

adriansberg (changes requested):
Couple of comments that need to be addressed before I approve.

src/components/search/index.tsx
mortennordseth: Good idea! I've changed this now 👍

src/components/search/index.tsx
```diff
@@ -1,35 +1,34 @@
...
 }: SearchProps) {
-  const [query, setQuery] = useState(initialQuery);
+  const [query, setQuery] = useState('');
   const { data } = useAutocomplete(query);
 
   return (
     <Downshift<GeocoderFeature>
-      initialInputValue={query}
+      initialSelectedItem={initialFeature}
+      initialInputValue={geocoderFeatureToString(initialFeature)}
```
adriansberg, line 28: Is this needed in addition to the `itemToString`?

src/page-modules/assistant/layout.tsx
mortennordseth: I've made some changes now to reflect what we discussed @adriansberg . Take a look and see if it's something along the lines you were thinking. It's still an issue with updating the `inputValue`, but hopefully I'll find a solution for that soon.

src/page-modules/departures/departures.module.css
```diff
@@ -33,6 +32,18 @@
   margin-bottom: var(--spacings-medium);
 }
 
+.geolocationButton {
+  height: var(--height);
+  background-color: var(--static-background-background_0-background);
+  border: none;
+  padding: var(--spacings-small);
+  border-bottom-right-radius: 0.75rem;
```
adriansberg, line 40: I know you just moved this code, but isn't one of the spacings 0.75rem?

src/page-modules/departures/layout.tsx
```diff
@@ -84,6 +85,14 @@ function DeparturesLayout({
             <Search
               label={t(PageText.Departures.search.input.from)}
               onChange={setSelectedFeature}
+              button={
+                <GeolocationButton
+                  className={style.geolocationButton}
+                  onGeolocate={(feature: GeocoderFeature) => {
+                    setSelectedFeature(feature);
+                  }}
```
adriansberg, line 93:
```suggestion
                  onGeolocate={setSelectedFeature}
```

src/page-modules/assistant/layout.tsx
adriansberg: Looks like we need to either make the Search controlled, and set the state when routing, or trigger a reload of the page when routing such that the initialFeature is used. Not sure what's the best solution here.

src/components/search/index.tsx
mortennordseth: No 😄

src/page-modules/departures/departures.module.css
mortennordseth: It does for spacings, but not border-radius 🤔

adriansberg: Right! That's why I hadn't used a variable in the first place.

src/page-modules/assistant/server/journey-planner/index.ts
mortennordseth: What do you think? Is it better to create a new type or do you see an alternative solution to this?

adriansberg: I'm not a huge fan of the indexed type, but I see the arguments for not creating a second type  that needs to be in sync. @mikaelbr thoughts?

src/page-modules/assistant/layout.tsx
mortennordseth: We could skip the `arriveBy` when it's false. The date can also be set server side, so if it is not sent we use "now". I'm not sure if leaving out those parameters affects anything else or if it improves the code though 🤔

adriansberg: How does the querying deal with "depart by"?

mortennordseth:
> Is there a limit to how far into the future trips can be looked up? For instance if I search from my current location to e.g. Trondheim S, I don't get any results because the next bus departure isn't until 3-4 hours or so from now. Wouldn't it be better to get the trips even though they are some time away?

One should be able to get trips even though they are some time away. I haven't been able to reproduce this issue. For example, searching from Molde trafikkterminal to Molde sjukehus with departure on saturday gives me results: 

image

src/page-modules/assistant/layout.tsx
mortennordseth: That would be `arriveBy = false` and `when` the date you would like to depart by. Similar to "now", but with different date

adriansberg:
> > Is there a limit to how far into the future trips can be looked up? For instance if I search from my current location to e.g. Trondheim S, I don't get any results because the next bus departure isn't until 3-4 hours or so from now. Wouldn't it be better to get the trips even though they are some time away?
> 
> One should be able to get trips even though they are some time away. I haven't been able to reproduce this issue. For example, searching from Molde trafikkterminal to Molde sjukehus with departure on saturday gives me results:
> 
> image

I don't think those stops are the best to use in an example.

For example, my nearest stop place to Stjørdal stasjon, "Now". The next bus is departing in hour and a half. It is not showing when searching with "Now":

image

If I set it to "Arrival" and set the time to 17:07 I get a trip that arrives before that time.

image

src/page-modules/assistant/layout.tsx
mortennordseth: Alright, I've updated this to use the `arriveBy` and `departBy` values.

src/page-modules/assistant/server/journey-planner/index.ts
mikaelbr: There is a lot of mapping here that is mostly undefined/defined stuff. Feels like this is something that we could do with `zod` because I don't think all of these should actually be nullable? But I think that can also just be noted as a future improvement.

src/page-modules/assistant/layout.tsx
mortennordseth: I agree! Good suggestion 🎉

mortennordseth:
> For example, my nearest stop place to Stjørdal stasjon, "Now". The next bus is departing in hour and a half. It is not showing when searching with "Now":

I've been looking into this now and there is a max time ahead that the journey planner API returns trip patterns for. The max time ahead seems to be 1 hour and 20 minutes. In the OMS-app we solve this by looking up the next trip patterns if a result is empty until we have a certain amount of trip patterns. When searching for the next trip patterns we use the same query but include the next-page-cursor.

We need to add this functionality as well, but maybe I should create a separate PR for that? 🤔 Let me know what you think, @adriansberg

adriansberg:
> > For example, my nearest stop place to Stjørdal stasjon, "Now". The next bus is departing in hour and a half. It is not showing when searching with "Now":
> 
> 
> 
> I've been looking into this now and there is a max time ahead that the journey planner API returns trip patterns for. The max time ahead seems to be 1 hour and 20 minutes. In the OMS-app we solve this by looking up the next trip patterns if a result is empty until we have a certain amount of trip patterns. When searching for the next trip patterns we use the same query but include the next-page-cursor.
> 
> 
> 
> We need to add this functionality as well, but maybe I should create a separate PR for that? 🤔 Let me know what you think, @adriansberg 

Sure, can definitely be addressed in a separate PR. It'd be nice if it was addressed at least.

mikaelbr:
Might have been commented before, but should be some min width on label so from/to search is placed on the same line
Screenshot 2023-10-26 at 08 33 38

src/page-modules/assistant/layout.tsx
mortennordseth: I've updated to use `selectedItem` so that the input is updated correctly without having to reload the page

adriansberg (approved):

mortennordseth:
All of the issues that we found should have been fixed now, except for handling the search when there are no trip patterns that match the search in the next 1.3 hours. I'll address that in a separate PR

adriansberg (approved):
Great work!

mikaelbr:
Great

PR #40, opened 2023-10-25
style: add animation to departure date selector

adriansberg:
Added `framer-motion` as a dependency and added some animation to the departure date selector.

2023-10-25 at 15 00 30


mortennordseth (approved):
✨

PR #41, opened 2023-10-26
style: add animation to alternatives section

adriansberg:
Adds enter/exit animation to the alternatives section.

<p>

2023-10-26 at 11 25 34
2023-10-26 at 11 26 01

</p>
</details>


src/page-modules/departures/layout.tsx
mortennordseth, line 1: Should we add this to the assistant page too?

adriansberg: Yes! Added the animation there as well now.

mortennordseth (approved):
As you mentioned, we should probably refactor some parts of the code to clean up duplicate code etc., but I think this looks good for now! Good job ⭐

PR #42, opened 2023-10-29
feat: show trip patterns / travel suggestions

adriansberg:
Added initial version of list of travel suggesions based on the search.

<p>
image

image

image

</p>
</details> 

Fixes https://github.com/AtB-AS/kundevendt/issues/11319 (partly at least)

### TODO

- [ ] Mobile view
- [ ] Accessiblity
- [ ] Show only three first legs
- [ ] Unit tests
- [ ] Improve lines implementation


src/components/transport-mode/transport-icon/index.tsx
```diff
@@ -42,6 +42,34 @@ export function TransportIcon({ mode }: TransportIconProps) {
...
+  return (
+    <span className={style.transportIconWithLabel} style={{ backgroundColor }}>
+      <MonoIcon
+        icon={getTransportModeIcon(mode)}
+        role="img"
+        alt={t(transportModeToTranslatedString(mode))}
+        overrideMode={overrideMode}
+      />
+      {label && (
+        <span style={{ color: overrideMode === 'light' ? 'black' : 'white' }}>
```
mortennordseth, line 65: I think we should use the `text-colors-primary` color for the label here 👍

src/page-modules/assistant/server/journey-planner/validators.ts
```diff
@@ -1,5 +1,6 @@
 import { TransportModeType, TransportSubmodeType } from '@atb-as/config-specs';
 import { z } from 'zod';
+import { TripPattern } from '@atb/page-modules/assistant/trip-pattern';
```
mortennordseth, line 3:
```suggestion
```

mortennordseth (approved):
I think this looks very good! 💯  I've added some small comments and I'm okay with merging this now, fixing the things you've listed as `TODO` in another PR 👍

src/page-modules/assistant/trip-pattern/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  index: number;
+};
+export function TripPattern({ tripPattern, index }: TripPatternProps) {
+  const { t, language } = useTranslation();
+
+  const duration = secondsToDuration(tripPattern.duration, language);
+  const fromPlace = tripPattern.legs[0].fromPlace.name ?? '';
+
+  return (
+    <motion.div
```
mikaelbr, line 22: I think actually this should be an anchor tag that wraps the entire element here. In modern HTML using `div` inside `a` is OK as `a` has a transparent content model (so everything but interactive elements essentially). And this should be links instead of buttons as we do a navigation. This would also make the entire element focusable / clickable

src/components/transport-mode/transport-icon/transport-icon.module.css
```diff
@@ -6,15 +6,24 @@
   gap: 0.125rem;
 }
 
-.transportIcon {
+.transportIcon,
+.transportIconWithLabel {
   display: inline-block;
 
   /* Specialized sizing. */
   /* @TODO investigate generic way to handle this */
```
mikaelbr, line 14: Comment no longer true?
```suggestion
```

src/page-modules/assistant/trip-pattern/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+type TripPatternProps = {
+  tripPattern: TripPattern;
+  index: number;
+};
+export function TripPattern({ tripPattern, index }: TripPatternProps) {
+  const { t, language } = useTranslation();
+
+  const duration = secondsToDuration(tripPattern.duration, language);
+  const fromPlace = tripPattern.legs[0].fromPlace.name ?? '';
```
mikaelbr, line 19: Can this contain no legs? So this would cause fromPlace of undefined error

src/page-modules/assistant/trip-pattern/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+          >
+            <div className={style.legs__leg__icon}>
+              {leg.mode ? (
+                <TransportIconWithLabel
+                  mode={{ mode: leg.mode }}
+                  label={leg.line?.publicCode}
+                />
+              ) : (
+                <div className={style.legs__leg__walkIcon}>
+                  <MonoIcon icon={'transportation/Walk'} />
```
mikaelbr, line 57:
```suggestion
                  <MonoIcon icon="transportation/Walk" />
```

src/page-modules/assistant/trip-pattern/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+              {formatLocaleTime(tripPattern.expectedEndTime, language)}
+            </Typo.span>
+          </div>
+        ))}
+
+        <div className={style.legs__lastLeg}>
+          <div className={style.legs__lastLeg__line}></div>
+          <div className={style.legs__lastLeg__destination}>
+            <div className={style.legs__lastLeg__destination__icon}>
+              <MonoIcon icon={'places/Destination'} />
```
mikaelbr, line 72:
```suggestion
              <MonoIcon icon="places/Destination" />
```

src/page-modules/assistant/trip-pattern/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+            </Typo.span>
+          </div>
+        </div>
+      </div>
+
+      <footer className={style.footer}>
+        <Button
+          title={t(ComponentText.TripPattern.details)}
+          size="compact" // TODO: no padding with this size, should there be some? Makes hover look weird.
+          icon={{ right: <MonoIcon icon={'navigation/ArrowRight'} /> }}
```
mikaelbr, line 85:
```suggestion
          icon={{ right: <MonoIcon icon="navigation/ArrowRight" /> }}
```

src/components/transport-mode/transport-icon/index.tsx
adriansberg: That won't work. The label needs to be the same color as the icon and can't change depending on the set theme. 

I'll add this to the todo, to figure out how to solve it better.

src/page-modules/assistant/trip-pattern/index.tsx
adriansberg: Will implement that. 👍

src/components/transport-mode/transport-icon/transport-icon.module.css
adriansberg: Sort of. I've deviated from the design here I see, and used a variable. We should go through the design and the implementations and look at stuff like this. I think it's preferrable to use set variables.

src/page-modules/assistant/trip-pattern/index.tsx
adriansberg: That might very well be. Adding a `?.` to avoid that at least. Should probably be handled and not showing a trip pattern at all if that's the case? Have you experienced any trip patterns with no legs, @mortennordseth?

mortennordseth: Not that I can think of 🤔

src/page-modules/assistant/trip-pattern/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+                  <MonoIcon icon="transportation/Walk" />
+                </div>
+              )}
+            </div>
+
+            <Typo.span textType="body__tertiary">
+              {formatLocaleTime(tripPattern.expectedEndTime, language)}
+            </Typo.span>
+          </div>
+        ))}
```
mikaelbr, line 66: Maybe this should be wrapped in a container which has `overflow: hidden`? Think this will solve most cases on smaller screens 🤔

src/translations/components/trip-pattern.ts
mikaelbr, line 1: If trip-patterns is a shared component it should be in component folder and not page-modules/assistant. if it should be limited to `page-modules/assistant` this Translation file should be part of Assistant in some way. Hierarchy of translations should reflect the hierarchy of modules/components/pages

adriansberg: Was jumping back on forth on this, unsure where to put it. Moved it to `page-modules/assistant` now.

src/page-modules/assistant/trip-pattern/index.tsx
adriansberg: Can do that, taking that into account when optimizing for mobile. According to the design we shall only show the 3(?) first legs as well, and truncate the rest into a count only.

src/page-modules/assistant/trip-pattern/index.tsx
adriansberg: Added optional chaining to avoid error. Should look into handling no legs if that's ever the case.

src/page-modules/assistant/trip-pattern/index.tsx
mikaelbr: Is this always true no matter size? Is it simplification or just practical? Should talk to Julia I guess

adriansberg: Should talk to Julia, yes.

mikaelbr (approved):
Ok for now and followup on TODOs in separate PR

PR #44, opened 2023-10-30
feat: adds non-transit data to trip patterns.

mikaelbr:
Adds non-transit search results and removes walking from original trip result.

## Notes

Not happy about `TransportModeType`, `TransportSubmodeType` usage inside components. Want to refactor, but seeing if that could be separate PR.

Assistant page should be handled better but see that that work is done in #45 , so leaving as is for now.

Fixes https://github.com/AtB-AS/kundevendt/issues/11320

<p>

Screenshot 2023-10-31 at 08 42 05

</p>
</details>

### DIFF
--- a/src/components/transport-mode/types.ts
+++ b/src/components/transport-mode/types.ts
@@ -1,4 +1,27 @@
-import { TransportModeType, TransportSubmodeType } from '@atb-as/config-specs';
+import { TransportSubmodeType } from '@atb-as/config-specs';
+import { z } from 'zod';
+
+export const TransportModeType = z.union([
+  z.literal('air'),
+  z.literal('bus'),
+  z.literal('foot'),
+  z.literal('bicycle'),
+  z.literal('cableway'),
+  z.literal('coach'),
+  z.literal('funicular'),
+  z.literal('lift'),
+  z.literal('metro'),
+  z.literal('monorail'),
+  z.literal('rail'),
+  z.literal('taxi'),
+  z.literal('tram'),
+  z.literal('trolleybus'),
+  z.literal('unknown'),
+  z.literal('water'),
+]);
+export type TransportModeType = z.infer<typeof TransportModeType>;
+
+export { TransportSubmodeType };
 
 export type TransportModeGroup = {
   mode: TransportModeType;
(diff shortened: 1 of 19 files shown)


src/page-modules/assistant/non-transit-pill/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  tp: TripPattern,
+  t: TranslateFunction,
+): { mode: TransportModeType; modeText: string } => {
+  let mode = tp.legs[0].mode ?? 'unknown';
+  let text = t(PageText.Assistant.nonTransit.unknown);
+
+  if (tp.legs.some((leg) => leg.rentedBike)) {
+    mode = 'bicycle';
+    text = t(PageText.Assistant.nonTransit.bikeRental);
+  } else if (tp.legs[0].mode === 'foot') {
```
mortennordseth, line 52:
```suggestion
  } else if (mode === 'foot') {
```

src/page-modules/assistant/non-transit-pill/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+    text = t(PageText.Assistant.nonTransit.foot);
+  } else if (tp.legs[0].mode === 'bicycle') {
```
mortennordseth, line 54:
```suggestion
  } else if (mode === 'bicycle') {
```

mortennordseth (approved):
Nice 👍 I agree with your note about the transport mode types and that it can be refactored in a separate PR

src/components/transport-mode/types.ts
```diff
@@ -1,4 +1,27 @@
-import { TransportModeType, TransportSubmodeType } from '@atb-as/config-specs';
+import { TransportSubmodeType } from '@atb-as/config-specs';
+import { z } from 'zod';
+
+export const TransportModeType = z.union([
```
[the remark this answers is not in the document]
mikaelbr: I think that we should convert from external types to internal types to avoid having them spread around. This PR is kind of an example of the effect when we use external types without the ability to map/convert/adapt. So using external types to verify the "edges of the application" but use "safe type inside" in a way.

PR #45, opened 2023-10-30
feat: load more trip patterns

mortennordseth:
This PR adds functionality for loading more trip patterns for a trip. I've updated some of the types, but I'm sure these can be improved. 

<details>
<summary>Recording</summary>

https://github.com/AtB-AS/planner-web/assets/43166974/93fbffbb-22ff-4f85-b78c-68e7db1d1c5e

</details>

Fixes sub-task "Load more trip patterns button" in https://github.com/AtB-AS/kundevendt/issues/11319


src/page-modules/assistant/trip/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+  return (
+    <>
+      {tripPatterns.map((tripPattern, i) => (
+        <TripPattern
+          key={`tripPattern-${tripPattern.expectedStartTime}-${i}`}
+          tripPattern={tripPattern}
+          delay={tripPattern.transitionDelay}
+        />
+      ))}
+      <Button
```
mortennordseth, line 87: The button type used in Figma doesn't exist and this should be changed after https://github.com/AtB-AS/kundevendt/issues/12072 is done

src/page-modules/assistant/trip/index.tsx
```diff
@@ -0,0 +1,177 @@
+import { formatLocaleTime, secondsToDuration } from '@atb/utils/date';
```
mortennordseth, line 1: I renamed the `trip-pattern` folder to `trip` because I thought the `Trip` and `TripPattern` components could co-exist here

src/page-modules/assistant/types.ts
```diff
@@ -1,10 +1,48 @@
...
+  toLayer: z.union([z.literal('address'), z.literal('venue')]),
+  filter: z.string().optional(),
+  arriveBy: z.string().optional(),
+  departBy: z.string().optional(),
+  cursor: z.string().optional(),
+});
+
+export type TripQuery = z.infer<typeof TripQuerySchema>;
+
+export type TripQueryObject = {
```
mortennordseth, line 36: I'm struggling a bit with the namings here because the difference is really just that `TripQuery` has a lot of strings, while the `TripQueryObject` does have numbers as well. We might not need two types, but couldn't see a better approach to handle this right now

src/page-modules/assistant/client/journey-planner/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  cursor: string,
+): Promise<TripApiReturnType> {
+  const queryString = Object.keys(query)
+    .map(
+      (key) =>
+        encodeURIComponent(key) +
+        '=' +
+        encodeURIComponent(String(query[key as keyof TripQueryObject])),
+    )
+    .join('&');
```
mikaelbr, line 17: I think you can do this easier:

```suggestion
  const queryString = new URLSearchParams(query).toString()
```

mortennordseth: I would need to change the `TripQueryObject` to only contain strings to do so, but that makes sense. No need for duplicate types really. I'll update it 👍

src/page-modules/assistant/trip/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+    setTripPatterns([
+      ...tripPatterns,
+      ...tripPatternsWithTransitionDelay(trip.tripPatterns),
+    ]);
+    departureType === 'arrival'
+      ? setPreviousPageCursor(trip.previousPageCursor)
+      : setNextPageCursor(trip.nextPageCursor);
+
+    setIsFetchingTripPatterns(false);
+  };
```
mikaelbr, line 77: I think this can be extracted as a custom hook to clean up the function here and make it more easy to read.

src/page-modules/assistant/trip/index.tsx
mikaelbr: Are we sure the design isn't the one needing to change? 🤔

src/page-modules/assistant/trip/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+    </>
+  );
+}
+
+type TripPatternProps = {
+  tripPattern: TripPattern;
+  delay: number;
+};
+
+export function TripPattern({ tripPattern, delay }: TripPatternProps) {
```
mikaelbr, line 102: Do we need to export this? Trying to keep only one exported component per file

src/page-modules/assistant/utils.ts
```diff
@@ -1,38 +1,50 @@
...
+  departureDate?: DepartureDate,
+): TripQuery => {
+  const filter =
+    transportModeFilter && filterToQueryString(transportModeFilter);
+  const arriveBy = departureDate?.type === DepartureDateState.Arrival;
+  const departBy = departureDate?.type === DepartureDateState.Departure;
+  const fromToQuery = featuresToFromToQuery(fromFeature, toFeature);
+
+  return {
+    ...(filter ? { filter } : {}),
```
mikaelbr, line 39: Isn't this the same? 🤔
```suggestion
    filter,
```

src/pages/api/assistant/trip.ts
```diff
@@ -0,0 +1,59 @@
+import { parseFilterQuery } from '@atb/components/transport-mode-filter/utils';
```
mikaelbr, line 1: I don't think we should use utils from components as such. Trying to keep responsibilities clear, so either transport mode should be a module with components or utils should be extracted in some way.

src/pages/api/assistant/trip.ts
```diff
@@ -0,0 +1,59 @@
...
+import { errorResultAsJson, tryResult } from '@atb/modules/api-server';
+import { TripApiReturnType } from '@atb/page-modules/assistant/client/journey-planner';
+import { handlerWithAssistantClient } from '@atb/page-modules/assistant/server';
+import { parseTripQuery } from '@atb/page-modules/assistant/utils';
```
mikaelbr, line 5: Same as above. But maybe here it could be exported by `@atb/page-modules/assistant` and that will be ok

src/pages/api/assistant/trip.ts
```diff
@@ -0,0 +1,59 @@
...
+        parseFloat(tripQuery.toLat),
+        parseFloat(tripQuery.toLon),
+        tripQuery.toLayer,
+      );
+      const transportModeFilter = parseFilterQuery(tripQuery.filter);
+
+      const arriveBy = 'arriveBy' in tripQuery;
+      const departureDate = arriveBy
+        ? new Date(Number(tripQuery.arriveBy))
+        : new Date(Number(tripQuery.departBy)) || new Date();
```
mikaelbr, line 28: Maybe this logic should happen inside `client.trip` 🤔

src/pages/api/assistant/trip.ts
```diff
@@ -0,0 +1,59 @@
...
+import { TripApiReturnType } from '@atb/page-modules/assistant/client/journey-planner';
+import { handlerWithAssistantClient } from '@atb/page-modules/assistant/server';
+import { parseTripQuery } from '@atb/page-modules/assistant/utils';
+import { ServerText } from '@atb/translations';
+import { constants } from 'http2';
+
+export default handlerWithAssistantClient<TripApiReturnType>({
+  async GET(req, res, { client, ok }) {
+    const tripQuery = parseTripQuery(req.query);
+    if (tripQuery) {
```
mikaelbr, line 12: Could reduce nesting with early exit here. I think this will make it easier to follow the code further down as you don't have to remember the context in the same way:

```suggestion
    if (!tripQuery) {
      return errorResultAsJson(
        res,
        constants.HTTP_STATUS_BAD_REQUEST,
        ServerText.Endpoints.invalidData,
      );
    }
```

src/pages/api/assistant/trip.ts
```diff
@@ -0,0 +1,59 @@
...
+        tripQuery.toLayer,
+      );
+      const transportModeFilter = parseFilterQuery(tripQuery.filter);
+
+      const arriveBy = 'arriveBy' in tripQuery;
+      const departureDate = arriveBy
+        ? new Date(Number(tripQuery.arriveBy))
+        : new Date(Number(tripQuery.departBy)) || new Date();
+
+      if (from && to) {
```
mikaelbr, line 30: Same comment on early exit here.

src/pages/assistant/index.tsx
```diff
@@ -47,50 +41,46 @@ export default AssistantPage;
...
-          },
-        };
-      }
+        if (from && to) {
+          const trip = await client.trip({
+            from,
+            to,
+            ...(arriveBy
+              ? { arriveBy: departureDate }
+              : { departBy: departureDate }),
```
mikaelbr, line 69: Having this multiple places feels like we should do a better abstraction. Like for instance:

```
searchDate: { departureDate, mode: 'arriveBy' }
```

And instead of using boolean all around we have modelled named entities.

src/pages/assistant/index.tsx
```diff
@@ -47,50 +41,46 @@ export default AssistantPage;
...
+            transportModes: transportModeFilter,
+          });
 
+          return {
+            props: {
+              initialFromFeature: from,
+              initialToFeature: to,
+              initialTransportModesFilter: transportModeFilter,
+              trip,
+              departureType: arriveBy ? 'arrival' : 'departure',
```
mikaelbr, line 79: So in stead of having to convert here, it should be consistent all around

src/pages/assistant/index.tsx
```diff
@@ -47,50 +41,46 @@ export default AssistantPage;
...
-      let initialToFeature;
-      if (query.toLat && query.toLon && query.toLayer) {
-        initialToFeature = await client.reverse(
-          parseFloat(query.toLat.toString()),
-          parseFloat(query.toLon.toString()),
-          parseLayerQueryString(query.toLayer.toString()),
+      const tripQuery = parseTripQuery(query);
+      if (tripQuery) {
+        const from = await client.reverse(
+          parseFloat(tripQuery.fromLat),
```
mikaelbr, line 47: What happens here if `fromLat`, `fromLon` etc is not parsable? Meaning if someone changes the URL. And why should we have to parse to float right after parsing the entire query (in `parseTripQuery(query)`)? I would expect after executing `parseTripQuery(query)` everything I needed was safely sanitised and ready to use.

src/page-modules/assistant/utils.ts
mortennordseth: The filter could be null, but I've updated it to be either a string or undefined which makes more sense in this case 👍

src/page-modules/assistant/client/journey-planner/index.ts
```diff
@@ -7,7 +7,14 @@ export async function nextTripPatterns(
...
 ): Promise<TripApiReturnType> {
-  const queryString = new URLSearchParams(query).toString();
+  const queryString = Object.keys(query)
+    .map(
+      (key) =>
+        encodeURIComponent(key) +
+        '=' +
+        encodeURIComponent(String(query[key as keyof TripQuery])),
+    )
+    .join('&');
```
mikaelbr, line 17: If this is necessary due to types I would still create is as a named helper function in the bottom of the file:

```ts
function objectToQuery(input: object): string {
  // ... implementation
}
```

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -55,14 +55,16 @@ export function createJourneyApi(
...
+        ? new Date(input.departureDate)
+        : new Date();
 
       const result = await client.query<TripsQuery, TripsQueryVariables>({
         query: TripsDocument,
         variables: {
           from,
           to,
-          arriveBy: 'arriveBy' in input,
+          arriveBy: input.departureMode === 'arriveBy',
```
mikaelbr, line 67: This is better I think!

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -55,14 +55,16 @@ export function createJourneyApi(
         name: input.to.name,
       };
 
-      const when = 'arriveBy' in input ? input.arriveBy : input.departBy;
+      const when = input.departureDate
+        ? new Date(input.departureDate)
+        : new Date();
```
mikaelbr, line 60: Would it be clearer if this was a Date type all the way through? So String when posting to HTTP request, but inside the application it is Date (serialize / deserialize "on edges of app" as I wrote somewhere else)

src/page-modules/assistant/trip/index.tsx
mortennordseth: Updated this with a custom hook now! 👍

src/page-modules/assistant/client/journey-planner/index.ts
mortennordseth: As far as I can see this is necessary. Makes sense, I'll do that 👍

src/page-modules/assistant/utils.ts
```diff
@@ -1,38 +1,82 @@
...
+
+  const optionalNumericFields = ['departureDate'];
+  optionalNumericFields.forEach((field) => {
+    if (query[field] && typeof query[field] === 'string')
+      query[field] = Number(query[field]);
+  });
+
+  const parsed = TripQuerySchema.safeParse(query);
+  if (!parsed.success) {
+    console.log(parsed.error);
```
mikaelbr, line 78: Don't think we should log to client side 🤔
```suggestion
```

src/page-modules/assistant/client/journey-planner/index.ts
```diff
@@ -0,0 +1,28 @@
+import { TripData } from '@atb/page-modules/assistant/server/journey-planner/validators';
+import { TripQuery } from '@atb/page-modules/assistant/';
```
mikaelbr, line 2: I'm being somewhat particular on these imports now, but I try to follow rules as:

- Locally inside a module: Relative paths.
- Usage from outside module: Absolute path

This is to indicate coupling of files and separate between external files and internal files.

mikaelbr (approved):
Latest changes look good I think! Seems cleaner and less data handling back and forth

src/page-modules/assistant/utils.ts
mortennordseth: The art of debugging... 😄

PR #46, opened 2023-10-31
feat: set assistant as front page

mikaelbr:
Set the frontpage as Assistant page


adriansberg (approved):
This is a nice change, we don't need another front page than this.

PR #47, opened 2023-10-31
ci: fix build

mikaelbr:
Fixes build for multiple orgs

### DIFF
--- a/scripts/build-docker.sh
+++ b/scripts/build-docker.sh
@@ -4,10 +4,14 @@
 # by setting up the environment, building the application, and organizing the output 
 # into a organization specific subdirectiory in the 'dist' directory.
 
+# Clean potential previous builds
+
+rm -rf .next/
+
 mkdir -p dist
 mkdir -p dist/$NEXT_PUBLIC_PLANNER_ORG_ID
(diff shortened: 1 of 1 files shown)


scripts/build-docker.sh
```diff
@@ -4,10 +4,14 @@
...
 
+# Clean potential previous builds
+
+rm -rf .next/
+
 mkdir -p dist
 mkdir -p dist/$NEXT_PUBLIC_PLANNER_ORG_ID
 echo "Running yarn setup && yarn build for $NEXT_PUBLIC_PLANNER_ORG_ID"
-yarn setup $NEXT_PUBLIC_PLANNER_ORG_ID
+yarn setup
```
adriansberg, line 14: Isn't orgId required here?

2023-10-31 at 13 31 17

adriansberg: Is it set in the context of where this script is run possibly

mikaelbr: But `NEXT_PUBLIC_PLANNER_ORG_ID` is set as an environment variable while running that script. Setup in yarn doesn't actually take any argument:


```
    "setup": "generate-assets all $NEXT_PUBLIC_PLANNER_ORG_ID -o ./public/assets/ -ts -ts-o ./src/components/icon",

```

adriansberg (approved):

PR #48, opened 2023-10-31
fix: change color of FRAM's footer logo

adriansberg:


mikaelbr (approved):
Looks good in dark mode and light mode!

PR #49, opened 2023-10-31
feat: base trip pattern header text on mode of first leg

adriansberg:
This PR introduces a valuable new feature which dynamically bases the trip pattern header text on the mode of the first leg of the journey. It seeks to enhance the user experience by providing a clearer, context-aware narrative about the upcoming trip details.

Fixes https://github.com/AtB-AS/kundevendt/issues/14917

### DIFF
--- a/src/page-modules/assistant/__tests__/assistant.test.tsx
+++ b/src/page-modules/assistant/__tests__/assistant.test.tsx
@@ -11,8 +11,7 @@ import { GeocoderFeature } from '@atb/page-modules/departures';
 import { FeatureCategory } from '@atb/components/venue-icon';
 import { GeocoderApi } from '@atb/page-modules/departures/server/geocoder';
 import AssistantLayout from '../layout';
-import { getQuayName, getStartModeAndPlace, NonTransitTripData } from '..';
-import { mockTrip } from './test-data';
+import { NonTransitTripData } from '@atb/page-modules/assistant';
 
 afterEach(function () {
   cleanup();
@@ -126,40 +125,5 @@ describe('assistant page', function () {
 
     expect(submitButton).toBeInTheDocument();
     expect(submitButton).toBeDisabled();
-  });
-
-  it('should get quay name from quay', () => {
-    const quayName = getQuayName({
-      publicCode: '',
-      name: 'Quay',
-      id: 'NSR:Quay:1',
-      situations: [],
-    });
-
-    expect(quayName).toBe('Quay');
-  });
-
-  it('should get quay name with public code from quay', () => {
-    const quayName = getQuayName({
-      publicCode: '1',
-      name: 'Quay',
-      id: 'NSR:Quay:1',
-      situations: [],
-    });
-
-    expect(quayName).toBe('Quay 1');
-  });
-
-  it('should get quay name from trip', () => {
-    const quayName = getQuayName(mockTrip.legs[0].fromPlace.quay);
-
-    expect(quayName).toBe('From 1');
-  });
-
-  it('should get start mode and start place from trip', () => {
-    const { startMode, startPlace } = getStartModeAndPlace(mockTrip);
-
-    expect(startMode).toBe('bus');
-    expect(startPlace).toBe('From 1');
   });
 });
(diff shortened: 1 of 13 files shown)


src/translations/pages/assistant.ts
```diff
@@ -38,7 +39,61 @@ export const Assistant = {
...
+              `Sykkel fra ${place}`,
+              `Bike from ${place}`,
+              `Sykkel frå ${place}`,
+            );
+          case 'foot':
+            return _(
+              `Gå fra ${place}`,
+              `Walk from ${place}`,
+              `Gå frå ${place}`,
+            );
```
mikaelbr, line 92: I think the point of this header should be the first thing you have to look for. So in the app that is things like stop places. I don't think "walk from" make sense as you in most cases know where you are or what you have searched for. So I think the logic should be "find first leg that has a quay".

But the specification of transport mode I think is good!

src/page-modules/assistant/trip/index.tsx
```diff
@@ -136,6 +136,7 @@ function TripPattern({ tripPattern, delay }: TripPatternProps) {
 
   const duration = secondsToDuration(tripPattern.duration, language);
   const fromPlace = tripPattern.legs[0]?.fromPlace.name ?? '';
+  const fromMode = tripPattern.legs[0]?.mode ?? 'unknown';
```
mikaelbr, line 139: So from the comment below this might be `tripPattern.legs.find(l => l.mode != 'foot')` or something. Check in the app how that works

adriansberg: I see, will expand on this then!

src/translations/pages/assistant.ts
adriansberg: Yes, thought it looked weird with "Walk from"

mikaelbr:
This might be the thing that makes sense to have tests on

adriansberg:
> This might be the thing that makes sense to have tests on

Certainly! Have a separate task for tests for the component, but should I add tests for this now instead?

mikaelbr:
I think it makes sense to include with the specific feature. Flow like this: we found an issue/bug, add testcase which ensures that the bug is fixed.

adriansberg:
Added some simple tests. Any functionality you think should be tested?

src/page-modules/assistant/__tests__/assistant.test.tsx
```diff
@@ -11,7 +11,8 @@ import { GeocoderFeature } from '@atb/page-modules/departures';
 import { FeatureCategory } from '@atb/components/venue-icon';
 import { GeocoderApi } from '@atb/page-modules/departures/server/geocoder';
 import AssistantLayout from '../layout';
-import { NonTransitTripData } from '..';
+import { getQuayName, getStartModeAndPlace, NonTransitTripData } from '..';
```
mikaelbr, line 14: I think to keep APIs "clean"™ here, it is better to extract the helpers as its own module and test them instead of exporting them as part of external API.

So index.ts should be external APIs, but these are probably not something you want to export for all to see

src/page-modules/assistant/__tests__/test-data.ts
mikaelbr, line 1: These are often called fixtures. So `trip.fixture.ts`

src/page-modules/assistant/__tests__/assistant.test.tsx
mikaelbr: Or are these a part of the API of the module? 🤔 Might it then be a component?

adriansberg: Can create a component of the header, and contain the functions within that component and have a separate test file for it. Is that what you meant?

mikaelbr: An open question more than anything, but does it make sense to instead of exporting the tools, you just export a component that handles it with an easier API. Low level building blocks vs Ready To Use limited component

src/page-modules/assistant/__tests__/test-data.ts
adriansberg: Yes! Renaming both this and the existing test-data.ts in departures.

src/page-modules/assistant/__tests__/assistant.test.tsx
adriansberg: How about something like this, [dbebd0e](https://github.com/AtB-AS/planner-web/pull/49/commits/dbebd0e1d62e84b10194d4adb6ee5e90531f493f)?

I think a Ready To Use limited component is a good fit for this case.

mikaelbr (approved):
👍

PR #50, opened 2023-10-31
feat: recursive trip pattern search and day label in trip pattern list

mortennordseth:
With this PR we now search for trip patterns until we have found at least 5 trip patterns, with a maximum of 15 attempts. I have also added a day label to the trip patterns list so that it's easier to distinguish day of travel. 

<details>

image

image

image

</details>

## TODO 
- [x] Verify and update the day label look when it's added to the design
- [x] Fix issue with wrong trip patterns when `departureMode` is `arriveBy`

### DIFF
--- a/src/page-modules/assistant/server/journey-planner/index.ts
+++ b/src/page-modules/assistant/server/journey-planner/index.ts
@@ -26,6 +26,16 @@ import {
   isTransportSubmodeType,
 } from '@atb/page-modules/departures/server/journey-planner';
 
+const MIN_NUMBER_OF_TRIP_PATTERNS = 8;
+const MAX_NUMBER_OF_SEARCH_ATTEMPTS = 5;
+const DEFAULT_JOURNEY_CONFIG = {
+  numTripPatterns: 8, // The maximum number of trip patterns to return.
+  waitReluctance: 1.5, // Setting this to a value lower than 1 indicates that waiting is better than staying on a vehicle.
+  walkReluctance: 1.5, // This is the main parameter to use for limiting walking.
+  walkSpeed: 1.3, // The maximum walk speed along streets, in meters per second.
+  transferPenalty: 10, // An extra penalty added on transfers (i.e. all boardings except the first one)
+};
+
 export type JourneyPlannerApi = {
   trip(input: TripInput): Promise<TripData>;
   nonTransitTrips(input: NonTransitTripInput): Promise<NonTransitTripData>;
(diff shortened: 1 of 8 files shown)


src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -122,7 +124,43 @@ export function createJourneyApi(
...
-      const validated = tripSchema.safeParse(data);
+      const trip: RecursivePartial<TripData> = data;
+
+      let tripPatterns = data.tripPatterns || [];
+      let searchAttempt = 1;
+
+      const MINIMUM_NUMBER_OF_TRIP_PATTERNS = 5;
+      const MAX_NUMBER_OF_TRIES = 15;
+
+      while (
```
mikaelbr, line 135: This might be a case for a do-while loop 😱

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -122,7 +124,43 @@ export function createJourneyApi(
         result.data.trip,
       );
 
-      const validated = tripSchema.safeParse(data);
+      const trip: RecursivePartial<TripData> = data;
+
+      let tripPatterns = data.tripPatterns || [];
+      let searchAttempt = 1;
+
+      const MINIMUM_NUMBER_OF_TRIP_PATTERNS = 5;
```
adriansberg, line 132: Is this the same behaviour as in the app, with the retries and minimum number of patterns to fetch?

mortennordseth: I see that the app searches for 8 initial hits with a maximum of 5 chained searches. I can update to use the same config as the app 👍

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -97,21 +97,23 @@ export function createJourneyApi(
...
+        to,
+        arriveBy: input.departureMode === 'arriveBy',
+        when,
+        modes: journeyModes,
+        cursor: input.cursor,
+        numTripPatterns: 10,
+        waitReluctance: 1.5,
+        walkReluctance: 1.5,
+        walkSpeed: 1.3,
+        transferPenalty: 10,
```
mikaelbr, line 111: This isn't a problem with this PR it self but something from earlier. But these might be considered magic numbers and not easy to understand that these are default settings. 

This could be extracted as own constant in the top like:

```
const DEFAULT_JOURNEY_CONFIG = {
    numTripPatterns: 10,
    waitReluctance: 1.5,
    walkReluctance: 1.5,
    walkSpeed: 1.3,
    transferPenalty: 10
}
```

and possibly comments on some of them if needed

mortennordseth: Someone (you) once told me
>  Leave it better than you found it

Words to live by 🤝 Thanks for the reminder

adriansberg (approved):
Looks good 👍

PR #51, opened 2023-11-01
feat: use initial state in departure date selector

adriansberg:
Added the feature to get the initial state for the departure date selector from the query. This ensures that if the page is reloaded or the URL is copied and opened in a new tab etc., the departure date selector will keep the correct state. 

Fixes https://github.com/AtB-AS/kundevendt/issues/15103

### DIFF
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -24,6 +24,7 @@ export type AssistantLayoutProps = PropsWithChildren<{
   initialFromFeature?: GeocoderFeature;
   initialToFeature?: GeocoderFeature;
   initialTransportModesFilter?: TransportModeFilterOption[] | null;
+  initialDepartureMode?: DepartureDate;
 }>;
 
 function AssistantLayout({
@@ -31,6 +32,7 @@ function AssistantLayout({
   initialFromFeature,
   initialToFeature,
   initialTransportModesFilter,
+  initialDepartureMode,
 }: AssistantLayoutProps) {
   const { t } = useTranslation();
   const router = useRouter();
@@ -42,9 +44,11 @@ function AssistantLayout({
   const [selectedToFeature, setSelectedToFeature] = useState<
     GeocoderFeature | undefined
   >(initialToFeature);
-  const [departureDate, setDepartureDate] = useState<DepartureDate>({
-    type: DepartureDateState.Now,
-  });
+  const [departureDate, setDepartureDate] = useState<DepartureDate>(
+    initialDepartureMode ?? {
+      type: DepartureDateState.Now,
+    },
+  );
   const [transportModeFilter, setTransportModeFilter] = useState(
     getInitialTransportModeFilter(initialTransportModesFilter),
   );
--- a/src/page-modules/assistant/types.ts
+++ b/src/page-modules/assistant/types.ts
@@ -11,7 +11,7 @@ export enum DepartureMode {
 export type TripInput = {
   from: GeocoderFeature;
   to: GeocoderFeature;
-  departureMode: DepartureMode;
+  departureMode?: DepartureMode;
   departureDate?: number;
   transportModes?: TransportModeFilterOption[];
   cursor?: string;
@@ -28,7 +28,7 @@ export const TripQuerySchema = z.object({
   toLayer: z.union([z.literal('address'), z.literal('venue')]),
   filter: z.string().optional(),
   departureDate: z.number().optional(),
-  departureMode: z.nativeEnum(DepartureMode),
+  departureMode: z.nativeEnum(DepartureMode).optional(),
   cursor: z.string().optional(),
 });
 
@@ -57,7 +57,7 @@ export enum StreetMode {
 export type NonTransitTripInput = {
   from: GeocoderFeature;
   to: GeocoderFeature;
-  departureMode: DepartureMode;
+  departureMode?: DepartureMode;
   departureDate?: number;
   directModes: StreetMode[];
 };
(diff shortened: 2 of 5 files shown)


src/page-modules/assistant/layout.tsx
```diff
@@ -24,13 +24,15 @@ export type AssistantLayoutProps = PropsWithChildren<{
   initialFromFeature?: GeocoderFeature;
   initialToFeature?: GeocoderFeature;
   initialTransportModesFilter?: TransportModeFilterOption[] | null;
+  initialDepartureMode?: DepartureDate;
```
mortennordseth, line 27: It might be confusing that his is named `DepartureMode` and the type is `DepartureDate` when we have a type that is named `DeperatureMode`. Could change it to `initialDepartureDate` instead?

src/page-modules/assistant/types.ts
```diff
@@ -11,7 +11,7 @@ export enum DepartureMode {
 export type TripInput = {
   from: GeocoderFeature;
   to: GeocoderFeature;
-  departureMode: DepartureMode;
+  departureMode?: DepartureMode;
```
mortennordseth, line 14: Is it necessary that this is optional? I think it might be more explicit and easy to follow if this is either `departBy` or `arriveBy`. No "hidden knowledge" that `departBy` is default

src/pages/assistant/index.tsx
```diff
@@ -80,7 +82,11 @@ export const getServerSideProps = withGlobalData(
               initialToFeature: to,
               initialTransportModesFilter: transportModeFilter,
               trip,
-              departureMode: tripQuery.departureMode,
+              departureMode: tripQuery.departureMode ?? DepartureMode.DepartBy,
+              initialDepartureMode: departureModeToDepartureDate(
```
mortennordseth, line 86: Same comment as above, but here I think it becomes more clear that `departureMode` and `initialDepartureMode` can be confusing, and that `initialDepartureDate` might be easier to follow

src/page-modules/assistant/types.ts
adriansberg: Couple of things with this:

1. Do we want to include the current date when "Now" is set in the URL? Will that affect subsequent searches made?
2. The URL is _very_ long. I thought that by not including the `departureMode` and `departureDate` when "Now" is set could help making it shorter. But we might be at a point where the URL is so long that it doesn't matter.
3. It looked to me that if "Now" is set, the `departureMode` is set to `departBy`. Isn't that "hidden knowledge" the other way?

mortennordseth: 1. I guess we have to decide if we want a linked URL when searched with 'now' should give the same result or not. 
2. I feel you, but I am unsure what is too long. 
3. Good point. The idea was that we shouldn't use the `DepartureDateSelector` component-specific `DepartureDate` type throughout the app. For the rest of the app we really just need to know if it's `departBy` or `arriveBy`. This might be confusing and a better approach is to move the `DepartureDate` type to the assistant page-module.

adriansberg: > I guess we have to decide if we want a linked URL when searched with 'now' should give the same result or not

If we go for "Now" on page load we have to know from the URL whether to do that or not. We can omit the `departureDate` and keep the `departureMode` for instance.

- Arrival: `departureMode=arriveBy&departureDate=1698928046`
- Departure: `departureMode=departAt&departureDate=1698928046`
- Now: `departureMode=departBy`

--- 
> The idea was that we shouldn't use the DepartureDateSelector component-specific DepartureDate type throughout the app.

Can use `DepartureMode` in `pages/assistant.tsx` and move the logic that converts from `DepartureMode` to `DepartureDate` to the component (`departureModeToDepartureDate`). Need to add `departureDate` from the query to the props then to get the time.

mortennordseth: > We can omit the departureDate and keep the departureMode for instance. 

I think this is a good option. We do need `departureMode` when choosing which cursor to use when loading more trip patterns. We could add code to use `nextPageCursor` if `departureMode` is undefined of course, but not sure if that is the best solution. 

> Can use DepartureMode in pages/assistant.tsx and move the logic that converts from DepartureMode to DepartureDate to the component (departureModeToDepartureDate). Need to add departureDate from the query to the props then to get the time.

This sounds good to me!

src/page-modules/assistant/layout.tsx
```diff
@@ -24,13 +28,17 @@ export type AssistantLayoutProps = PropsWithChildren<{
   initialFromFeature?: GeocoderFeature;
   initialToFeature?: GeocoderFeature;
   initialTransportModesFilter?: TransportModeFilterOption[] | null;
+  departureMode?: DepartureMode;
```
mortennordseth, line 31:
```suggestion
  departureMode: DepartureMode;
```

mortennordseth (approved):
One suggested change, but I think the rest looks good! 🎉

src/page-modules/assistant/layout.tsx
adriansberg: This is needed to please TypeScript. Either this or rearrange the types. We might want to do the latter here. The `{ empty: true }` type is unioned with the content type, but when we also have props in the layout it gets difficult. Should probably move the `{ empty: true }` type to be unioned with both the type for content and layout, i.e. `(AssistantContentProps & AssistantLayoutProps) | { empty: true }`. Thoughts?

mortennordseth: Ah! I see. I'm not really sure what is the best option here. I kind of feel like the `{ empty: true }` is more correct to be unioned with the content type, as the content can be empty and the layout still has props 🤔 I might be wrong here and that your suggestion makes more sense.

mortennordseth: Might want to keep `departureMode` optional and look at rearranging the types in a separate issue?

adriansberg: Yes. I think we should have a session regarding types in their entirety. Have gotten confused with graphql types, duplicated types etcetc.

mortennordseth: Yep, it's a bit confusing. Let's have a session later and merge this for now. I think it looks good otherwise. This is also needed to complete https://github.com/AtB-AS/planner-web/pull/50

PR #52, opened 2023-11-03
feat: add trip summary a11y label

adriansberg:
Reused the code for the trip summary from the app. Need to clean up usage of helper functions, usage of types etc.

Example:

image

Result number 1.
Train from Oslo S 13.
Train number FLY1.
At 14:20.
3 transfers.
Total of 382 meters to walk.
Start time 14:20, arrival time 17:19. Total travel time 2 hours, 59 minutes

Fixes https://github.com/AtB-AS/kundevendt/issues/14915


src/translations/dictionary.ts
```diff
@@ -24,6 +24,42 @@ const dictionary = {
     m: _('m', 'm', 'm'),
   },
   a11yRouteTimePrefix: _('rutetid ', 'route time ', `rutetid `),
+  missingRealTimePrefix: _('ca. ', 'ca. ', `ca. `),
```
mortennordseth, line 27: Should this be "approx." in English? 🤔

adriansberg: Apparently "ca." is used in the English language as an abbreviation for "circa", thus can be used in this instance.

src/utils/transportation-names.ts
mortennordseth, line 1: I know it shouldn't and that we are going to refactor the location of the transport mode functions etc, but as it is now, should this function be together with the other similar functions? `/src/components/transport-mode/utils.ts` where we have `transportModeToTranslatedString` and `severalTransportModesToTranslatedStrings`

mortennordseth: I figured that was the reason. I think we should move it so that it at least is consistent, and revisit it as you say 👍

src/page-modules/assistant/trip/utils.ts
```diff
@@ -0,0 +1,256 @@
...
+  secondsToDuration,
+} from '@atb/utils/date';
+import {
+  Leg,
+  TripPattern,
+} from '@atb/page-modules/assistant/server/journey-planner/validators';
+import { Language, TranslateFunction, PageText } from '@atb/translations';
+import dictionary from '@atb/translations/dictionary';
+import { screenReaderPause } from '@atb/components/typography/utils';
+import { getQuayName } from '@atb/page-modules/assistant/trip/trip-pattern-header';
```
mortennordseth, line 13: Relative paths for the imports of `Leg, TripPattern` and `getQuayName` as we discussed on Slack 👍

mortennordseth (approved):
Looks good and works great! 🎉

src/utils/transportation-names.ts
adriansberg: Ended up using the already existing function.

mortennordseth (approved):
Nice tests! 🤩

mikaelbr (approved):

PR #54, opened 2023-11-03
feat: loading indicator when searching for travels and stop places

mortennordseth:
This PR introduces searching state information when searching for travel patterns and stop places. I've added a loading indicator to the search button and a text about loading travel suggestions/stop places/departures according to the design sketches.  

Fixes https://github.com/AtB-AS/kundevendt/issues/14924

TODO:
- [x] See if there is a better approach to this
- [x] Add correct translations
- [x] Add to departure page
- [x] Animate with framer motion

### DIFF
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -24,6 +24,7 @@ import {
 import SwapButton from '@atb/components/search/swap-button';
 import GeolocationButton from '@atb/components/search/geolocation-button';
 import { AnimatePresence, motion } from 'framer-motion';
+import React from 'react';
 
 export type AssistantLayoutProps = PropsWithChildren<{
   initialFromFeature?: GeocoderFeature;
@@ -60,7 +61,7 @@ function AssistantLayout({
   const [isSwapping, setIsSwapping] = useState(false);
   const [isSearching, setIsSearching] = useState(false);
 
-  const onSwap = () => {
+  const onSwap = async () => {
     if (!selectedToFeature || !selectedFromFeature) return;
     setIsSwapping(true);
     const query = createTripQuery(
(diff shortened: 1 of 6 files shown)


src/page-modules/assistant/layout.tsx
```diff
@@ -213,11 +217,20 @@ function AssistantLayout({
             mode="interactive_0"
             disabled={!selectedFromFeature || !selectedToFeature}
             buttonProps={{ type: 'submit' }}
+            state={isSearching ? 'loading' : undefined}
           />
         </div>
       </form>
 
-      <section className={style.contentContainer}>{children}</section>
+      <section className={style.contentContainer}>
```
adriansberg, line 225: Maybe we don't need to swap out the children if there are search results present already?

I like the loading state in the button, and the "Loading search results..." could also be nice, but I don't necessarily think that we need to show it when there are results there.

src/page-modules/assistant/layout.tsx
```diff
@@ -106,7 +108,9 @@ function AssistantLayout({
         : undefined,
       transportModeFilter,
     );
-    router.push({ pathname: '/assistant', query });
+    router
+      .push({ pathname: '/assistant', query })
+      .then(() => setIsSearching(false));
```
adriansberg, line 113: Could `await router.push()` instead, to keep it mostly consistent with the rest, using `async/await` syntax and not `Promise` syntax.

src/page-modules/assistant/layout.tsx
```diff
@@ -97,6 +98,7 @@ function AssistantLayout({
   const onSubmitHandler: FormEventHandler<HTMLFormElement> = (e) => {
     e.preventDefault();
     if (!selectedFromFeature || !selectedToFeature) return;
+    setIsSearching(true);
```
adriansberg, line 101: Should `isSearching()` be set when swapping also?

src/page-modules/assistant/layout.tsx
mortennordseth: I agree! I've updated to reflect this now 👍

src/page-modules/assistant/layout.tsx
mortennordseth: I refactored this so that `isSwapping` is only true when we are actually swapping. Then, when we start the search (i.e. `router.push`) we set `isSearching` to true. Having both buttons spin at the same time was a bit confusing, but I think this solution is more clear 👍

adriansberg (approved):

PR #55, opened 2023-11-04
feat: dynamically collapse legs based on available space

adriansberg:
### Dynamically collapse legs based on available space

_This PR introduces a significant enhancement in leg-handle feature in our system. It adds a dynamic collapsing functionality based on the available space. This attribute adapts to the space available and smartly collapses the leg elements to ensure optimal utilization of the space without compromising on the user's interaction._

_This feature not only optimizes the design for minimal spaces but also enhances the user interface by providing more flexibility and usability, especially in designs with space constraints. Now, the elements on our interface will be optimally adjusted to provide the best user experience in any given viewing area._

I have reused the logic used in the app for dynamically collapsing the legs based on the width available. The implemented logic only collapses the legs however, we might want to look into expanding them as well if the user e.g. shrinks then grows their browser, the wanted effect is probably expanding the previously collapsed legs.

Introduced the `@react-hook/resize-observer` and made a custom hook for getting the width of the elements needed to decide whether legs should collapse or not. Could have not added this hook and gone for a solution that only collapses on load, but I figured it would be nice to collapse on the fly when shrinking as well.

<details><summary>👇 Screenshots and a GIF 👇</summary>
<p>

image

image

2023-11-04 at 23 21 30

</p>
</details> 

Fixes https://github.com/AtB-AS/kundevendt/issues/14914


src/page-modules/assistant/trip/trip-pattern/trip-pattern-header/__tests__/trip-pattern-header.test.tsx
```diff
@@ -2,8 +2,8 @@ import { cleanup, render, screen } from '@testing-library/react';
 import mockRouter from 'next-router-mock';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import { createDynamicRouteParser } from 'next-router-mock/dynamic-routes';
-import { getQuayName, getStartModeAndPlace, TripPatternHeader } from '..';
-import { tripFixture } from './trip.fixture';
+import { getQuayName, getStartModeAndPlace, TripPatternHeader } from '../index';
```
mortennordseth, line 5:
```suggestion
import { getQuayName, getStartModeAndPlace, TripPatternHeader } from '..';
```

src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -0,0 +1,130 @@
+import { useClientWidth } from '@atb/utils/use-client-width';
+import { motion } from 'framer-motion';
+import { useEffect, useState } from 'react';
+import {
+  getFilteredLegsByWalkOrWaitTime,
+  tripSummary,
+} from '@atb/page-modules/assistant/trip/trip-pattern/utils';
```
mortennordseth, line 7:
```suggestion
import { getFilteredLegsByWalkOrWaitTime, tripSummary } from './utils';
```

mortennordseth (commented):
It overflows sometimes when I use the device toolbar in chrome 🤔 

image

image

src/page-modules/assistant/trip/index.tsx
```diff
@@ -1,39 +1,33 @@
...
-import { nextTripPatterns } from '@atb/page-modules/assistant/client';
+import { nextTripPatterns } from '../client';
 import {
   DepartureMode,
   NonTransitTripData,
   createTripQuery,
   filterOutDuplicates,
   getCursorByDepartureMode,
-} from '@atb/page-modules/assistant';
+} from '..';
```
mikaelbr, line 22: Is this now essentially circular dependency? 🤔

adriansberg: Good question 🤔

adriansberg: Thought about this when I moved it all. Maybe the utils.ts should move to the trip folder now?

adriansberg: Looked into moving the utils to `trip/`, but several of the util functions are used in several places, e.g. `layout.tsx` and `assistant/trip/index.tsx`, or `assistant/trip/index.tsx` and `journey-planner/index.tsx`. How should this be solved?

mikaelbr: Thnigs inside the assistant directory shouldn't require things from index, but have index expose the API externally. If its utils I'd place it in a utils file along side index.

adriansberg: I think it is in a utils folder. Removing the export from index. 

It should be rare that a page-module exports something that's used elsewhere, other than in `pages` right?

mikaelbr: Yeah, I think exported things for page-modules should only be the corresponding page. If imported by something else it is not page-module

adriansberg:
> It overflows sometimes when I use the device toolbar in chrome 🤔
> 
> image image

Should be fixed now, please test.

src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+    filteredLegs.length,
+  );
+
+  const [legsParentWidth, legsParentRef] = useClientWidth<HTMLDivElement>();
+  const [legsContentWidth, legsContentRef] = useClientWidth<HTMLDivElement>();
+
+  // Dynamically collapse legs to fit horizontally
+  useEffect(() => {
+    if (legsParentWidth && legsContentWidth) {
+      if (legsContentWidth >= legsParentWidth - LAST_LEG_PADDING) {
```
adriansberg, line 47: Added a `LAST_LEG_PADDING` here to make some space between the destination leg and second to last leg. Couldn't do it in a nice way with CSS without sacrificing the line in some way.

mortennordseth:
I'm getting a lot of errors on the assistant page, but might not be relevant to this PR 🤔 

> framework-63157d71ad419e09.js:9 Uncaught Error: Minified React error #423; visit https://reactjs.org/docs/error-decoder.html?invariant=423 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

> There was an error while hydrating. Because the error happened outside of a Suspense boundary, the entire root will switch to client rendering.

adriansberg:
> I'm getting a lot of errors on the assistant page, but might not be relevant to this PR 🤔 
> 
> 
> 
> > framework-63157d71ad419e09.js:9 Uncaught Error: Minified React error #423; visit https://reactjs.org/docs/error-decoder.html?invariant=423 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
> 
> 
> 
> > There was an error while hydrating. Because the error happened outside of a Suspense boundary, the entire root will switch to client rendering.

Hmm, haven't seen this.

src/page-modules/assistant/trip/trip-pattern/index.tsx
mortennordseth: I'm not seeing a better approach right now, so I'm fine with this 👍

mortennordseth (approved):
Nice work 🤩

PR #56, opened 2023-11-06
refactor: create grapql fragments module

mortennordseth:
I've started looking into the details page (https://github.com/AtB-AS/kundevendt/issues/9159) and I see that some of the fragments are needed for multiple graphql queries. I therefore think we should create a shared module to store these instead of duplicating them, so that they can be reused, similar to what is done in atb-bff. 

Double-check that this doesn't go against the architecture.


adriansberg:
In regards to our discussion regarding architecture, do we need all the data in this fragments?

In terms of whether these fragments should be a module or not I'm not sure, @mikaelbr can answer that possibly.

mortennordseth:
> In regards to our discussion regarding architecture, do we need all the data in this fragments?

Thinking about it, this would increase the chance of us retrieving more data than needed. I'll close this PR and instead be more explicit in what data we fetch in each query.

PR #57, opened 2023-11-06
feat: show icon for situation or notice

adriansberg:
Show an icon when there's a situation or notice for the trip pattern. The icon is shown in the top right corner of a trip card.

Logic from the app is reused to maintain the same functionality across platforms.

Did an attempt to create a modue for situations and notices since there probably will be more coming for that, for the details page etc. Not sure if this is the way modules should be handled though.

Note: branched this off of the other PR, since the folder structure have been altered some. 

image

Fixes https://github.com/AtB-AS/kundevendt/issues/14919


mortennordseth (approved):
👍

PR #58, opened 2023-11-06
refactor: add module search time

adriansberg:
This module contains the types, utility functions and components related to search time. I have merged the types `DepartureDate` and `DepartureMode` to one, since in my eyes they were practically the same.


src/page-modules/assistant/__tests__/assistant.test.tsx
```diff
@@ -72,8 +72,11 @@ describe('assistant page', function () {
     (await expectProps(result)).toContain({
       initialFromFeature: fromFeature,
       initialToFeature: toFeature,
+      // initialSearchTime: {
+      //   mode: 'departBy',
+      //   dateTime: 123,
+      // },
```
mortennordseth, line 78:
```suggestion
```

mortennordseth (commented):
I think this looks much better 💯

src/modules/search-time/types.ts
```diff
@@ -0,0 +1,11 @@
+export const SEARCH_MODES = ['now', 'arriveBy', 'departBy'] as const;
+export type SearchMode = (typeof SEARCH_MODES)[number];
+
+export type SearchTime =
+  | {
+      mode: 'now';
+    }
+  | {
+      mode: 'arriveBy' | 'departBy';
+      dateTime: number;
```
mikaelbr, line 10: Should this be Date?

src/page-modules/assistant/__tests__/assistant.test.tsx
adriansberg: Yes. The reason I commented this out is because the test fails if it's there, but I can't see why...

src/modules/search-time/types.ts
adriansberg: It certainly can. Will look into why it was a number and possibly make it a Date.

adriansberg: This was the main reason IIRC. 
2023-11-08 at 20 29 20@2x

Need to handle the conversion to date somewhere, and if it can't be in the `getServerSideProps` I'm afraid it can create more noise than just using the millisecond representation.

I might be thinking wrong here and there's an easy solution to this problem.

mikaelbr: Aah, thats right. This is passed over server side props. https://github.com/blitz-js/next-superjson-plugin is a solution for this, but I think number is just fine for now 😅

adriansberg: TIL about that plugin 🤩

src/page-modules/assistant/types.ts
```diff
@@ -30,8 +30,12 @@ export const TripQuerySchema = z.object({
   toLat: z.number(),
   toLayer: z.union([z.literal('address'), z.literal('venue')]),
   filter: z.string().optional(),
-  departureDate: z.number().optional(),
-  departureMode: z.nativeEnum(DepartureMode),
+  searchMode: z.union([
+    z.literal('now'),
+    z.literal('arriveBy'),
+    z.literal('departBy'),
```
mortennordseth, line 36: Not sure if it's possible or that we should, but could it be wise to use the const SEARCH_MODES here? To avoid having to update the values two places

mortennordseth (approved):
One comment/question, but this looks like a big improvement to me! 👏

src/page-modules/assistant/types.ts
adriansberg: 2023-11-09 at 15 18 18@2x


Seems like we can't

mikaelbr: Can specify it originally as zod object and infer type from that and use that in reference

adriansberg: Good idea. So we'll end up having `SEARCH_MODES` (for mapping through in the component, maybe not necessary, could just use an array there), the zod type, and inferred type from zod type.

adriansberg: Pushed changes to that now, [18341a3](https://github.com/AtB-AS/planner-web/pull/58/commits/18341a39ed232f86e2fe802900e4e1849b5fb6f8)

PR #60, opened 2023-11-08
refactor: removes loads of unused data

mikaelbr:
Only pass on what is needed for showing non-transit suggestions.

Needs to be extended when the detail page is implemented (with added key/id)


adriansberg (approved):
Nice! I like this.

PR #62, opened 2023-11-10
feat: departure details page

mortennordseth:
This PR includes the new details page for a departure. Some of the components may be reused when we start working on the assistant departure page and could be moved to the components/modules folder then.

As described in the issue, this PR does not include situations/notices and the map. 

<details>
<summary> Screenshots </summary>

image

image

image

</details>

## TODO
- [x] Add tests

Closes https://github.com/AtB-AS/kundevendt/issues/9159


src/page-modules/departures/details/decoration-line/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  hasStart,
+  hasCenter,
+  hasEnd,
+}: DecorationLineProps) {
+  const colorStyle = { backgroundColor: color };
+  return (
+    <div className={style.decoration} style={colorStyle}>
+      {hasStart && (
+        <div
+          className={[
```
adriansberg, line 21: Can use the CSS until function `and` here.

src/page-modules/departures/details/estimated-call-rows.tsx
```diff
@@ -0,0 +1,199 @@
...
+        )}
+      </Row>
+      {collapseButton}
+    </div>
+  );
+}
+
+type RowProps = PropsWithChildren<{
+  rowLabel?: React.ReactNode;
+  alignChildren?: 'flex-start' | 'flex-end' | 'center';
```
adriansberg, line 160: We need this configurable as a prop?

src/styles/globals.css
```diff
@@ -97,6 +98,10 @@ textarea {
 .dark {
   --maxPageWidth: 1024px;
   --contentPadding: var(--spacings-xLarge);
+  --labelWidth: 5rem;
```
adriansberg, line 101: Is this the correct place for these variables?

mortennordseth: Are you thinking of globals.css in general or that specific class?

src/page-modules/departures/details/estimated-call-rows.tsx
mortennordseth: We need to know if the items should be aligned at the start (first quay in a group) or at the end (last quay in a group), or be centered. Alternatively, we could change it and add the props `isStartOfGroup` and `isEndOfGroup` if you think that is better

src/styles/globals.css
adriansberg: `globals.css`. For me, these variables look more like local variables rather than global variables accessible everywhere. Can't see that these variables are needed outside the details page? (Haven't hunted down where they are used, so it might be)

src/page-modules/departures/details/estimated-call-rows.tsx
adriansberg: I see now, that's fine. :)

src/styles/globals.css
mortennordseth: The reason why I chose .dark/.light is that we have other global variables defined here. Could move them all to `:root` or similar if that's better. 

These variables are needed by both the `Row` and `DecorationLine` components. Instead of defining them twice I thought it would be nice to have them in one place and I therefore went with the globals.css.

Can you think of a better place for them or another approach to this?

mortennordseth: ^ I didn't refresh and didn't see your answer before posting 😄

adriansberg: Personally I would have placed the variables in their common parent, which looks to be `.rowContainer`, if I'm looking at this correctly.

src/components/line-chip/line-chip.module.css
```diff
@@ -0,0 +1,14 @@
+.container {
+  min-width: 4.125rem;
```
adriansberg, line 2: 2023-11-10 at 14 13 37@2x


The line chip looks a bit off for me with single digit public codes.

Is this `min-width` needed?

src/modules/time-representation/index.ts
```diff
@@ -0,0 +1,31 @@
...
+  expectedTime,
+}: TimeValues): TimeRepresentationType {
+  if (missingRealTime) {
+    return 'no-realtime';
+  }
+  if (!expectedTime) {
+    return 'no-significant-difference';
+  }
+  const secondsDifference = Math.abs(secondsBetween(aimedTime, expectedTime));
+  return secondsDifference <= DEFAULT_THRESHOLD_AIMED_EXPECTED_IN_MINUTES * 60
```
adriansberg, line 28: Wouldn't it make sense to include the `* 60` in the constant?

src/page-modules/departures/details/details.module.css
```diff
@@ -0,0 +1,100 @@
...
+
+.callRows {
+  margin-top: var(--spacings-large);
+}
+.callRows__container {
+  padding: var(--spacings-medium);
+}
+.rowContainer {
+  position: relative;
+  --labelWidth: 5rem;
```
adriansberg, line 65: I see I've been using kebab case for variables, but probably should have used camel case as it is here. We should probably be consistent with it. The variables in `globals.css` are camel cased so that's probably the way to go.

src/page-modules/departures/details/decoration-line/decoration-line.module.css
```diff
@@ -0,0 +1,35 @@
+.decoration {
```
adriansberg, line 1: The decoration line should be right between the left column and the right column. As it is now the right column is closer.

image

Easily solved with a `transform: translateX(-50%)`, possibly some other easy solution as well.

Becomes like this:

image

src/page-modules/departures/details/details.module.css
```diff
@@ -0,0 +1,100 @@
...
+  gap: var(--spacings-small);
+  font-size: 0.875rem;
+  color: var(--text-colors-secondary);
+}
+
+.callRows {
+  margin-top: var(--spacings-large);
+}
+.callRows__container {
+  padding: var(--spacings-medium);
```
adriansberg, line 61: In the design the spacing between the first group(?) and the second is `1.25rem`, this padding adds up to `1.5rem`. Not too big of a deal I think.

image

src/components/line-chip/line-chip.module.css
mortennordseth: This was done to ensure the same width for all chips in scenarios like this: 

image

We might want to rethink this.

src/page-modules/departures/details/details.module.css
```diff
@@ -0,0 +1,100 @@
...
+  --labelWidth: 5rem;
+  --decorationContainerWidth: 1.25rem;
+  --decorationLineWidth: 0.25rem;
+  --decorationLineEndWidth: 0.75rem;
+}
+.boardingInfo {
+  color: var(--text-colors-secondary);
+}
+.collapseButton {
+  padding-bottom: var(--spacings-medium);
```
adriansberg, line 74: This padding makes the focus outline a bit weirder than it already is (IMO this button mode should have _some_ padding to make the outline be a tad bit better).

2023-11-10 at 14 28 35@2x

src/page-modules/departures/details/details.module.css
mortennordseth: I agree that we should be consistent! I can change this if you prefer kebab case. I think I'm a bit all over the place regarding this...

src/components/line-chip/line-chip.module.css
adriansberg: Right, of course. Let's keep it as is.

src/page-modules/departures/details/details.module.css
adriansberg: No strong preference either. @mikaelbr can decide.

src/page-modules/departures/details/details.module.css
mortennordseth: Would have to change the padding to `0.625rem` instead of `var(--spacings-medium)` then 🤔

src/page-modules/departures/details/details.module.css
mortennordseth: Fixed 🎉

src/page-modules/departures/details/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+          title={t(PageText.Departures.details.back)}
+          icon={{ left: <MonoIcon icon="navigation/ArrowLeft" /> }}
+        />
+        <div className={style.header}>
+          <LineChip
+            transportMode={serviceJourney.transportMode}
+            transportSubmode={serviceJourney.transportSubmode}
+            publicCode={serviceJourney.line.publicCode}
+          />
+          <Typo.h2 textType="heading--big">{title}</Typo.h2>
```
adriansberg, line 48: Should there be any accessibility label for the header of the page, saying something about the chosen stop place and potentially the real time information? Not sure how it should be done when theres no natural focusable element.

src/page-modules/departures/details/estimated-call-rows.tsx
```diff
@@ -0,0 +1,199 @@
...
+    </div>
+  );
+}
+
+type EstimatedCallRowProps = {
+  call: EstimatedCallWithMetadata;
+  mode: TransportModeType;
+  collapseButton: JSX.Element | null;
+};
+function EstimatedCallRow({
```
adriansberg, line 106: I think an estimated call should have a descriptive a11y label. As it is now it's like "Sanntid 15:01 15:05".

src/page-modules/departures/details/index.tsx
mortennordseth: I was thinking about this and I think we should. I'm not sure how to solve this either, and I see that we are missing this from the other pages as well 🤔 Should we make an issue to find a solution for this and update where needed?

src/page-modules/departures/details/details.module.css
adriansberg: Or could do some css fanciness. `padding-bottom: var(spacings-large);` on first-child, and `padding-top: 0;` on last-child. Not sure if that's a better solution than accepting a larger gap though. And maybe there are more potential callRows__containers inbetween?

src/page-modules/departures/details/index.tsx
adriansberg: Sure, that works.

src/page-modules/departures/details/estimated-call-rows.tsx
mortennordseth: It should be "Sanntid 15:01 15:05 Persaunet leir". Are you not getting the quay name? 🤔 This could be improved for sure. It's currently the same as in the app

adriansberg: Probably got that. But which one is sanntid for instance.

adriansberg: Point is, maybe it should be something more descriptive that describes which is the realtime and which is the expected. "Departure from {stopplace}. Expected {time}, real time {time}" or something.

src/page-modules/departures/details/index.tsx
mortennordseth: https://github.com/AtB-AS/kundevendt/issues/15431

src/page-modules/departures/details/estimated-call-rows.tsx
mortennordseth: My bad, it should. Fixed now! 
"Sanntid 13:47 Rutetid 13:45"

src/page-modules/departures/details/details.module.css
```diff
@@ -0,0 +1,104 @@
...
+  max-height: 37.5rem;
+  width: 100%;
+  margin: 0 auto;
+}
+.serviceJourneyContainer {
+  grid-area: serviceJourney;
+}
+@media (max-width: 650px) {
+  .serviceJourneyContainer {
+    margin: 0 auto;
```
adriansberg, line 39: Should this be added for smaller screens? Looks like the content gets too far to the right with this, instead of allowing it to use the full width.

mortennordseth: Nope, this is wrong. Fixed now 👍

src/page-modules/departures/details/details.module.css
mortennordseth: This has been solved now, thanks for the help on Slack @adriansberg 🎉

adriansberg (approved):
LGTM

PR #63, opened 2023-11-12
refactor: refactor transport mode to a module

adriansberg:
Refactoring of transport mode from components into a module. Added both transport icon and transport mode filter as directories in the module and exported them from the entry point.

I left the `TransportMode` type/enum in `@atb/modules/graphql-types` alone since it was used for casting values passed to the GraphQL query, was not sure how we should handle stuff like this.


mortennordseth (approved):
I think this looks much better 🎉

src/page-modules/departures/server/journey-planner/index.ts
```diff
@@ -34,17 +34,12 @@ import {
   ServiceJourneyData,
   serviceJourneySchema,
 } from './validators';
+import { TransportMode as GraphQlTransportMode } from '@atb/modules/graphql-types';
```
adriansberg, line 37: @mikaelbr, what do you think about the `TransportMode` in `modules/graphql-types`? Should that one be left alone and used to cast like it's done here?

mikaelbr: Haven't seen every detail here, but at least generally I think the type from graphql should be converted into something "safer" at first interaction with the type in some way. So inside API fn mapping typically

adriansberg: Looks like it's handled in some way. When mapping from GraphQL (https://github.com/AtB-AS/planner-web/blob/ede4246d07b409a33ab06e83afc0ad2b7a331309/src/page-modules/departures/server/journey-planner/index.ts#L119) we check the transport mode that is received and parse it with zod against the schema (https://github.com/AtB-AS/planner-web/blob/ede4246d07b409a33ab06e83afc0ad2b7a331309/src/modules/transport-mode/utils.ts#L25-L43). When passing to GraphQL it is currently just casted to the GraphQL type since `'bus' === 'bus'`.

PR #65, opened 2023-11-12
feat: add transport mode filter to estimated calls query

adriansberg:
Adds transport modes as a variable to the estimated calls/nextDepartures query such that the additional estimated calls fetched adhere to the same filters as the original query. Not sure how much this matters.

### DIFF
--- a/src/page-modules/departures/server/journey-planner/index.ts
+++ b/src/page-modules/departures/server/journey-planner/index.ts
@@ -253,7 +253,7 @@ export function createJourneyApi(
           numberOfDepartures: 6, // TODO: Have to add one to account for the last departure of the original set of departures?
           startTime: new Date(input.startTime),
           transportModes:
-            (input.transportModes as TransportMode[]) ?? null,
+            (input.transportModes as GraphQlTransportMode[]) ?? null,
         },
       });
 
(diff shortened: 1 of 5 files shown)


src/page-modules/departures/server/journey-planner/index.ts
```diff
@@ -249,8 +250,10 @@ export function createJourneyApi(
         query: QuayEstimatedCallsDocument,
         variables: {
           id: input.quayId,
-          numberOfDepartures: 5,
+          numberOfDepartures: 6, // TODO: Have to add one to account for the last departure of the original set of departures?
```
mortennordseth, line 253: Not sure if I understand 🤔 Are you thinking of the duplicates that may happen? I don't think that always happens, but I might be wrong about this. Thinking about it, we should probably update this to do queries until we get a minimum number of departures, similar to what has been done for the trip query.

adriansberg: I only saw 4 new estimated calls being added. I saw that a set was being made to keep uniques, and thought maybe since the original set's last estimated call's start time (that's a mouthful) is used to get the next departures. Soo, thought maybe it was an overlap there with the last and first.

mortennordseth: Looks like you're correct. I think it's fine to update this to 6 then, but we should consider using a minimum number of departures at some point 👍

mortennordseth (approved):

PR #66, opened 2023-11-12
fix: fix page title for AtB

adriansberg:
This should probably be updated to proper titles as well. In Figma the title is "AtB Reisesøk". Not sure what it shall be called.

### DIFF
--- a/src/translations/common/titles.ts
+++ b/src/translations/common/titles.ts
@@ -2,7 +2,7 @@ import { translation as _ } from '@atb/translations/commons';
 import { orgSpecificTranslations } from '../utils';
 
 const TitlesInternal = {
-  siteTitle: _('AtB (nb)', 'AtB (en)', 'AtB  (nn)'),
+  siteTitle: _('AtB', 'AtB', 'AtB'),
 };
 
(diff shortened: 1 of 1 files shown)


src/translations/common/titles.ts
```diff
@@ -2,7 +2,7 @@ import { translation as _ } from '@atb/translations/commons';
 import { orgSpecificTranslations } from '../utils';
 
 const TitlesInternal = {
-  siteTitle: _('AtB (nb)', 'AtB (en)', 'AtB  (nn)'),
+  siteTitle: _('AtB', 'AtB', 'AtB'),
```
mortennordseth, line 5: Might want to update all of these to "AtB Reisesøk", "FRAM Reisesøk", etc., so that we are in line with the sketches. That way it's easier for Julia to verify the design - I think

adriansberg: What will that be in english? "FRAM Travel Search" etc.?

mortennordseth: I guess so! We can verify with Julia when we go through the implementation

PR #70, opened 2023-11-13
feat: map and polyline of service journey in departure detail

mortennordseth:
This PR includes the map with a polyline of a service journey on the departure detail page. 

Closes https://github.com/AtB-AS/kundevendt/issues/9161

<details>

image

image

image

image

</details>


src/page-modules/departures/server/journey-planner/index.ts
```diff
@@ -358,3 +383,54 @@ type RecursivePartial<T> = {
...
+    fromStopPlace?.latitude || 0,
+    fromStopPlace?.longitude || 0,
+  ];
+
+  const mainStartIndex = findIndex(points, fromCoordinates);
+  const mainEndIndex = points.length - 1;
+
+  const beforeLegCoords = points.slice(0, mainStartIndex + 1);
+  const mainLegCoords = points.slice(mainStartIndex, mainEndIndex + 1);
+  const afterLegCoords = points.slice(mainEndIndex);
```
mortennordseth, line 405: FYI, the `afterLegCoords` is not really needed before we start working on the trip details page.

adriansberg:
Should the map have a border radius?

src/page-modules/departures/server/journey-planner/index.ts
```diff
@@ -302,15 +310,30 @@ export function createJourneyApi(
...
+      )?.quay?.stopPlace;
+
       const data: RecursivePartial<ServiceJourneyData> = {
         id: serviceJourney?.id,
-        transportMode: isTransportModeType(serviceJourney?.transportMode)
-          ? serviceJourney?.transportMode
-          : 'unknown',
-        transportSubmode: serviceJourney?.transportSubmode,
+        transportMode: transportMode,
+        transportSubmode: transportSubmode,
```
adriansberg, line 327:
```suggestion
        transportMode,
        transportSubmode,
```

src/page-modules/departures/server/journey-planner/index.ts
```diff
@@ -358,3 +383,54 @@ type RecursivePartial<T> = {
...
+  transportMode: TransportModeType,
+  transportSubmode: TransportSubmodeType | undefined,
+  fromStopPlace?: { id: string; latitude?: number; longitude?: number },
+) => {
+  if (!pointsOnLink || !pointsOnLink.points) return [];
+  const points = polyline.decode(pointsOnLink.points);
+  const fromCoordinates: [number, number] = [
+    fromStopPlace?.latitude || 0,
+    fromStopPlace?.longitude || 0,
+  ];
```
adriansberg, line 398: I prefer using nullish coalescing in cases like these (`??`). Not sure what's the best convention however, probably doesn't matter.

src/page-modules/departures/details/index.tsx
```diff
@@ -55,7 +56,16 @@ export function DeparturesDetails({
         )}
       </div>
 
-      <div className={style.mapContainer}></div>
+      <div className={style.mapContainer}>
+        <Map
+          position={[
```
adriansberg, line 61: Is it possible to pass this as an object with named properties instead of an array? I'm in the same boat as @mikaelbr, not knowing which comes first.

src/modules/transport-mode/icon/index.tsx
```diff
@@ -97,7 +97,7 @@ const TRANSPORT_SUB_MODES_BOAT: TransportSubmodeType[] = [
   'sightseeingService',
 ];
 
-function modeToColor(
+export function modeToColor(
```
adriansberg, line 100: Should this be renamed to e.g. `transportModeToColor` now that it is used outside this module, to be more descriptive?

src/components/map/use-fullscreen-map.ts
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  const closeFullscreen = useCallback(() => {
+    if (!mapWrapperRef.current || !mapRef.current) return;
+    mapWrapperRef.current.style.display = '';
+    mapRef.current.resize();
+    setIsFullscreen(false);
+  }, [mapWrapperRef, mapRef]);
+
+  const handlEscapeKey = useCallback(
```
adriansberg, line 23: `handlEscapeKey` -> `handleEscapeKey`

src/components/map/map.tsx
```diff
@@ -1,45 +1,36 @@
...
 import mapboxgl from 'mapbox-gl';
 import style from './map.module.css';
 import 'mapbox-gl/dist/mapbox-gl.css';
 import { mapboxData } from '@atb/modules/org-data';
-import { isFeaturePoint, isStopPlace } from './utils';
 import Button from '../button/button';
 import { MonoIcon } from '@atb/components/icon';
 import { ComponentText, useTranslation } from '@atb/translations';
 import { FocusScope } from '@react-aria/focus';
+import { MapLegType } from '@atb/page-modules/departures/server/journey-planner/validators';
```
adriansberg, line 10: Should probably not import this?

package.json
```diff
@@ -22,6 +22,7 @@
     "@atb-as/config-specs": "^3.6.1",
     "@atb-as/theme": "^7.1.1",
     "@leile/lobo-t": "^1.0.5",
+    "@mapbox/polyline": "mapbox/polyline",
```
adriansberg, line 25: Is `mapbox/polyline` the version of the package?

src/page-modules/departures/server/journey-planner/index.ts
mortennordseth: In this case, I guess `??` is more correct since `fromStopPlace.longitude` is either a number or undefined 🤔 I'll update.

src/components/map/map.tsx
adriansberg: I'd we are to import from page-modules (which we should try to avoid), we should import from the server path based on what Mikael have said.

src/components/map/map.module.css
```diff
@@ -51,6 +51,13 @@
   width: 100%;
 }
 
+.borderRadius__top canvas {
```
adriansberg, line 54: Probably doesn't matter in the slightest, but would it be better to have the radius on the container and `overflow: hidden;`? Probably not, same same.

src/components/map/map.tsx
mortennordseth: `@atb/page-modules/departures/server` that is?

adriansberg: Yes. But are we able to use an unnamed type with only the properties we use or have a fragment type?

src/components/map/map.module.css
mortennordseth: I think that's better 👍 Let me know if you can think of a better CSS class name...

adriansberg (approved):

PR #75, opened 2023-11-16
fix: quay header

mortennordseth:
This makes it so that the quay header behaves similarly to the app. 
Fixes https://github.com/AtB-AS/kundevendt/issues/15497

Before:
image

With fix: 
image


adriansberg (approved):
Very nice!

PR #76, opened 2023-11-17
fix: trip decoration line color

mortennordseth:
I forgot to add the transport sub-mode to the estimated call row, resulting in the wrong trip decoration line color. 

Fixes https://github.com/AtB-AS/kundevendt/issues/15499


adriansberg (approved):
Ooh, didn't notice that 😬

PR #77, opened 2023-11-17
fix: quay name

mortennordseth:
This adds the public code to the quay name if defined (P1, 1, etc.). 

image

Fixes https://github.com/AtB-AS/kundevendt/issues/15506

### DIFF
--- a/src/page-modules/departures/details/utils.ts
+++ b/src/page-modules/departures/details/utils.ts
@@ -57,3 +57,9 @@ export function getSituationsToShowForCall(
       !alreadyShownSituationNumbers.includes(s.situationNumber),
   );
 }
+
+export function formatQuayName(quayName?: string, publicCode?: string | null) {
+  if (!quayName) return;
+  if (!publicCode) return quayName;
+  return `${quayName} ${publicCode}`;
+}
--- a/src/page-modules/departures/details/estimated-call-rows.tsx
+++ b/src/page-modules/departures/details/estimated-call-rows.tsx
@@ -19,7 +19,7 @@ import {
   SituationMessageBox,
   SituationOrNoticeIcon,
 } from '@atb/modules/situations';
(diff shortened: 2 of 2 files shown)


src/page-modules/departures/details/utils.ts
```diff
@@ -57,3 +57,9 @@ export function getSituationsToShowForCall(
       !alreadyShownSituationNumbers.includes(s.situationNumber),
   );
 }
+
+export function formatQuayName(quayName?: string, publicCode?: string | null) {
```
adriansberg, line 61: Is this util function necessarily needed? Can't see that it brings a lot of benefit over just using the properties directly.

adriansberg (approved):
Great! One remark only.

src/page-modules/departures/details/utils.ts
mortennordseth: Good point 🤔 I guess having an extra space at the end of the name doesn't really matter

adriansberg: It shouldn't, as long as it doesn't affect any potential wrapping or anything, but it won't wrap mid word either as it is now.

mortennordseth: Actually, it can be `null` which would result in 
```ts
const name = 'Olav tryggvason';
const publicCode = null;
console.log(`${name} ${publicCode}`) // Olav tryggvason null
```

adriansberg: `call.quay.publicCode ?? ''`

mortennordseth: True!

mortennordseth: Maybe easier to read the utils function? 😄

adriansberg: No worries having it either.

PR #79, opened 2023-11-17
fix: dark mode

mortennordseth:
To be honest, I'm not sure why this is the case, but the input for this function seems to be a string resulting in `i == true` returning `false`. This is why the dark mode isn't working properly. This fixes that, but it might be a better approach to this. My best guess is that cookies are stored as strings, but then I don't see why this isn't failing in the webshop. 🤷

Fixes https://github.com/AtB-AS/kundevendt/issues/15498

### DIFF
--- a/src/modules/cookies/cookies-context.tsx
+++ b/src/modules/cookies/cookies-context.tsx
@@ -44,7 +44,7 @@ export function AppCookiesProvider({
   const darkmode = useCookie<boolean>(
     DARKMODE_COOKIE_NAME,
     initialCookies.darkmode ?? undefined,
-    (i) => i == true,
+    (i) => (typeof i === 'string' ? i === 'true' : i == true),
   );
 
   return (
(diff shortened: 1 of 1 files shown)


src/modules/cookies/cookies-context.tsx
```diff
@@ -44,7 +44,7 @@ export function AppCookiesProvider({
   const darkmode = useCookie<boolean>(
     DARKMODE_COOKIE_NAME,
     initialCookies.darkmode ?? undefined,
-    (i) => i == true,
+    (i) => (typeof i === 'string' ? i === 'true' : i == true),
```
adriansberg, line 47: Could we do a triple equals in the last part now after checking types?

adriansberg: And will `i` always be a string?

adriansberg (approved):

src/modules/cookies/cookies-context.tsx
mortennordseth: Not entirely sure about all of this tbh, but it can be `string | boolean | null | undefined`. I've updated to use triple equals.

PR #82, opened 2023-11-21
feat: show error when geolocation fails/is blocked

adriansberg:
This PR introduces a new feature that improves the user experience by showing an informative error message whenever geolocation services fail or are blocked. It enhances the app's transparency about its operations and guides users to resolve location-based issues.

With this update, we can avoid confusion, assure the user about the application's functionality and increase overall reliability. 

<p>

image

</p>
</details> 

Fixes https://github.com/AtB-AS/kundevendt/issues/15347

### DIFF
--- a/src/components/loading-empty-results/index.tsx
+++ b/src/components/loading-empty-results/index.tsx
@@ -8,11 +8,13 @@ import style from './loading-empty-results.module.css';
 
 export type LoadingEmptyResultsProps = PropsWithChildren<{
   isSearching: boolean;
+  isGeolocationError?: boolean;
   type: keyof typeof ComponentText.EmptySearch.searching;
 }>;
 
 export default function LoadingEmptyResults({
   isSearching,
+  isGeolocationError = false,
   type,
   children,
 }: LoadingEmptyResultsProps) {
@@ -27,7 +29,7 @@ export default function LoadingEmptyResults({
 
   return (
     <>
-      {!isSearching && hasEmptyChild && (
+      {!isSearching && !isGeolocationError && hasEmptyChild && (
         <EmptyMessage
           title={t(ComponentText.EmptySearch.notSearched)}
           details={t(ComponentText.EmptySearch.emptyDetails[type])}
@@ -45,6 +47,12 @@ export default function LoadingEmptyResults({
             {t(ComponentText.EmptySearch.searching[type])}
           </Typo.p>
         </motion.div>
+      )}
+      {!isSearching && isGeolocationError && hasEmptyChild && (
+        <EmptyMessage
+          title={t(ComponentText.EmptySearch.nearbyNoGeolocation.title)}
+          details={t(ComponentText.EmptySearch.nearbyNoGeolocation.details)}
+        />
       )}
       {!isSearching && !hasEmptyChild && children}
     </>
--- a/src/components/search/geolocation-button/index.tsx
+++ b/src/components/search/geolocation-button/index.tsx
@@ -2,16 +2,31 @@ import { Component, useEffect, useState } from 'react';
 import { LoadingIcon } from '@atb/components/loading';
 import { reverse } from '@atb/page-modules/departures/client';
 import { GeocoderFeature } from '@atb/page-modules/departures';
-import { ComponentText, useTranslation } from '@atb/translations';
+import {
+  ComponentText,
+  TranslateFunction,
+  useTranslation,
+} from '@atb/translations';
 import { MonoIcon } from '@atb/components/icon';
 
 type GeolocationButtonProps = {
   onGeolocate: (feature: GeocoderFeature) => void;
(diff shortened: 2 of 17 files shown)


src/page-modules/departures/layout.tsx
```diff
@@ -207,9 +217,20 @@ function DeparturesLayout({
                 : t(PageText.Departures.search.searching.nearby)}
             </Typo.p>
           </motion.div>
-        ) : (
-          children
         )}
+
+        {!isSearching && geolocationError !== null && hasEmptyChild && (
+          <EmptyPageContent
```
adriansberg, line 223: Should we show something similar for the assistant page? Different text of course, since what it says about departures don't match.

mikaelbr:
hoho. sorry. I renamed somethings 😬

adriansberg:
> hoho. sorry. I renamed somethings 😬

image

src/page-modules/departures/layout.tsx
```diff
@@ -128,6 +130,13 @@ function DeparturesLayout({
               options={['now', 'departBy']}
             />
           </div>
+
+          {geolocationError !== null && (
+            <div className={style.geolocationError}>
+              <MonoIcon icon="status/Info" />
+              <Typo.p textType="body__primary">{geolocationError}</Typo.p>
+            </div>
```
mikaelbr, line 138: This might be a message component at some point?

src/components/loading-empty-results/index.tsx
```diff
@@ -8,11 +8,13 @@ import style from './loading-empty-results.module.css';
 
 export type LoadingEmptyResultsProps = PropsWithChildren<{
   isSearching: boolean;
+  isGeolocationError?: boolean;
   type: keyof typeof ComponentText.EmptySearch.searching;
 }>;
 
 export default function LoadingEmptyResults({
   isSearching,
+  isGeolocationError = false,
```
mikaelbr, line 17: Wonder if these could be union types instead 🤔 Instead of having several semi-contradictory booleans. But that would maybe move the logic responsibility to the consumer 🤔

src/components/search/geolocation-button/index.tsx
```diff
@@ -2,16 +2,31 @@ import { Component, useEffect, useState } from 'react';
...
+  className,
+}: GeolocationButtonProps) {
   const { t } = useTranslation();
-  const { getPosition, isLoading, isUnavailable } = useGeolocation(onGeolocate);
+  const { getPosition, isLoading, isUnavailable, error } =
+    useGeolocation(onGeolocate);
+
+  useEffect(() => {
+    if (error) onError?.(getErrorMessage(error.code, t));
+    else onError?.(null);
```
mikaelbr, line 28: wonder if onError rather should be passed to useGeolocation instead of having to use effect here. So just use it as a normal callback

src/page-modules/departures/layout.tsx
mikaelbr: Yeah, think this should be Message component of type `info`. Here I think webshop can be reused/copied directly?

mortennordseth: We already have this component (https://github.com/AtB-AS/planner-web/blob/main/src/modules/situations/situation-message-box.tsx#L101) actually. Should not be a problem to move the `MessageBox` component to the components folder and use it for other scenarios than situations or notices

adriansberg: Extracted it to components and used it instead. Nice, good idea.


Didn't bother adding the column-spanning to get the message box to span the entire width, but maybe I should do that? Requires either adding a wrapper element with the correct styling, or make it possible to pass a className to the message box component. IMO it looks okay being only beneath the search inputs.

2023-11-28 at 19 36 09@2x

src/components/loading-empty-results/index.tsx
adriansberg: Yeah, they are sort of mutually exclusive. At least in the output of the component.

Not sure which solution is better though. Making it a union type moves the responsibility over to the consumer as you say, and might be more complex when ternaries are used to set the property based on isSearching or isGeolocationError.

src/components/search/geolocation-button/index.tsx
adriansberg: Good idea, implemented that.

src/components/loading-empty-results/index.tsx
mikaelbr: lets not solve it now in any case

src/page-modules/departures/layout.tsx
mikaelbr: Could wrap it inside a div that has col span yeah. Think that is the better approach. 🤔

adriansberg: Will do

mikaelbr (approved):
:shipit:

PR #83, opened 2023-11-21
fix: adds empty search state

mikaelbr:
Fixes empty state / default state for assistant and departure views.

Screenshot 2023-11-22 at 09 28 45

### DIFF
--- a/src/components/empty-search/index.tsx
+++ b/src/components/empty-search/index.tsx
@@ -1,0 +1,66 @@
+import { motion } from 'framer-motion';
+import { PropsWithChildren } from 'react';
+import { Image } from '@atb/components/icon';
+
+import style from './empty-search.module.css';
+import { Typo } from '../typography';
+import { ComponentText, useTranslation } from '@atb/translations';
+import React from 'react';
+
+export type EmptyLoadingSearchProps = PropsWithChildren<{
+  isSearching: boolean;
+  type: keyof typeof ComponentText.EmptySearch.searching;
+}>;
+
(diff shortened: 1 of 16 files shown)


src/components/empty-search/index.tsx
```diff
@@ -0,0 +1,66 @@
+import { motion } from 'framer-motion';
+import { PropsWithChildren } from 'react';
+import { Image } from '@atb/components/icon';
+
```
adriansberg, line 4: Any logic behind separation of imports?

src/components/empty-search/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const hasEmptyChild = React.Children.toArray(children).some((child) => {
+    if (React.isValidElement(child)) {
+      return 'empty' in child.props;
+    }
+    return false;
+  });
+
+  return (
+    <>
+      {!isSearching && hasEmptyChild && (
```
adriansberg, line 31: Would it be clean to bake this into a switch?

src/components/empty-search/index.tsx
mikaelbr: Copy paste issue. I like to separate css from the others.

adriansberg: I'm on board with that.

src/components/empty-search/index.tsx
mikaelbr: I think when extending to error / geolocation we should bake this into type system

PR #84, opened 2023-11-23
fix: general performance for travel planner

mikaelbr:
Improves performance by:

- Making more calls in parallell
- Removing unused data from trip planner query
- Increasing search window for trips.
- Fixing bug in do-while loop condition.

### DIFF
--- a/src/page-modules/assistant/server/journey-planner/index.ts
+++ b/src/page-modules/assistant/server/journey-planner/index.ts
@@ -223,23 +223,14 @@ function mapResultToTrips(
           distance: leg.distance,
           duration: leg.duration,
           aimedStartTime: leg.aimedStartTime,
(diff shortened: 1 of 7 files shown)


adriansberg:
Did you do any measurements?

mikaelbr:
No, but this in many cases removes 3 gql trip calls and parallels 2 calls, making it essentially going from 5 calls in sequence to 2 calls. So 2/5 of the original time.

mortennordseth (approved):
Nice! 👍 Some of the data that you've removed is needed for the trip details that I am currently working on. I can add what is needed or it might actually be better to create a new query that is used when getting a specific trip for trip details. Either way, I'll handle that in the PR for trip details

mikaelbr:
I think it is better to have separate fetch points for details and list, as they have different requirements and require different data

PR #85, opened 2023-11-27
feat: single trip query

mortennordseth:
This PR introduces a new query for fetching trips with more details, functionality to create a compressed query string that can be used to look up a specific trip, and the base layout for the assistant details page. 

Partially closes https://github.com/AtB-AS/kundevendt/issues/15639

PR #86, opened 2023-11-27
feat: trip details header

mortennordseth:
This PR adds the trip details header information. 

image

image

Based on https://github.com/AtB-AS/planner-web/pull/85 (awaiting merge)

Partially closes https://github.com/AtB-AS/kundevendt/issues/15639

### DIFF
--- a/src/page-modules/assistant/details/index.tsx
+++ b/src/page-modules/assistant/details/index.tsx
@@ -18,7 +18,7 @@ export type AssistantDetailsProps = {
 
 export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
   const { t } = useTranslation();
-  const { fromName, toName, weekdayAndDate, timeRange, tripDuratrion } =
+  const { fromName, toName, weekdayAndDate, timeRange, tripDuration } =
     useTripDetailsHeader(tripPattern);
   return (
     <div className={style.container}>
@@ -52,7 +52,7 @@ export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
             <MonoIcon icon="time/Duration" />
             <Typo.p textType="body__primary">{timeRange}</Typo.p>
             <Typo.p textType="body__primary--bold">
-              {t(PageText.Assistant.details.header.travelTime(tripDuratrion))}
+              {t(PageText.Assistant.details.header.travelTime(tripDuration))}
             </Typo.p>
           </div>
         </div>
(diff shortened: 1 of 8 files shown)


mikaelbr:
Hard to separate PRs, think merge and rebase of #85 makes it easier

src/page-modules/assistant/details/index.tsx
```diff
@@ -12,8 +18,8 @@ export type AssistantDetailsProps = {
 
 export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
   const { t } = useTranslation();
-  const fromName = tripPattern.legs[0].fromPlace.name;
-  const toName = tripPattern.legs[tripPattern.legs.length - 1].toPlace.name;
+  const { fromName, toName, weekdayAndDate, timeRange, tripDuration } =
+    useTripDetailsHeader(tripPattern);
```
mikaelbr, line 22: Can tripPattern be null?

mortennordseth:
My bad, @mikaelbr. I thought I handled it by rebasing from the other PR. It should be easier to read now 👍

src/page-modules/assistant/server/journey-planner/journey-gql/trip-with-details.gql
```diff
@@ -28,6 +28,7 @@ query TripsWithDetails(
     tripPatterns {
       expectedStartTime
       expectedEndTime
+      duration
```
mikaelbr, line 31: This is fantastic. Introducing new fields when they are needed. love it

src/page-modules/assistant/details/index.tsx
```diff
@@ -37,7 +43,46 @@ export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
...
+            <Typo.p textType="body__primary">{weekdayAndDate}</Typo.p>
+          </div>
+          <div className={style.duration}>
+            <MonoIcon icon="time/Duration" />
+            <Typo.p textType="body__primary">{timeRange}</Typo.p>
+            <Typo.p textType="body__primary--bold">
+              {t(PageText.Assistant.details.header.travelTime(tripDuration))}
+            </Typo.p>
+          </div>
+        </div>
```
mikaelbr, line 58: Could this instead of being custom hook, be a separate component?

src/page-modules/assistant/details/index.tsx
mortennordseth: The server can return null, but if so we will return the `EmptyMessage` component instead of `AssistantDetails`. 

https://github.com/AtB-AS/planner-web/blob/main/src/pages/assistant/%5Bid%5D.tsx#L19

src/page-modules/assistant/details/index.tsx
mikaelbr: Now were essentially decoupling data transformation and UI. But they are semantically tightly coupled. Maybe introducing this as a component and instead of having `useTripDetailsHeader` let this be transformed inside the component, it is coupled more correctly?

mortennordseth: I think that makes sense 👍 I've updated to reflect this now

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -229,6 +229,7 @@ export function createJourneyApi(
       const data: RecursivePartial<TripPatternWithDetails> = {
         expectedStartTime: singleTripPattern?.expectedStartTime,
         expectedEndTime: singleTripPattern?.expectedEndTime,
+        duration: singleTripPattern?.duration || 0,
```
adriansberg, line 232:
```suggestion
        duration: singleTripPattern?.duration ?? 0,
```

adriansberg (approved):

PR #87, opened 2023-11-27
fix: Change widget to init by passing URL to client

mikaelbr:
New example of usage:

```
  // Ensure that url base is same origin as the page where widget is loaded and the travel planner API
  const widget = window.PlannerWeb.createWidget({
    urlBase: 'https://reiseplanlegger.example.no/',
  });

  // After loading JS and CSS file it can be initialized using the following code:
  widget.init();

```

This simplifies build and allow us to bundle same asset files to staging and production. Integration strategy remains the same.

Part of https://github.com/AtB-AS/kundevendt/issues/14939

### DIFF
--- a/src/widget/widget.ts
+++ b/src/widget/widget.ts
@@ -6,278 +6,53 @@ import Combobox from '@github/combobox-nav';
 
 import style from './widget.module.css';
 
-const URL_BASE = import.meta.env.VITE_WIDGET_BASE_URL;
-
-export { URL_BASE };
-export const URL_JS_UMD = `${URL_BASE}widget/planner-web.umd.js`;
-export const URL_JS_ESM = `${URL_BASE}widget/planner-web.mjs`;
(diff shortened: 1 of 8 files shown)


src/widget/widget.ts
```diff
@@ -316,6 +91,402 @@ export function init() {
...
+    const li = el('li', [img, title, locality], style.listItem);
+    li.role = 'option';
+    li.setAttribute('data-feature-id', item.id);
+    return li;
+  }
+
+  function venueIcon(item: GeocoderFeature) {
+    const icon = iconData(item.category);
+
+    // http://localhost:3000/assets/mono/light/transportation-entur/Bus.svg
```
adriansberg, line 108: Is this comment here as an example?

mikaelbr: This is not the line you're looking for...

adriansberg (approved):
Looking good. Nice to remove the need of defining an URL during build time.

PR #89, opened 2023-11-27
feat: Add support for search date for widget

mikaelbr:
Adds support for searching on specific arrival time and departure time.

Part of https://github.com/AtB-AS/kundevendt/issues/14939

### DIFF
--- a/src/widget/widget.ts
+++ b/src/widget/widget.ts
@@ -82,7 +82,7 @@ function init() {
     ?.addEventListener('submit', (e) => {
       e.preventDefault();
       const form = e.currentTarget as HTMLFormElement;
-      submitDeparture(form.action, fromTo.from!);
+      submitDeparture(form, fromTo.from!);
     });
 
   document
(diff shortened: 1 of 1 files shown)


src/widget/widget.ts
```diff
@@ -286,16 +352,21 @@ function createOutput({ URL_BASE }: SettingConstants) {
         >
           <div class="${style.selector_dateAndTimeSelectors}">
             <div class="${style.selector_dateSelector}">
-              <label for="searchTimeSelector-date">Dato</label>
+              <label for="${`${prefix}-searchTimeSelector-date`}">Dato</label>
               <input
                 type="date"
-                id="searchTimeSelector-date"
-                value="2023-11-10"
+                name="dateinput"
```
adriansberg, line 358: Would this name ever conflict with other names on the page it is used?

mikaelbr: names are scoped to forms

PR #91, opened 2023-11-27
feat: assistant details map

mortennordseth:
✅  https://github.com/AtB-AS/planner-web/pull/90 should be merged before this one.

This PR adds the map with a polyline to the assistant details page. I've also moved the map functionality to the module folder to avoid duplicates and make it easier to reuse. 

Partially closes https://github.com/AtB-AS/kundevendt/issues/15639

<details>

image

image

</details>


mikaelbr:
I know I'm being difficult now, but I actually think map might be a component here. Only thing that makes it up for discussion is its dependence on `orgData`. But we could for instance make that injectable instead of using it directly. Or we could lift orgData to not be a module because it is technically not really a module but hardcoded static data, which makes it kind of an odd one out. But the map here isn't dependent on any IO or data. It is simply a component taking props and returning something. It doesn't require anything other than itself (and some types)

mortennordseth:
> But the map here isn't dependent on any IO or data. It is simply a component taking props and returning something.

Very true. I forgot to think of it like this. It felt wrong to import the `MapLegType` and its utility functions from the component folder, but that could maybe be global types and functions? 🤔

mikaelbr:
Components can refer to other components as the are "on the same level".

mortennordseth:
In this case, I need the type (or schema to be precise) for the validators server side for both the departure and assistant page modules. I might be thinking of it wrong here and that there is a better approach to this though 🤔

src/components/map/map.tsx
```diff
@@ -3,18 +3,18 @@ import mapboxgl from 'mapbox-gl';
 import style from './map.module.css';
 import 'mapbox-gl/dist/mapbox-gl.css';
 import { mapboxData } from '@atb/modules/org-data';
-import Button from '../button/button';
+import Button from '@atb/components/button/button';
```
mikaelbr, line 6: Should this be?
```suggestion
import { Button } from '@atb/components/button';
```

mortennordseth: Yep!

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -230,6 +231,7 @@ export function createJourneyApi(
         expectedStartTime: singleTripPattern?.expectedStartTime,
         expectedEndTime: singleTripPattern?.expectedEndTime,
         duration: singleTripPattern?.duration ?? 0,
+        walkDistance: singleTripPattern?.streetDistance ?? 0,
```
adriansberg, line 234: Should we rename this to streetDistance internally as well?

adriansberg: walkDistance makes sense though.

mortennordseth: We could do that, but as we are displaying it as walk distance in the UI it might make sense to name it walk distance? 🤔

adriansberg: That was a thought I had after I made this comment too. Can keep it as is.

mikaelbr: I think that is a good point also. our type is shown as walk distance, so mapping it makes sense to me

mikaelbr (approved):

PR #92, opened 2023-11-27
feat: rail replacement bus information

mortennordseth:
This PR adds rail replacement information and fixes a text color bug in dark mode for the `MessageBox` component

image

image


src/page-modules/assistant/details/trip-section/index.tsx
```diff
@@ -81,6 +85,19 @@ export default function TripSection({
           </TripRow>
         )}
 
+        {leg.transportSubmode === 'railReplacementBus' && (
```
mikaelbr, line 88: Is this something that could be modelled in the new flexible notification system @toretefre @rosvik ?

rosvik: Global messages are best for messages that are org specific and/or likely to change at some point, which I don't think this is. Have also tried to keep them "one collection of messages per page", not like this which would have one <GlobalMessages> per trip leg.

So I think hard-coding this is better.

mikaelbr: Good input! Thanks

mikaelbr (approved):

PR #93, opened 2023-11-27
feat: show interchanges in trip details

mortennordseth:
This PR adds information about interchanges in trip details

image


src/page-modules/assistant/details/trip-section/index.tsx
```diff
@@ -119,30 +131,33 @@ export default function TripSection({
...
+                      PageText.Assistant.details.tripSection.interchangeWithUnknownFromPublicCode(
+                        interchangeDetails.publicCode,
+                        interchangeDetails.fromPlace,
+                      ),
+                    )
+              }
+            />
+          </TripRow>
+        </div>
+      )}
```
mikaelbr, line 160: Maybe this should be its own component to encapsulate some details and to colocate typings?

mikaelbr: Also the utils for this specific task

mikaelbr (changes requested):

src/page-modules/assistant/details/trip-section/index.tsx
mortennordseth: I agree 👍 I've updated to reflect this now

PR #94, opened 2023-11-27
feat: trip details wait section

mortennordseth:
This PR adds information about wait times on the trip details page. 

image


src/page-modules/assistant/details/trip-section/index.tsx
```diff
@@ -102,30 +121,36 @@ export default function TripSection({
...
+                        interchangeDetails.publicCode,
+                        interchangeDetails.fromPlace,
+                      ),
+                    )
+              }
+            />
+          </TripRow>
+        </div>
+      )}
+      {legWaitDetails?.mustWaitForNextLeg && legWaitDetails.waitTime > 30 && (
```
mikaelbr, line 151: These types of checks can also be named. 
```suggestion
      {showWaitSection && (
```

It makes it easier to read components i think

mikaelbr: Could also potentially pass in `mustWaitForNextLeg` and have the conditional inside `WaitSection`. Reducing the overhead in reading the component. Drawback is that it isn't as that it is optional, but I think the Component Prop signature with `mustWaitForNextLeg | undefined` can be good enough a indicator

mikaelbr (changes requested):
See comment on simplifying consume of wait time

src/page-modules/assistant/details/trip-section/index.tsx
mortennordseth: I like this 👍 I actually think that it is easier to read the component using the named check than passing the `mustWaitForNextLeg` to the component 🤔 I'm up for both changes, but I think the way it is now is good

```tsx
{showInterchangeSection && (
  <InterchangeSection
    interchangeDetails={interchangeDetails}
    publicCode={leg.line?.publicCode}
  />
)}
{showWaitSection && (
  <WaitSection waitTimeInSeconds={legWaitDetails.waitTime} />
)}
```

mikaelbr: I might have misread as to that `mustWaitForNextLeg` was part of the same `legWaitDetails` object. But yeah, I also think this is fine enough!

mortennordseth: It actually is, and when you say it like that I guess it makes more sense 😄 I didn't include the `mustWaitForNextLeg` to the `WaitSection` props as it was not needed, but I guess passing the "entire" object and check if it should return something makes sense

mortennordseth: Then again, we should probably do the same with the `InterchangeSection` and pass `guaranteed` as a prop 🤔

src/page-modules/assistant/details/index.tsx
```diff
@@ -70,12 +71,13 @@ export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
...
             key={index}
             isFirst={index === 0}
-            isLast={index === tripPattern.legs.length}
+            isLast={index === tripPattern.legs.length - 1}
             leg={leg}
             interchangeDetails={getInterchangeDetails(
               tripPattern.legs,
               leg.interchangeTo?.toServiceJourney.id,
             )}
+            legWaitDetails={getLegWaitDetails(leg, tripPattern.legs[index + 1])}
```
mikaelbr, line 80: This kind of still feels odd for me. Both interchange and this. I don't have any better options for now, but feels kind of fragmented domain logic wise 🤔

mikaelbr (approved):

src/page-modules/assistant/details/trip-section/index.tsx
mikaelbr: That might make sense

src/page-modules/assistant/details/index.tsx
mortennordseth: We might want to handle this server side?

src/page-modules/assistant/details/utils.ts
```diff
@@ -23,3 +24,24 @@ export function formatLineName(
...
+    leg.expectedEndTime,
+    nextLeg.expectedStartTime,
+  );
+  const mustWaitForNextLeg = waitTime > 0;
+
+  return {
+    waitTime,
+    mustWaitForNextLeg,
+  };
+}
```
mikaelbr, line 47: I think this maybe should be moved to wait section and be made a part of that API as it is so closely coupled. Does it make sense to use it outside that component? if not, it should be colocated

mortennordseth: Very true!

src/page-modules/assistant/details/index.tsx
mikaelbr: Yes! Good point. If this is just data transformation I think it should happen in the service layer.

PR #96, opened 2023-11-28
feat: end and start text labels for a trip on the map

mortennordseth:
This PR adds text labels to the start and end points of a trip on the map. I've made the labels look the same as in the app and not the Figma design. Due to technical constraints, it was not feasible to implement the text background as depicted in the Figma designs.

Fixes https://github.com/AtB-AS/kundevendt/issues/9161#issuecomment-1822271071 reported by @HanneLH . 

image


HanneLH (approved):
🌸 
It looks good for me!

mikaelbr (approved):

PR #98, opened 2023-11-29
feat: situations and notices on travel details

mortennordseth:
This PR adds situations and notices to the travel details page. 

image

### DIFF
--- a/src/page-modules/assistant/server/journey-planner/journey-gql/trip-with-details.gql
+++ b/src/page-modules/assistant/server/journey-planner/journey-gql/trip-with-details.gql
@@ -46,6 +46,9 @@ query TripsWithDetails(
         line {
           name
           publicCode
+          notices {
+            ...notice
+          }
         }
         fromPlace {
           name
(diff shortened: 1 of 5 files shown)


src/page-modules/assistant/server/journey-planner/journey-gql/trip-with-details.gql
```diff
@@ -82,6 +96,9 @@ query TripsWithDetails(
         intermediateEstimatedCalls {
           aimedDepartureTime
         }
+        situations {
+          ...situation
```
adriansberg, line 100: Is everything in `situation` needed?

mortennordseth: We can remove ID, advice, and infoLinks. Should update that for all situations in the app then

mortennordseth: Instead of removing those, I added them to the situation dialog instead to have the same information as in the app.

adriansberg: Nice, that's fine 👏

adriansberg (approved):

PR #100, opened 2023-11-29
feat: walking distance unit and format decimals

mortennordseth:
This PR adds what unit the walking distance is in and removes the decimals from the distance. 

Closes https://github.com/AtB-AS/kundevendt/issues/15792

### DIFF
--- a/src/translations/pages/assistant.ts
+++ b/src/translations/pages/assistant.ts
@@ -337,9 +337,9 @@ export const Assistant = {
         ),
       walkDistance: (distance: string) =>
         _(
-          `Total gangavstand: ${distance}`,
-          `Total walking distance: ${distance}`,
-          `Total gangavstand: ${distance}`,
+          `Total gangavstand: ${distance} m`,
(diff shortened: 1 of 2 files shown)


src/translations/pages/assistant.ts
```diff
@@ -337,9 +337,9 @@ export const Assistant = {
         ),
       walkDistance: (distance: string) =>
         _(
-          `Total gangavstand: ${distance}`,
-          `Total walking distance: ${distance}`,
-          `Total gangavstand: ${distance}`,
+          `Total gangavstand: ${distance} m`,
+          `Total walking distance: ${distance} m`,
+          `Total gangavstand: ${distance} m`,
```
adriansberg, line 342: We always keep the distance in meters I presume? 

Does the app show the distance in the abbreviated form like this or in full form?

mortennordseth: Always meters and like this ☝️

adriansberg (approved):
:shipit:

PR #101, opened 2023-11-29
feat: support from / to individually from URL

mikaelbr:
This adds ability to set from and to independently from the URL. It also refactors and simplifies using from/to in the assistant component as to not using state.

Every time we use URL we should use that as a state also. This simplifies things usually, and will make sure the two states doesn't get out of sync.

More URL navigation also shows the slowness of trip API more. Should definitely move trip search to be client side only.

There is still som type mess here and there that I'm not quite happy about. But didn't take the time to fix it here. E.g. Assistant referencing Departure etc.

Things to test out:

- [ ] Works as expected with search, swap, geolocation.
- [ ] Works as expected with submit
- [ ] Filters
- [ ] Search time
- [ ] Load more
- [ ] Load more when search time
- [ ] Filters and load more

### DIFF
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -1,6 +1,7 @@
 import { Button } from '@atb/components/button';
-import EmptySearch from '@atb/components/loading-empty-results';
 import { MonoIcon } from '@atb/components/icon';
+import EmptySearch from '@atb/components/loading-empty-results';
+import { MessageBox } from '@atb/components/message-box';
 import Search, { GeolocationButton, SwapButton } from '@atb/components/search';
 import { Typo } from '@atb/components/typography';
 import type { SearchTime } from '@atb/modules/search-time';
@@ -8,7 +9,6 @@ import SearchTimeSelector from '@atb/modules/search-time/selector';
 import {
   TransportModeFilter,
   getInitialTransportModeFilter,
-  type TransportModeFilterOption,
 } from '@atb/modules/transport-mode';
 import type { GeocoderFeature } from '@atb/page-modules/departures';
 import { PageText, useTranslation } from '@atb/translations';
@@ -17,9 +17,8 @@ import { AnimatePresence, motion } from 'framer-motion';
 import { useRouter } from 'next/router';
 import { FormEventHandler, PropsWithChildren, useState } from 'react';
 import style from './assistant.module.css';
-import { createTripQuery } from './utils';
-import { MessageBox } from '@atb/components/message-box';
 import { FromToTripQuery } from './trip-query-fetcher';
+import { createTripQuery } from './utils';
 
 export type AssistantLayoutProps = PropsWithChildren<{
   tripQuery: FromToTripQuery;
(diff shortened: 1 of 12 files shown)


mortennordseth (approved):
This makes a lot of sense 🙌 I've done some testing and it works great

src/page-modules/assistant/layout.tsx
```diff
@@ -17,76 +17,61 @@ import { AnimatePresence, motion } from 'framer-motion';
...
-    setIsSearching(true);
-    const query = createTripQuery(
-      selectedFromFeature,
-      selectedToFeature,
-      searchTime,
-      transportModeFilter,
-    );
-    await router.push({ pathname: '/assistant', query });
-    setIsSearching(false);
+    setValuesWithLoading({});
```
adriansberg, line 67: Why empty object here?

src/page-modules/assistant/layout.tsx
```diff
@@ -103,22 +88,20 @@ function AssistantLayout({
 
             <Search
               label={t(PageText.Assistant.search.input.from)}
-              onChange={setSelectedFromFeature}
-              initialFeature={initialFromFeature}
-              selectedItem={selectedFromFeature}
+              onChange={onFromSelected}
+              selectedItem={tripQuery.from ?? undefined}
```
adriansberg, line 92: Isn't this the same as just passing `tripQuery.from` since it can be undefined?

src/page-modules/assistant/utils.ts
```diff
@@ -72,13 +78,13 @@ export function createTripQuery(
...
 
 export function parseTripQuery(query: any): TripQuery | undefined {
-  const requiredNumericFields = ['fromLat', 'fromLon', 'toLat', 'toLon'];
-  requiredNumericFields.forEach((field) => {
-    if (typeof query[field] === 'string')
-      query[field] = parseFloat(query[field]);
-  });
-
-  const optionalNumericFields = ['searchTime'];
+  const optionalNumericFields = [
```
adriansberg, line 81: Is it possible to connect this with the type somewhat? To simplify when we eg refactor so that this gives an error because names don't match.

src/pages/api/assistant/trip.ts
```diff
@@ -45,12 +23,13 @@ export default handlerWithAssistantClient<TripApiReturnType>({
     return tryResult(req, res, async () => {
       return ok(
         await client.trip({
-          from,
-          to,
-          searchTime,
-          transportModes:
-            getAllTransportModesFromFilterOptions(transportModeFilter),
-          cursor: tripQuery.cursor,
+          from: tripQuery.from!,
```
adriansberg, line 26: Any alternatives to using `!` here?

src/page-modules/assistant/layout.tsx
mikaelbr: Gah, this is something I'm not very happy with. But difference between `null` and `undefined`. Need to support null for serializing between server and client

src/page-modules/assistant/layout.tsx
mikaelbr: Could be undefined, but feel just as OK to have empty override i feel to avoid the undefined stuff.

src/pages/api/assistant/trip.ts
mikaelbr: Typescript issue more than anything. The check implicitly sets that it cannot be null in this case, but TS doesn't get that. Could do some type guarding. But I'm not entirely happy about the fromId, fromLat nullable etc types the way they are either. But It will require some more refactoring to avoid that now

PR #102, opened 2023-11-29
fix: improves search for departments and autoselects

mikaelbr:
Makes searching for departures a bit more ergonomic. Will search for departures when selecting place

- Auto search on select
- Reduces Search debounce
- Avoids revalidation on autocomplete (calling autocomplete API unnecessary amount of times without any additional value)

### DIFF
--- a/src/page-modules/departures/layout.tsx
+++ b/src/page-modules/departures/layout.tsx
@@ -53,9 +53,7 @@ function DeparturesLayout({
   const [isSearching, setIsSearching] = useState(false);
   const [geolocationError, setGeolocationError] = useState<string | null>(null);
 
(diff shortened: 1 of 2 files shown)


toretefre (approved):
Makes sense, let's try

src/page-modules/departures/layout.tsx
```diff
@@ -91,6 +89,16 @@ function DeparturesLayout({
     setIsSearching(false);
   };
 
+  const onSubmitHandler: FormEventHandler<HTMLFormElement> = (e) => {
```
adriansberg, line 92: Couldn't the doSearch just be passed as the submitHandler?

mikaelbr: Need prevent default, and other places calling doSearch doesn't have access to event which has preventdefault

PR #103, opened 2023-11-30
refactor: make departure state query handling consistant with assistant

mikaelbr:
Same naming and concept as with assistant to keep it consistent.
Same test plan also.

### DIFF
--- a/src/page-modules/departures/utils.ts
+++ b/src/page-modules/departures/utils.ts
@@ -1,0 +1,60 @@
+import {
+  TransportModeFilterState,
+  filterToQueryString,
+} from '@atb/modules/transport-mode';
+import { FromDepartureQuery } from './types';
+import { searchTimeToQueryString } from '@atb/modules/search-time';
+import { Url } from 'url';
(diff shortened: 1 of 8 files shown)


src/page-modules/departures/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+      };
+    }
+  }
+
+  const searchTimeQuery = searchTimeToQueryString(tripQuery.searchTime);
+
+  if (tripQuery.from?.layer == 'venue') {
+    return {
+      pathname: '/departures/[[...id]]',
+      query: {
```
adriansberg, line 32: Cf the TODO comment that was in the original code regarding URL encoding; is it preferable for us to encode the URL (like it is done automatically with this method) or don't and keep some legibility? I guess the URL is so complex that legibility for the user might not be a concern.

adriansberg (approved):

src/page-modules/departures/utils.ts
mikaelbr: I don't think I get the question

PR #104, opened 2023-11-30
fix: move map fullscreen button to top of page

mortennordseth:
Fixes https://github.com/AtB-AS/kundevendt/issues/15756

The "see in map" button was hard to find on smaller screens. This fixes that by moving it to the top of the page. 

image


adriansberg (approved):
Quick and easy 🚀

PR #105, opened 2023-11-30
fix: use transport sub mode color

mortennordseth:
Fixes https://github.com/AtB-AS/kundevendt/issues/15782

This fixes the issue with the wrong transport icon color for transport sub-modes. 

image


adriansberg (approved):
Very nice

PR #107, opened 2023-11-30
fix: info icon size for short wait

mortennordseth:
The info icon size for short wait was smaller than intended. 

Fixes https://github.com/AtB-AS/kundevendt/issues/15791

image


adriansberg (approved):
Who knew setting `"small"` could make it small? 😮

PR #111, opened 2023-12-01
fix: revalidation on connect for autocomplete

mikaelbr:
Some cases happening where it doesn't search due to debounce. ensure revalidate on connect to avoid


toretefre (approved):

toretefre:
🐌

PR #112, opened 2023-12-01
feat: use correct href for "Travel to/from" buttons

adriansberg:
Update the `href` for the buttons above the map to route correctly.

Fixes https://github.com/AtB-AS/kundevendt/issues/15785

### DIFF
--- a/src/components/map/map-header.tsx
+++ b/src/components/map/map-header.tsx
@@ -10,13 +10,18 @@ import {
 } from '@atb/modules/transport-mode';
 
 export type MapHeaderProps = {
-  id: string;
   name: string; // StopPlace name or address
   layer: 'address' | 'venue';
   transportModes?: TransportModeType[];
+  position: { lat: number; lon: number };
 };
 
-export function MapHeader({ id, name, layer, transportModes }: MapHeaderProps) {
+export function MapHeader({
+  name,
+  layer,
+  transportModes,
+  position,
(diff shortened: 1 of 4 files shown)


src/components/map/map-header.tsx
```diff
@@ -59,13 +64,13 @@ export function MapHeader({ id, name, layer, transportModes }: MapHeaderProps) {
         <div className={style.header__buttons}>
           <ButtonLink
             mode="interactive_0"
-            href={`/planner?travelFrom=${id}`}
+            href={`/assistant?fromLat=${position.lat}&fromLon=${position.lon}&fromLayer=${layer}`}
```
mikaelbr, line 67: I think this could also be URL and maybe easier to read?

```
            href={{ paths: '/assistant', query: { your object here} }}
```


IF i added that to the ButtonLink back in the day 🤔

adriansberg: Probably. Can look into it if we want it, but have to be later.

mikaelbr (approved):

adriansberg:
We can certainly improve on this with the ID mapping. Allow for id as query and have some more logic in the query fetcher that resolves it perhaps.

mikaelbr:
That was what I was thinking. And some of the reasons behind the refactor. But I think this is a good temporary solution that we can live with

PR #113, opened 2023-12-01
feat: fade out trips in the past

mikaelbr:
Fixes https://github.com/AtB-AS/kundevendt/issues/15783

Only for the assistant view, currently. Unkown if this should be solved in departures.


adriansberg:
The app doesn't allow searching for departures in the past, so it's not needed there. Maybe we should enforce that as well.

adriansberg (approved):
A little bit weird with the animation (x) when loading more trips, might be only me. Other than that it's looking good!

PR #114, opened 2023-12-04
refactor: filter button

mortennordseth:
I was working on https://github.com/AtB-AS/kundevendt/issues/12072 and had a chat with Christian about this button. The buttons used in the Figma sketches for the planner-web are from the app design-system and we've reused the webshop design-system buttons. 

AtB wants to redo these components to look more the same, but haven't had time to prioritize this. The components are missing states and are therefore hard to implement for now. I tried to implement a new mode, `transparent--bordered`, as a temporary solution, but using the `text-colors-primary` color as a border (as is done for the component in Figma) would give us contrast issues. The button in the travel planner web Figma sketches has overwritten these colors with #FFFFFF which would cause issues elsewhere in the app. 

I personally think it's not a good idea to expand the button component to have a new mode that only works in certain scenarios to match the Figma sketches and that we should consider revisiting the sketches instead. I tweaked the button behavior to match what it actually does and made sure the colors match between the assistant and departures pages. I think this is an improvement at least. 

Let me know if you think we should make AtB prioritize redoing the button components in Figma, but I doubt they will have time before the release. Or if you think we should add the new mode

<details>

image

image

image

image

</details>

### DIFF
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -174,7 +174,7 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
                 : t(PageText.Assistant.search.buttons.alternatives.more)
             }
             className={style.button}
-            mode={showAlternatives ? 'interactive_3' : 'interactive_2'}
+            mode="interactive_2"
             onClick={() => setShowAlternatives(!showAlternatives)}
             icon={{ right: <MonoIcon icon="actions/Adjust" /> }}
           />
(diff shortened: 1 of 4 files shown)


src/page-modules/assistant/layout.tsx
```diff
@@ -168,9 +168,13 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
...
           <Button
-            title={t(PageText.Assistant.search.buttons.alternatives.title)}
+            title={
+              showAlternatives
+                ? t(PageText.Assistant.search.buttons.alternatives.less)
+                : t(PageText.Assistant.search.buttons.alternatives.more)
+            }
             className={style.button}
-            mode="interactive_2"
+            mode={showAlternatives ? 'interactive_3' : 'interactive_2'}
```
adriansberg, line 177: Isn't interactive 3 a green color for FRAM?

adriansberg (approved):

PR #115, opened 2023-12-04
refactor: improve transport mode filter

adriansberg:
Rework the transport mode filter based on feedback from user testing and new design.

<p>

image
image
image
image
image

</p>
</details> 

Have some concerns for the hover colors for FRAM light mode, have raised a question to Julia.

<p>

image

</p>
</details> 

Closes https://github.com/AtB-AS/kundevendt/issues/15753

### DIFF
--- a/src/modules/transport-mode/filter/filter.module.css
+++ b/src/modules/transport-mode/filter/filter.module.css
@@ -21,14 +21,14 @@
 }
 
 .transportMode label {
-  padding: var(--spacings-xSmall) var(--spacings-xSmall) var(--spacings-xSmall) var(--spacings-small);
+  padding: var(--spacings-xSmall) var(--spacings-small) var(--spacings-xSmall) var(--spacings-medium);
   border-radius: var(--border-radius-circle);
   background-color: var(--interactive-interactive_2-default-background);
   color: var(--interactive-interactive_2-default-text);
   transition: border-color 150ms, background-color 150ms;
   cursor: pointer;
   display: flex;
-  gap: var(--spacings-xSmall);
+  gap: var(--spacings-small);
   align-items: center;
   height: 1.75rem;
 }


mortennordseth:
I see that the gap between the text and check icon is 6 px in the Figma sketches and I guess you've used 4 px because 6 px does not exist in the design system. I think it's a bit small, to be honest. Would it be better to use `var(--spacings--small)` instead? Also, there should be 12px padding on the left side I see. 

Currently:

image

With gap `var(--spacings-small)` and `padding: var(--spacings-xSmall) var(--spacings-small) var(--spacings-xSmall) var(--spacings-medium);`

image

Just a thought, could also consider having the same padding on the left and right side? 🤔 
`padding: var(--spacings-small) var(--spacings-medium);`

image

mortennordseth:
Also, are we doing this in a separate PR?
From the issue:
- Change the behavior when changing from all to a specific filter
- Debouncing (update search after x time after filter have changed)

adriansberg:
> With gap var(--spacings-small) and padding: var(--spacings-xSmall) var(--spacings-small) var(--spacings-xSmall) var(--spacings-medium);

Adding this.

> Just a thought, could also consider having the same padding on the left and right side? 🤔

IMO, the checkmark circle fits better with uniform spacing above and to the right.

> Also, are we doing this in a separate PR?

> > Change the behavior when changing from all to a specific filter

Not sure what this entails, do you know?

> > Debouncing (update search after x time after filter have changed)

This we can do as a part of https://github.com/AtB-AS/kundevendt/issues/15757

mortennordseth:
> IMO, the checkmark circle fits better with uniform spacing above and to the right.

Fine by me 👍 

> Not sure what this entails, do you know?

My best guess is that when having all selected and clicking "bus" we only want to have "bus" selected 🤔 Not sure though

adriansberg:
You might be right. So essentially an opposite behaviour when all are selected, but should behave as it should when not all are selected.

Can you confirm/deny, @mikaelbr?

mikaelbr:
Yes I think this behaviour is the conclusion of what we saw in user testing. So clicking on Bus when having "all" selected will mean going to only BUS selected.

mikaelbr:
Can be done in a separate pr

src/modules/transport-mode/filter/filter.tsx
```diff
@@ -98,18 +98,37 @@ function FilterCheckbox({ option, checked, onChange }: FilterCheckboxProps) {
...
+    <svg
+      width="12"
+      height="12"
+      viewBox="0 0 12 12"
+      fill="none"
+      className={className}
+    >
+      <path
+        d="M10.8979 3.77299L4.89795 9.77299C4.84569 9.82543 4.78359 9.86704 4.71522 9.89543C4.64685 9.92382 4.57354 9.93843 4.49951 9.93843C4.42548 9.93843 4.35217 9.92382 4.2838 9.89543C4.21543 9.86704 4.15333 9.82543 4.10107 9.77299L1.47607 7.14799C1.42375 7.09566 1.38224 7.03355 1.35393 6.96518C1.32561 6.89682 1.31104 6.82355 1.31104 6.74955C1.31104 6.67555 1.32561 6.60228 1.35393 6.53392C1.38224 6.46555 1.42375 6.40344 1.47607 6.35111C1.5284 6.29879 1.59051 6.25728 1.65888 6.22897C1.72724 6.20065 1.80051 6.18607 1.87451 6.18607C1.94851 6.18607 2.02178 6.20065 2.09014 6.22897C2.15851 6.25728 2.22062 6.29879 2.27295 6.35111L4.49998 8.57814L10.102 2.97705C10.2077 2.87138 10.351 2.81201 10.5004 2.81201C10.6499 2.81201 10.7932 2.87138 10.8989 2.97705C11.0046 3.08272 11.0639 3.22604 11.0639 3.37549C11.0639 3.52493 11.0046 3.66825 10.8989 3.77393L10.8979 3.77299Z"
+        fill="var(--static-background-background_2-text)"
```
mikaelbr, line 130: This can also be done using CSS with classname as above so it is solved one place.

mikaelbr (approved):
Should change click order as mentioned in the comments and maybe also let css file handle fill to be consistent. But all in all, looks like a good improvement!

mortennordseth (approved):

PR #116, opened 2023-12-04
style: fix colors of header and footer

adriansberg:
Revert the colors of the header and footer to the same colors as the webshop.

Partly solves https://github.com/AtB-AS/kundevendt/issues/15755. This solves the colors of the header/footer, not the links.


mortennordseth:
Should we do something about this? 🤔 It might not be that easy though

image

adriansberg:
I assume you mean the background color of the header being the same as the background color of the search area? Not the simplest thing to just change color on, no.

mortennordseth:
Yep. I don't know if it's a blocker for this PR, but let's double-check with Julia? 🤔 They do have different colors in the Figma sketches, so I guess it's just an oversight that these are the same in dark mode for FRAM. It might be an option to use accent 3 or something similar for the filter wrapper?

mikaelbr:
Top navigation and search input should have same color now to match webshop and new frammr, if that's the thing we're discussing. Or is it the filter background?

adriansberg:
Filter background in dark mode, yes.

mikaelbr (approved):

PR #121, opened 2023-12-07
feat: tariff zones in map

mortennordseth:
This adds functionality to connect to Firestore and draw tariff zone boundaries on the map. 

We'll have to look into caching the data from Firestore. I'm not including caching in this PR to allow us to move forward with other issues that requires the Firebase setup. 

<details>

image

</details>

Closes https://github.com/AtB-AS/kundevendt/issues/15814
Closes https://github.com/AtB-AS/kundevendt/issues/15775


src/components/map/use-map-tariff-zones.ts
```diff
@@ -0,0 +1,107 @@
+import { useCallback, useEffect, useState } from 'react';
+import { Language, useTranslation } from '@atb/translations';
+import { TariffZone } from '@atb/modules/firebase/types';
+import mapboxgl from 'mapbox-gl';
+import getTariffZones from '@atb/modules/firebase';
+import { getReferenceDataName } from '@atb/utils/reference-data';
+import turfCentroid from '@turf/centroid';
```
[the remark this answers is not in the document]
mikaelbr: Why not use `@turf/turf`?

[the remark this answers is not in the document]
mortennordseth: I was thinking that we didn't want to add all functionality from @turf, but I guess the bundler handles that 🤔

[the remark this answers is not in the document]
mikaelbr: Yep, should be able to codesplit/dead code eliminate there. Can use bundle analyzer tools to verify

src/components/map/use-map-tariff-zones.ts
```diff
@@ -0,0 +1,112 @@
...
+          properties: {},
+          geometry: {
+            type: 'Polygon' as const,
+            coordinates: tariffZone.geometry.coordinates,
+          },
+        },
+        {
+          type: 'Feature' as const,
+          properties: {
+            name: getReferenceDataName(tariffZone, language),
```
mortennordseth, line 100: Would it be a good idea to prefix the zone names with "Zone"? E.g. "Zone Molde" so that we are more clear in what's shown on the map

src/modules/firebase/tariff-zones.ts
```diff
@@ -0,0 +1,27 @@
...
+  const firestore = getFirestore(app);
+  const docRef = doc(firestore, 'configuration', 'referenceData');
+
+  const snapshot = await getDoc(docRef);
+
+  if (snapshot.exists()) {
+    const data = snapshot.data();
+    const tariffZonesFromFirebase = data.tariffZones;
+
+    const potential = JSON.parse(tariffZonesFromFirebase);
```
adriansberg, line 15: Can this fail?

src/components/map/use-map-tariff-zones.ts
```diff
@@ -0,0 +1,112 @@
...
+  const [tariffZones, setTariffZones] = useState<TariffZone[]>([]);
+
+  useEffect(() => {
+    const fetchTariffZones = async () => {
+      const zones = await getTariffZones();
+      setTariffZones(zones);
+    };
+
+    fetchTariffZones();
+  }, []);
```
adriansberg, line 21: I guess this will be replaced at a later point with e.g. `useSWR`? Even though the caching might not work as we want it to with it, I think it's cleaner to use it instead of having a state and a `useEffect`. Should probably confirm whether it's unintended to use `useSWR` for non-`fetch`ing stuff too.

src/modules/firebase/tariff-zones.ts
mortennordseth: Hm, it could probably. The document is a JSON object, but that could be formatted wrong of course

mortennordseth: I've added a try catch to the parsing

PR #122, opened 2023-12-07
feat: sort quays

mortennordseth:
This adds sorting of quays by public code and departures. A quay with no departures are placed at the bottom. 

image

Closes https://github.com/AtB-AS/kundevendt/issues/15926
Closes https://github.com/AtB-AS/kundevendt/issues/15894

### DIFF
--- a/src/page-modules/departures/server/journey-planner/index.ts
+++ b/src/page-modules/departures/server/journey-planner/index.ts
@@ -166,6 +166,21 @@ export function createJourneyApi(
         throw validated.error;
       }
 
+      validated.data.quays.sort((a, b) => {
+        // Place quays with no departures at the end
+        if (a.departures.length === 0) return 1;
+        if (b.departures.length === 0) return -1;
+
+        if (!a.publicCode) return 1;
+        if (!b.publicCode) return -1;
+
+        if (parseInt(a.publicCode) && parseInt(b.publicCode)) {
+          return parseInt(a.publicCode) - parseInt(b.publicCode);
+        }
+
+        return a.publicCode.localeCompare(b.publicCode);
+      });
+
       return validated.data;
     },
 


src/page-modules/departures/server/journey-planner/index.ts
```diff
@@ -166,6 +166,21 @@ export function createJourneyApi(
...
+        // Place quays with no departures at the end
+        if (a.departures.length === 0) return 1;
+        if (b.departures.length === 0) return -1;
+
+        if (!a.publicCode) return 1;
+        if (!b.publicCode) return -1;
+
+        if (parseInt(a.publicCode) && parseInt(b.publicCode)) {
+          return parseInt(a.publicCode) - parseInt(b.publicCode);
+        }
```
mikaelbr, line 179: Maybe these can be combined to parse and check value as an integrated step?

mortennordseth: I'm not sure I understand 😅 Are you thinking of avoiding doing `parseInt` twice per variable? 🤔

mikaelbr, line 179: Something like

```suggestion
        if (a.departures.length === 0) return 1;
        if (b.departures.length === 0) return -1;
        
        const publicA = parseInt(a.publicCode, 10);
        const publicB = parseInt(b.publicCode, 10);
        
        if (Number.isNaN(publicB) || Number.isNaN(publica)) {
          return a.publicCode.localeCompare(b.publicCode);
        }

        return publicA - publicB;
```

There is something odd with the `if (parseInt(a.publicCode) && parseInt(b.publicCode))`, because if it is not parsed it is NaN so it will be if NaN && NaN which is technically false, but odd. But also if public code is 0 this will be false. Again, probably not any real world consequences in this specific case, but the check it self is kind of odd, so I wouldn't say it is a good check to have of principle.

mikaelbr: Also, do you know what type of mechanism is good to check edge cases for things like this? 😅

mortennordseth: .. I can write some tests 😄

adriansberg (approved):
LGTM

PR #123, opened 2023-12-07
feat: component to toggle between the assistant and departures pages. 

jonasbrunvoll:
Fixes: https://github.com/AtB-AS/kundevendt/issues/15755

- Added a component to toggle between the assistant and departures pages. 
- Removed excessive HeaderLinks from menu.

https://github.com/AtB-AS/planner-web/assets/59939294/cb9ff673-47c0-4363-8482-8c1b9eb70817


src/page-modules/assistant/layout.tsx
```diff
@@ -81,11 +82,12 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
           transition={{ duration: 0.25, ease: [0.04, 0.62, 0.23, 0.98] }}
           className={style.main}
         >
+          <AssistantDepartureToggleButton />
+
           <div className={style.input}>
             <Typo.p textType="body__primary--bold" className={style.heading}>
               {t(PageText.Assistant.search.input.label)}
             </Typo.p>
-
```
adriansberg, line 88: Have you set up some auto-formatting that removes these lines? 

We use Prettier in this project, so if you use e.g. VS Code you should configure format on save to use the Prettier config.

src/components/toggle-button/assistant-departure-toggle-button.module.css
```diff
@@ -0,0 +1,31 @@
...
+  display: flex;
+  flex-direction: row;
+  grid-column: span 2;
+  border-bottom: 1px solid var(--static-background-background_accent_0-text); /* Standard border-bottom-stil */
+  gap: var(--spacings-large);
+}
+
+.link {
+  text-decoration: none;
+  position: relative; /* Legg til position: relative */
```
adriansberg, line 11:
```suggestion
  position: relative;
```

src/page-modules/assistant/layout.tsx
jonasbrunvoll: Thanks! For some reason .prettier was not configured in my vs code. Configured it back now.

.gitignore
```diff
@@ -20,6 +20,7 @@
 .DS_Store
 *.pem
 .idea/
+.vscode/
```
adriansberg, line 23: I don't use VS Code currently, but I was under the impression that the `.vscode/` folder could be used to share project specific settings (like formatOnSave etc.) and recommended extensions. By ignoring this folder we lose this, don't we?

src/components/toggle-button/assistant-departure-toggle-button.tsx
```diff
@@ -0,0 +1,80 @@
...
+  },
+  {
+    name: 'departures',
+    title: PageText.Assistant.search.buttons.find.title,
+    link: '/departures',
+  },
+];
+
+const setInitialActiveLink = (href: string | undefined) => {
+  return href ? links[1].name : links[0].name;
```
adriansberg, line 32: Should this check the `links` array for matching href and link instead of indexing it?

src/components/toggle-button/assistant-departure-toggle-button.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+const links: LinkProps[] = [
+  {
+    name: 'assistant',
+    title: PageText.Assistant.title,
+    link: '/assistant',
+  },
+  {
+    name: 'departures',
+    title: PageText.Assistant.search.buttons.find.title,
```
adriansberg, line 26: This could get its own translation entry, I think.

Maybe both of these (L21, L26) should be in ComponentText instead of PageText?

src/components/toggle-button/assistant-departure-toggle-button.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const { t } = useTranslation();
+  const [activeLink, setActiveLink] = useState(() =>
+    setInitialActiveLink(href),
+  );
+
+  return (
+    <div className={style.container}>
+      {links.map((link, index) => (
+        <Link
+          key={index}
```
adriansberg, line 49: Indexes are generally not the best to use for keys. @mikaelbr probably has a blog post to link.

src/components/toggle-button/index.ts
```diff
@@ -0,0 +1 @@
+export { default as AssistantDepartureToggleButton } from './assistant-departure-toggle-button';
```
adriansberg, line 1: Is this a toggle button, or is it more like a tab link?

src/page-modules/assistant/layout.tsx
adriansberg: Great! Maybe we can add those deleted blank lines back?

src/components/toggle-button/assistant-departure-toggle-button.module.css
```diff
@@ -0,0 +1,31 @@
...
+
+.link {
+  text-decoration: none;
+  position: relative;
+  padding-bottom: var(--spacings-medium);
+  color: var(--static-background-background_accent_0-text);
+}
+
+.active {
+  opacity: 100%;
```
adriansberg, line 17: Where is `opacity` set to 0?

.gitignore
jonasbrunvoll: After configuring VS Code to use the .prettierrc file following this project, the local setting.json file, which is located in the .vscode folder, was initially added commited. To avoid pushing it, I added the folder to the .gitignore file. Since I did not find any .vscode direktory in the repo, I assumed this was ok fix for avoiding pushing unnecessary code.

adriansberg: If it's the setting for which formatter to use for the language, I think it's okay to commit that actually. It enforces the correct formatter for other new developers (and current ones).

src/components/toggle-button/assistant-departure-toggle-button.tsx
jonasbrunvoll: I the href is defined as an attribute, the initial href should be links[1].name ("/departures"). If else, it should be "/assistant". 

That this logic is perhaps unnecessarily difficult and can most likely be written in a more intuitive way.

src/components/toggle-button/assistant-departure-toggle-button.tsx
jonasbrunvoll: Agreed! I didn't realize that ComponentText existed before you pointed it out with this comment. Will fix.

src/components/toggle-button/index.ts
jonasbrunvoll: It is more like a tab link, currently used to switch between two URL´s. Do you have a suggestion to a more precise component name?

src/components/toggle-button/assistant-departure-toggle-button.tsx
```diff
@@ -0,0 +1,80 @@
...
+          className={`${style.link} ${
+            activeLink === link.name ? style.active : ''
+          }`}
+          href={link.link}
+          onClick={() => handleClick(link.name, setActiveLink)}
+        >
+          <Typo.p
+            textType={`${
```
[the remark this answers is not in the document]
jonasbrunvoll: Will fix👍

src/components/toggle-button/assistant-departure-toggle-button.tsx
jonasbrunvoll: Can you send me the link?

.gitignore
mikaelbr: Agreed! It adds value to have the .vscode directory. Can also add suggested extensions in that

src/components/toggle-button/assistant-departure-toggle-button.tsx
mikaelbr: This is usually a very good point and something to keep in mind every time you use `key` attribute. I think, however, as this uses static object this isn't a big issue here, as index will never change

src/components/toggle-button/assistant-departure-toggle-button.tsx
mikaelbr: create an image of a generic male figure with a strong physique, wearing a Batman costume

mikaelbr: This is usually a very good point and something to keep in mind every time `key` is used. I think however, for this particular case, this won't be an issue as the object is static and thus the index will never update.

src/components/toggle-button/index.ts
mikaelbr: I usually try to end "actions" (so things happening locally on site that doesn't trigger navigational URL change) with "Button" and navigational changes (links that trigger URL change) with "Link". Implying what kind of change they represent, and also specifying how they should be used.

src/components/tab-link/tab-link.tsx
```diff
@@ -0,0 +1,83 @@
...
+      {links.map((link, index) => (
+        <Link
+          key={index}
+          className={style.link}
+          href={link.link}
+          onClick={() => handleClick(link.name, setActiveLink)}
+        >
+          <Typo.p
+            textType={
+              activeLink === link.name ? 'body__primary--bold' : 'body__primary'
```
mikaelbr, line 61:
```suggestion
              activeLinkHref === link.link ? 'body__primary--bold' : 'body__primary'
```

src/components/tab-link/tab-link.tsx
```diff
@@ -0,0 +1,83 @@
...
+            }
+          >
+            {t(link.title)}
+          </Typo.p>
+          <AnimatePresence initial={false}>
+            {activeLink === link.name && (
```
mikaelbr, line 67:
```suggestion
            {activeLinkHref === link.link && (
```

src/components/tab-link/tab-link.tsx
```diff
@@ -0,0 +1,83 @@
...
+const links: LinkProps[] = [
+  {
+    name: 'assistant',
+    title: ComponentText.TabLink.assistant,
+    link: '/assistant',
+  },
+  {
+    name: 'departures',
+    title: ComponentText.TabLink.departures,
+    link: '/departures/[[...id]]',
```
mikaelbr, line 31: maybe this is more clear as 'href`?

src/components/tab-link/tab-link.tsx
```diff
@@ -0,0 +1,83 @@
...
+                transition={{ duration: 0.25 }}
+              />
+            )}
+          </AnimatePresence>
+        </Link>
+      ))}
+    </div>
+  );
+};
+export default AssistantDepartureToggle;
```
mikaelbr, line 83: if the module is named TabLink this should be named TabLink also, I think

src/components/tab-link/tab-link.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+const handleClick = (linkName: string, setActiveLink: Function) => {
+  setActiveLink(linkName);
+};
+
+const AssistantDepartureToggle = ({ href }: AssistantDepartureTogglProp) => {
+  const { t } = useTranslation();
+  const [activeLink, setActiveLink] = useState(() =>
+    setInitialActiveLink(href),
+  );
```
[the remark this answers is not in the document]
mikaelbr: Oh, looking closer you actually pass `href` as prop, so you can just use that.

src/components/tab-link/tab-link.tsx
jonasbrunvoll: Agreed!

src/components/tab-link/tab-link.tsx
```diff
@@ -0,0 +1,83 @@
...
+];
+
+const setInitialActiveLink = (href: string | undefined) => {
+  const link = links.find((link) => link.link === href);
+  return link?.name;
+};
+
+const handleClick = (linkName: string, setActiveLink: Function) => {
+  setActiveLink(linkName);
+};
```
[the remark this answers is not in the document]
jonasbrunvoll: Removed👍

mikaelbr (changes requested):
I think this should be updated: https://github.com/AtB-AS/planner-web/pull/123#discussion_r1420901086

src/components/tab-link/tab-link.tsx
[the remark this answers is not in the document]
jonasbrunvoll: Fixed. The activePath is now passed as a prop from the parent component to simplify the logic.

src/components/tab-link/tab-link.tsx
```diff
@@ -26,52 +28,44 @@ const links: LinkProps[] = [
...
-          className={style.href}
-          href={link.href}
-          onClick={() => setActivePath(rewritePathIfDynamic(router.pathname))}
-        >
-          <Typo.p
-            textType={
-              activePath === link.href ? 'body__primary--bold' : 'body__primary'
-            }
+      {links.map((link, index) => {
+        let isActive = rewritePathIfDynamic(activePath) === link.href;
```
mikaelbr, line 41: Could this be simpler and more generic like:

```
activePath.startsWith(link.href)
```

mikaelbr:
Should maybe add an standard indicator to set default active mode for assistant (when on `/`, on initial page load)

src/components/tab-link/tab-link.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+type LinkProps = {
+  title: TranslatedString;
+  href: string;
+};
+
+const links: LinkProps[] = [
+  {
+    title: ComponentText.TabLink.assistant,
+    href: '/' || '/assistant',
```
mikaelbr, line 23: this would evaluate in place, so essentially `href` will hold the value `/` as `'/' || '/assistant'` will say truthy value `/` or truthy value `/assistant` where `'/' == true` and will cause the conditional to return. You can check this in the REPL or in DevTools to see what it returns to.

But this will again cause the startsWith to always be true

jonasbrunvoll: Ah, I see. I´ll replaced '/' || '/assistant' with the regex expression. By the way, is it okay to leave comments in the code? I think regex expressions often can be a bit difficult to understand, and therefore have a habit of commenting the logic.

mikaelbr: From commit message

> Removed regex logic from code. This solution seems fine as it sets the correct active path for '/', '/assistant', and 'departures'

I promise you. It doesn't work. Or at least it doesn't work as intended, and if you observe the expected result when testing it, it works by "luck". JavaScript/TypeScript doesn't function this way. 

See what the result of this line is in the console:

Screenshot 2023-12-11 at 16 54 20

Without testing it out I think you see it working as isActive will be `true` for both cases, but the underline element with layoutId is reused by framer-animation and only shown once. Since the `/departures` case happens as the last case the underline element will be reused and position under departures-link.

jonasbrunvoll: I chose to move away from the regex expression as it became more complicated than necessary. Instead, type check was added to ensure that only valid routs can be passed as attributes in <TabLink/>. With this approach, the problem of having to distinguish between if either '/'  or '/assistant' was passed in as a prop is avoided.

src/components/tab-link/tab-link.tsx
```diff
@@ -0,0 +1,65 @@
...
+  useTranslation,
+} from '@atb/translations';
+import { Typo } from '../typography';
+import { AnimatePresence, motion } from 'framer-motion';
+
+type TabLinkProps = {
+  activePath: Route;
+};
+
+type LinkProps = {
```
mortennordseth, line 15: In other parts of the application, we name component prop types with the suffix `Props`. Since this type isn't a prop to the component, maybe we could rename it to `TabLink` or something similar? 
```suggestion
type TabLink = {
```

jonasbrunvoll: Noted. Renamed the type now.

src/layouts/shared/page-header.tsx
```diff
@@ -100,22 +100,6 @@ function Navigation({
       >
         <ul>
           <li></li>
```
adriansberg, line 102: These changes are good, this is more a general question that is related to this. Should we populate the header when the user is on the details page, or is it okay for the user to go back to the assistant/departure page then change view?

src/page-modules/assistant/layout.tsx
```diff
@@ -81,11 +82,12 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
           transition={{ duration: 0.25, ease: [0.04, 0.62, 0.23, 0.98] }}
           className={style.main}
         >
+          <TabLink activePath={`/assistant`} />
```
mikaelbr, line 85:
```suggestion
          <TabLink activePath="/assistant" />
```

src/page-modules/departures/layout.tsx
```diff
@@ -65,11 +65,12 @@ function DeparturesLayout({ children, fromQuery }: DeparturesLayoutProps) {
           transition={{ duration: 0.25, ease: [0.04, 0.62, 0.23, 0.98] }}
           className={style.main}
         >
+          <TabLink activePath={'/departures'} />
```
mikaelbr, line 68: The template literal syntax and `{}` is in this case just syntactic noise. Should try to always use regular strings directly when possible
```suggestion
          <TabLink activePath="/departures" />
```

src/page-modules/assistant/layout.tsx
mikaelbr: This is a good and simple solution by the way! Good jobs

mikaelbr (approved):
Looking good!

PR #129, opened 2023-12-10
fix: show correct trip duration based on rounded times

adriansberg:
Show correct trip duration based on rounded times. Previously the trip duration did not always correspond with the formatted times of departure and arrival. E.g. if a trip departs at 16:04 and arrives at 16:48, the duration might have said 42 minutes since the departure time rounds down (floors) and arrival times are rounded up (ceil)¹. 

So the solution was to base the duration of the trip on the formatted times of departure and arrival, ignoring the real duration of the trip. That way we get a duration corresponding with the times shown.

[1]: AtB have spent deciding how to round the departure and arrival times in the app, so that should be done the same in the travel planner; "We have chosen to display rounding that minimizes the likelihood of the user missing the bus - so departures are rounded down, while arrivals are rounded up."

Closes https://github.com/AtB-AS/kundevendt/issues/15928

### DIFF
--- a/src/page-modules/assistant/trip/trip-pattern/trip-pattern-header/index.tsx
+++ b/src/page-modules/assistant/trip/trip-pattern/trip-pattern-header/index.tsx
@@ -2,7 +2,7 @@ import { TripPattern, Quay } from '../../../server/journey-planner/validators';
 import style from './trip-pattern-header.module.css';
 import { Typo } from '@atb/components/typography';
 import { useTranslation, PageText } from '@atb/translations';
-import { secondsToDuration } from '@atb/utils/date';
+import { formatTripDuration } from '@atb/utils/date';
 import { flatMap } from 'lodash';
 import { getNoticesForLeg } from './utils';
 import { RailReplacementBusMessage } from './rail-replacement-bus';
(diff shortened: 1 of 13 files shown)


src/utils/date.ts
```diff
@@ -331,3 +331,26 @@ function getShortHumanizer(
 
   return humanizer(ms, opts);
 }
+
+export function formatTripDuration(
```
adriansberg, line 335: Not sure where this function should be placed. Used for the details page (header and below map) and for the trip pattern in the search results view.

src/page-modules/assistant/trip/trip-pattern/trip-pattern-header/index.tsx
```diff
@@ -15,7 +15,11 @@ type TripPatternHeaderProps = {
 export function TripPatternHeader({ tripPattern }: TripPatternHeaderProps) {
   const { t, language } = useTranslation();
 
-  const duration = secondsToDuration(tripPattern.duration, language);
```
mortennordseth, line 18: Should probably remove `duration` from the trip pattern object now if it's not in use

src/utils/date.ts
mortennordseth: I'm not sure either, but an alternative is to create a new file in the root utils folder 🤔 This might be fine though

mortennordseth (approved):
Looks good to me 👍 One comment that you can consider

mortennordseth:
I missed that the tests are failing, should fix that of course 😅

src/page-modules/assistant/trip/trip-pattern/trip-pattern-header/index.tsx
adriansberg: Removed duration.

PR #130, opened 2023-12-11
chore: upgrades all dependencies

mikaelbr:
Upgrades dependencies and fixes some issues with older versions of next (and more).

Fixes https://github.com/AtB-AS/kundevendt/issues/15811


package.json
```diff
@@ -20,73 +20,67 @@
...
-    "next": "13.4.19",
+    "mapbox-gl": "^3.0.1",
+    "next": "14.0.4",
     "next-router-mock": "^0.9.10",
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "react-syntax-highlighter": "^15.5.0",
     "swr": "^2.2.2",
-    "typescript": "5.2.2",
+    "typescript": "5.3.3",
```
jonasbrunvoll, line 53: Is there a reason for why typescript is added to dependencies and not devDependencies? My understanding is that typescript is used in the development phase to maintain better control over the code, but that it is redundant to include it in production, as it is the compiled JavaScript code that is used to run the application.

mikaelbr: No. To be honest, I don't really know why it is in "dependencies" instead of "devDependencies". But the difference between these actually doesn't impact a whole lot. This only affects wether they should be installed or not, for instance when doing `yarn install --production` you only install `"dependency"` and not `"devDependency"`. But in most cases you only ever do regular installs because you have to build bundles. So in most cases this is only ever used as an indication and sort of semantic categorization in code bases to indicate something on criticality I think. So for instance if something is in `devDependency` you pretty much know that if it compiles or runs locally you are good, and it is easier to remove.

But things included in `dependency` won't automatically be included in production. That is the bundlers task to include things in the compiled code if something refers to it in the code.

But I cleaned up some of the dependencies that for some reason was in `dependency`. (most common reason is forgetting the `-D` flag while installing)

jonasbrunvoll: Ok, got it! Thanks for the clarification😊

PR #132, opened 2023-12-11
feat: show duration of walking leg

adriansberg:
Show the duration of a walking leg in the search results view. The icon for a leg that is by foot is now showing how many minutes of walking it is (showing a minimum of 1 if the duration is less than a minute).

Closes https://github.com/AtB-AS/kundevendt/issues/15924

### DIFF
--- a/src/modules/transport-mode/icon/index.tsx
+++ b/src/modules/transport-mode/icon/index.tsx
@@ -1,13 +1,14 @@
 import { TransportModeGroup, TransportSubmodeType } from '../types';
-
 import { ContrastColor, Theme } from '@atb-as/theme';
 import MonoIcon, { MonoIconProps } from '@atb/components/icon/mono-icon';
 import { useTheme } from '@atb/modules/theme';
-
 import { useTranslation } from '@atb/translations';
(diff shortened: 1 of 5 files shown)


src/modules/transport-mode/icon/index.tsx
```diff
@@ -56,27 +57,57 @@ export function TransportMonoIcon({ mode }: TransportIconProps) {
 export type TransportIconWithLabelProps = {
   mode: TransportModeGroup;
   label?: string | null;
+  duration?: number | null;
```
mikaelbr, line 60: Does this really need to have both undefined AND null?

mikaelbr (approved):
To minor changes, other than that it looks good

src/modules/transport-mode/icon/index.tsx
adriansberg: No, probably not. Same with label I suppose, though in that case the `publicCode` (which is the main use for the label) is nullable and can be .

adriansberg: Removed `null` from both. A bit more verbose when passing the label though.

PR #134, opened 2023-12-11
breaking-change(widget): Build specific versions of widget

mikaelbr:
Does the following:

- Build specific version of the widget and storing them with long cache time.
- Lists all versions of the widget (with release date)
- Bundle for all orgs in build time checking in to repo

Checking in all orgs in the repo as long cached bundles. But using lz-string to just mask obvious org style imports. The JS bundles are the same, but have different styling values. Documentation shows only relevant org assets. 

Fixes https://github.com/AtB-AS/kundevendt/issues/14939

Screenshot 2023-12-13 at 09 57 11
Screenshot 2023-12-13 at 09 57 04


mortennordseth (approved):
It looks good to me 👍

mikaelbr:
Having some issues with resolving public files in production envs. So looking into that

PR #140, opened 2023-12-13
feat: disable the option to select previous dates and/or times when searching for departures.

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/15893

### Background

This feature will prevent users from being able to select earlier dates and/or times when searching for departures. By preventing this, users will not be able to mistakenly retrieve previous departures, which in turn can be confusing and lead to misunderstandings. 

#### Illustrations
<details>
Screenshot 2023-12-13 at 15 13 55

</details>

### Proposed solution

The proposed solution does prevents the user from setting past dates and/or times when searching after departures. The solution also automatically resets the selected date and time to current date and time, whenever necessary when reselecting from a date further forward in time, to an earlier date.    

### Acceptance Criteria

- [ ] Prevent user from being able to select past dates. 
- [ ] When reselecting from a future date to the current date, the time should automatically be set to the current time.  
- [ ] When reselecting from a future year to the current year, both the date and the time should automatically be set to the current date and time.


jonasbrunvoll:
I have not added any test yet, but assume I should add some to ensure that unintended behavior is noted. Should I make tests for only the rendered component, the internal arrow functions I added, or both?

src/modules/search-time/selector/index.tsx
```diff
@@ -43,14 +43,52 @@ export default function SearchTimeSelector({
...
+  };
+
+  const addCurrentDateToTime = (time: string) => {
+    const [hours, minutes] = time.split(':').map(Number);
+    const now = new Date();
+    now.setHours(hours, minutes, 0, 0);
+    return now;
+  };
+
+  const isPastDate = (selectedDate: string) => {
```
mikaelbr, line 68: I think this exists in date utils from before?

mikaelbr:
Test-wise you can use react testing library to render a component and try setting value of input box and verifying that the value doesn't update as expected.

src/modules/search-time/selector/index.tsx
```diff
@@ -43,14 +43,52 @@ export default function SearchTimeSelector({
...
+    const now = addCurrentDateToTime(time);
+    return now < selectedDate;
+  };
+
   const internalOnDateChange = (event: ChangeEvent<HTMLInputElement>) => {
     if (!event.target.value) return;
+    if (isPastDate(event.target.value)) return;
 
-    setSelectedDate(new Date(event.target.value));
+    setSelectedDate(addCurrentTimeToDate(event.target.value));
```
adriansberg, line 91: Should this be `addCurrentDateToTime`?

adriansberg:
Currently this disables the option to select a time (and date, but not as relevant possibly) at all times, regardless of if it is on the departures page or the assistant page. IMO the user should be able to search in the past, at least for the time, when searching for travels.

jonasbrunvoll:
> Currently this disables the option to select a time (and date, but not as relevant possibly) at all times, regardless of if it is on the departures page or the assistant page. IMO the user should be able to search in the past, at least for the time, when searching for travels.

Yes, I interpreted it as desirable to prevent the user the search after past departures on both the assistant and the departures page. Should I make it possible for the user to search after past departures? In that case, should the user be allowed to search after past departures om the assistant or the departures page? Or both? Allowing it for one but not the other can be confusing.

adriansberg:
Thinking about it, it might be okay to limit the date to minimum the current, but the time can be any time. I think the main issue with departures and searching in the past is that there is no data from previous days, but there is data from the current date (back to 00:00 I guess, needs confirming). And for travel search, how often is it needed to search on previous dates/days? I can see myself setting the time to be earlier, to e.g. see the original departure of a trip I'm on or something.

src/modules/search-time/selector/index.tsx
jonasbrunvoll: It seems that whether new Date() or addCurrentTimeToDate() is used here actually does not affect the behavior of the component. I can change back to using new Date(), as addCurrentTimeToDate() is excessive and adds unecessery complexity.

adriansberg: Agree. The date input only cares about the first date part of the Date, so the time does not matter here.

src/modules/search-time/selector/index.tsx
jonasbrunvoll: I'm not entirely sure what you are referring to as already existing here. Do you mean the isPastDate function?

jonasbrunvoll:
Agree. I can also see myself setting the time to be earlier, to search after the original departure of a trip I'm on. I´ll make it possible to search at any given time from the current date and forward. But, I still keep the logic to set the time automatically to the present if one sets the date from a future date back to today's date, as in my opinion it affects the user experience positively.

src/modules/search-time/selector/index.tsx
adriansberg: There's a function in `src/utils/date.ts`, `isInPast()`, that might work as you want here.

jonasbrunvoll: I tried to use isInPast() instead of  isPastDate(). This is what i found out: 
The event.target.value is received on the format 'yyyy-MM-dd'. When using isInPast(), this string is converted to date format. Due to some milliseconds that is being added during the conversion, the current date is regarded as a past date. 

We can use the isInPast() if we want to, but this still requires some additional logic to ensure that the current date is not view as a past date. I therefore suggest that we keep using the isPastDate().

src/modules/search-time/selector/index.tsx
```diff
@@ -43,12 +43,33 @@ export default function SearchTimeSelector({
...
+  };
+
+  const resetToCurrentDate = () => {
+    setSelectedDate(new Date());
+  };
+
+  const isPastDate = (selectedDate: string) => {
+    const today = new Date().toISOString().split('T')[0];
+
+    // If reselecting from (selectedYear > current year) to current year, the date will autmatically be reset to the current date.
```
adriansberg, line 60:
```suggestion
    // If reselecting from (selectedYear < current year) to current year, the date will automatically be reset to the current date.
```

src/modules/search-time/selector/index.tsx
```diff
@@ -43,12 +43,33 @@ export default function SearchTimeSelector({
...
+    if (selectedDate.substring(0, 4) < initialDate.getFullYear().toString()) {
+      resetToCurrentDate();
+    }
+
+    // To ensure that the time is not past whenever reselecting back the current date.
+    if (selectedDate <= today) resetToCurrentTime();
+    return selectedDate < today;
```
adriansberg, line 67: How does this logic work when the comparison is two strings against each other? E.g. `'2023-12-24' < '2023-12-31'`.

jonasbrunvoll: Screenshot 2023-12-18 at 08 54 43

adriansberg: Amazing, TIL

adriansberg (approved):

PR #146, opened 2023-12-14
fix: use point in "From" as focus point

adriansberg:
Use the point in "From" as focus point when searching for address, pier or stop place in "To". 

The property was removed at some point and the only thing needed was to add it again.


mikaelbr:
Did you also verify that fallback focus point is correctly set for fram?

adriansberg:
No, this PR only concerns the restoring of functionality of using the focus point in the "To" search field.

Setting the fallback, configuring it, will be in a later PR.

jonasbrunvoll (approved):

PR #147, opened 2023-12-14
refactor(filtering): include subMode, remove from departures

adriansberg:
This PR refactors how the transport mode filtering is done for a travel search, and it removes the transport mode filtering from departures since it does not work as we need it to.

The type used for the transport modes passed to the GraphQL query was refactored to include submodes. The issue (one of them) that was encountered was that if e.g. only `rail` was selected, it still showed buses. The reason for this was that for the filter option `rail`, one of the modes is `bus` with `subMode: ['railReplacementBus']`. It makes sense to show buses that are replacements for trains. So to fix this, the `transportSubModes` was included when passing to the GQL query.

> [!NOTE]
> In addition to including the sub mode, I removed transport mode filtering from departures in its entirety. I could and should have separated this removal to its own PR, but realised that when I was too deep in it, and after renaming types and refactoring, I didn't want to separate the changes. 🫣

Closes https://github.com/AtB-AS/kundevendt/issues/15753

### DIFF
--- a/src/components/line-chip/index.tsx
+++ b/src/components/line-chip/index.tsx
@@ -19,8 +19,8 @@ export default function LineChip({
   publicCode,
 }: LineChipProps) {
   const transportationColor = useTransportationThemeColor({
-    mode: transportMode,
-    subMode: transportSubmode,
+    transportMode: transportMode,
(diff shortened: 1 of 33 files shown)


mortennordseth (approved):
Nice! works great

PR #149, opened 2023-12-15
feat: connect to firebase

mortennordseth:
This PR set up the necessary configuration to connect to Firebase and Firestore. I've added a function to retrieve tariff zones as and example, which will be used in another issue where we will display them in the map. 

Service accounts for each environment are configured and deployed: https://github.com/AtB-AS/gcp-infra/pull/469

Closes https://github.com/AtB-AS/kundevendt/issues/15814


adriansberg (approved):
Looks good! :shipit:

mikaelbr:
need `measurementId` also :o

PR #150, opened 2023-12-15
refactor: use quay instead of pier

mortennordseth:
Christoffer reported that we use `quay` and not `pier` so I have changed this


adriansberg (approved):
O-quay!

PR #152, opened 2023-12-15
fix: assistant page search title

mortennordseth:
Request from Christoffer to change the assistant page search title


adriansberg (approved):
👍

PR #157, opened 2023-12-15
fix: keyboard navigation order with map

mikaelbr:
Now the order when navigating with keyboard is off. Using grid here however we can just reorder the HTML to make it better. Also added aria-hidden to map as it doesn't make sense for screen readers.


adriansberg (approved):
👍

PR #159, opened 2023-12-15
fix: add correct data to estimated calls for next departures

mortennordseth:
This PR fixes an issue where the website crashes on loading the next departures. This was because of some removed properties and newly added ones that were not mapped from GraphQL. 

The functionality is still not optimal and won't return new departures in some cases. I'll look into handling this as well, but this fixes the crashing at least.

### DIFF
--- a/src/page-modules/departures/server/journey-planner/index.ts
+++ b/src/page-modules/departures/server/journey-planner/index.ts
@@ -284,16 +284,25 @@ export function createJourneyApi(
         departures: result.data.quay?.estimatedCalls?.map((e) => ({
           id: e.serviceJourney.id,
           name: e.destinationDisplay?.frontText,
+          publicCode: e.serviceJourney.line.publicCode,
           date: e.date,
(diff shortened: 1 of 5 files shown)


src/page-modules/departures/server/journey-planner/index.ts
```diff
@@ -410,8 +419,8 @@ type RecursivePartial<T> = {
   [P in keyof T]?: T[P] extends (infer U)[]
     ? RecursivePartial<U>[]
     : T[P] extends object | undefined
-      ? RecursivePartial<T[P]>
-      : T[P];
+    ? RecursivePartial<T[P]>
```
adriansberg, line 422: My Prettier indented this, how come? 🤔

adriansberg (approved):
LGTM

src/page-modules/departures/server/journey-planner/index.ts
mortennordseth: Good question 🤔

PR #161, opened 2023-12-17
fix: update selected color of search time selector

adriansberg:
Update selected color of search time selector.

image

image

### DIFF
--- a/src/modules/search-time/selector/selector.module.css
+++ b/src/modules/search-time/selector/selector.module.css
@@ -53,7 +53,7 @@
 }
 
 .option__input:checked + .option__label {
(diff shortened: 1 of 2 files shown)


mortennordseth (approved):
Looks better! 👍 Can we apply these changes to the widget as well, without creating a new release? We can look into fixing that tomorrow of course

mikaelbr:
Yes, but the cache time would make the change slow for existing users. isn't a huge problem for now, but just makes it harder to verify. I can update the widget

adriansberg:
I have applied the changes to the widget's css.

mikaelbr:
oh yes, saw that now!

mikaelbr (approved):

PR #168, opened 2023-12-18
fix(widget): add base url to reverse and autocomplete fetch path

mortennordseth:
This adds the base URL config to the autocomplete and reverse fetch.


src/widget/widget.ts
```diff
@@ -935,9 +935,9 @@ async function autocomplete(q: string) {
   return data;
 }
 
-export async function reverse(coords: GeolocationCoordinates) {
+export async function reverse(urlBase: string, coords: GeolocationCoordinates) {
   const result = await fetch(
-    `/api/departures/reverse?lat=${coords.latitude}&lon=${coords.longitude}`,
+    `${urlBase}api/departures/reverse?lat=${coords.latitude}&lon=${coords.longitude}`,
```
mikaelbr, line 940: This now requires leading `/` in the urlBase, but I think icons does not require it. So should either: always ensure leading slash or be consistant. Think ensuring leading slash is the best here

adriansberg: Isn't that ensured here: https://github.com/AtB-AS/planner-web/blob/0e7fbcde1f51e2381684192a67a0a2d00a895996/src/widget/widget.ts#L29 ?

mortennordseth: This should be handled in the `createSettingsConstants` 🤔 

```
if (!urlBase.endsWith('/')) {
    urlBase += '/';
  }
```

mikaelbr: Dang it. Too smart for my own good

mikaelbr (approved):
Looks good!

PR #169, opened 2023-12-18
style: position home page link and page header

mortennordseth:
This moves the page header logo back to its original position and aligns the home page link. 

image


adriansberg (approved):
LGTM

PR #170, opened 2023-12-18
fix: makes widget version in sync

mikaelbr:
This is just to have version in widgets in sync with version in Releases again. Previous 2.2.1 should have also included version bump here


mortennordseth (approved):
My bad. I forgot about that when taking out the release. Noted for future releases 👍

PR #172, opened 2023-12-18
fix: set height of date/time selector to fix issue in widget

adriansberg:
The height of the label was not set explicitly for the date/time selectors. This did not cause an issue in the travel planner, but was a visible issue in the widget used on frammr.no.


mortennordseth (approved):
LGTM

PR #174, opened 2023-12-18
fix: prevent auto zoom when selecting an input field in mobile view

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/16011

### Background

When in mobile view, the browser automatically zooms in to focus on an input field whenever selected. As a result, the user must manually resize the page to get back to the original layout.   

#### Illustrations
<details>
<summary>screenrecordings/summary>

https://github.com/AtB-AS/planner-web/assets/59939294/47330b31-7ef7-4fa6-b2a7-d38639ca7807

https://github.com/AtB-AS/planner-web/assets/59939294/6adb746e-17e3-49c4-ae98-7517c372253b

</details>

Before fix / after fix

### Proposed solution

Adding a new meta tag to disable the auto zoom.

```
 <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0"/>
```
  
### Acceptance Criteria

_Acceptance criteria to test it and verify that the issue is solved._

- [ ] The browser does not zoom when focusing input field on mobile.

### DIFF
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -89,7 +89,6 @@ input[type='time'],
 input[type='url'],
 input[type='week'],
 select:focus,
-input,
 textarea {
   /* Explicitly set font size as 16px to avoid mobile zoom on focus. */
   font-size: 16px;
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -89,6 +89,7 @@ input[type='time'],
 input[type='url'],
 input[type='week'],
 select:focus,
+input,
 textarea {
   /* Explicitly set font size as 16px to avoid mobile zoom on focus. */
   font-size: 16px;
--- a/src/components/search/search.tsx
+++ b/src/components/search/search.tsx
@@ -59,6 +59,7 @@ export default function Search({
             {...getRootProps({}, { suppressRefError: true })}
           >
             <input
+              type="search"
               className={style.input}
               placeholder={placeholder}
               {...getInputProps()}
--- a/src/components/search/search.tsx
+++ b/src/components/search/search.tsx
@@ -59,7 +59,7 @@ export default function Search({
             {...getRootProps({}, { suppressRefError: true })}
           >
             <input
-              type="search"
+              type="text"
               className={style.input}
               placeholder={placeholder}
               {...getInputProps()}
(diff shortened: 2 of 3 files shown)


mikaelbr:
This is generally regarded as accessibility unfriendly as it prevents zoom overall. Meaning if someone has issues reading some information they won't be able to zoom. 

I think zooming happens usually if font sizes are less than 16px, and the usual "fix" (or hack) is to hardcode size for inputs. Hopefully this works:

```
input[type='text'],
input[type='number'],
textarea {
  font-size: 16px;
}
```

adriansberg:
> This is generally regarded as accessibility unfriendly as it prevents zoom overall. Meaning if someone has issues reading some information they won't be able to zoom.
> 
> I think auto-zooming on inputs happens usually if font sizes are less than 16px, and the usual "fix" (or hack) is to hardcode size for inputs. Hopefully this works:
> 
> ```
> input[type='text'],
> input[type='number'],
> textarea {
>   font-size: 16px;
> }
> ```

Will this in any way interfere with other stuff on mobile since the global font-size is set to 14px then? I guess it's okay to "sacrifice" the font-size of inputs to fix the zoom issue.

mikaelbr:
The 14px and 1rem is the reason for the zoom i guess. This hard codes just for i puts so wont affect anything else

jonasbrunvoll:
hmm, weird. It looks like this hack has already been added, without it preventing zooming on the input fields:

https://github.com/AtB-AS/planner-web/blob/5279f26475dec490f557daec636c1bb3936e0212/src/styles/globals.css#L77-L95

mikaelbr:
Need to check that nothing overrides it or if for some resson it doesnt work anymore. But locking zoom is not recommended at least

adriansberg:
The reason it's not working looks to be because the `input` for the search input does not have a type specified. Specifying type should make the CSS target it correctly.

src/styles/globals.css
```diff
@@ -89,6 +89,7 @@ input[type='time'],
 input[type='url'],
 input[type='week'],
 select:focus,
+input,
```
adriansberg, line 92: Not sure if there are any downsides to targeting all inputs, or if we instead should add `type="text"`, or even `type="search"` (checking all browsers to see if any of them adds any weirdness).

mikaelbr: I think its more on specificity issues tbh

jonasbrunvoll: Okey, i´ll investigate👍

adriansberg (approved):
Great 👍

PR #176, opened 2023-12-18
temp: redirect departures

mortennordseth:
Currently, the widget sends the users to `/departures?id=SomeId&searchMode=now` instead of `/departures/SomeId?searchMode=now`. To handle this I've set up a redirect until we have made a release of the widget.

### DIFF
--- a/src/pages/departures/[[...id]].tsx
+++ b/src/pages/departures/[[...id]].tsx
@@ -12,6 +12,7 @@ import { fetchFromDepartureQuery } from '@atb/page-modules/departures/fetch-depa
 import { withDepartureClient } from '@atb/page-modules/departures/server';
 import { FromDepartureQuery } from '@atb/page-modules/departures/types';
 import type { NextPage } from 'next';
+import { encode } from 'querystring';
 
 type DeparturesStopPlaceProps = {
   stopPlace: true;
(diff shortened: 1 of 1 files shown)


src/pages/departures/[[...id]].tsx
```diff
@@ -61,6 +62,16 @@ export const getServerSideProps = withGlobalData(
     DeparturesLayoutProps & DeparturesContentProps,
     { id: string[] | undefined }
   >(async function ({ client, params, query }) {
+    if (query?.id && !params?.id) {
+      const id = query.id;
+      delete query.id;
+      return {
+        redirect: {
+          destination: `/departures/${id}?${encode(query)}`,
+          permanent: true,
```
Comment (adriansberg) [inline src/pages/departures/[[...id]].tsx:71]: Why not temporary?

src/pages/departures/[[...id]].tsx
```diff
@@ -61,6 +62,16 @@ export const getServerSideProps = withGlobalData(
     DeparturesLayoutProps & DeparturesContentProps,
     { id: string[] | undefined }
   >(async function ({ client, params, query }) {
+    if (query?.id && !params?.id) {
+      const id = query.id;
+      delete query.id;
```
Comment (adriansberg) [inline src/pages/departures/[[...id]].tsx:67]: Does it matter if it's deleted or not?

adriansberg (approved):

src/pages/departures/[[...id]].tsx
Comment (mortennordseth) [reply src/pages/departures/[[...id]].tsx:71]: Should not be permanent, fixing

src/pages/departures/[[...id]].tsx
Comment (mortennordseth) [reply src/pages/departures/[[...id]].tsx:67]: I does not matter really, but I though it would be cleaner to remove it from the query

PR #178, opened 2023-12-19
fix: trip query transferSlack and walkReluctance config

mortennordseth:
Currently, the parameter to use for limiting walking is set to 1.5, which we thought was the default for the app. It turns out the default is 5 and Entur uses 4. With the current config (1.5) we have had reports that some transfers are missing, such as when traveling from Kristiansund to Molde Sjukehus. This change fixes that, so that line 702 is suggested instead of walking (see screenshots).

I've also added the transferSlack configuration, in line with what the app uses now and what the PO wanted. 

Closes https://github.com/AtB-AS/kundevendt/issues/15500

**Before**

image

**Now**

image


adriansberg (approved):

adriansberg:
Is 0 transfer slack wanted by AtB as well?

PR #179, opened 2023-12-19
feat: display both aimed start time and expected start time of departure in case of delay. 

jonasbrunvoll:
### Background

The search results currently only show the expected start time, but they should show both aimed (striked-through) and expected.

#### Illustrations
<details>
<summary>screenshot</summary>

Screenshot 2023-12-19 at 12 59 18

</details>

### Proposed solution

For all delayed departures, display both the expectedStartTime and the aimedStartTime (strike-through). Other departures that are on time only displays expectedStartTime, as for these departures expectedStartTime === aimedStartTime.  

### Acceptance Criteria
- [ ] Both the aimed start time and expected start time are shown if the start is delayed.
- [ ] Only the aimed start time (or expected, they should be the same) should show if the trip is on time.

### DIFF
--- a/src/page-modules/assistant/trip/trip-pattern/index.tsx
+++ b/src/page-modules/assistant/trip/trip-pattern/index.tsx
@@ -127,10 +127,7 @@ export default function TripPattern({
                         </Typo.span>
                       </>
                     ) : (
-                      <Typo.span
-                        textType="body__tertiary"
-                        className={style.outdatet}
-                      >
+                      <Typo.span textType="body__tertiary">
                         {formatToClock(leg.aimedStartTime, language, 'floor')}
                       </Typo.span>
                     )}
--- a/src/page-modules/assistant/trip/trip-pattern/index.tsx
+++ b/src/page-modules/assistant/trip/trip-pattern/index.tsx
@@ -143,6 +143,7 @@ export default function TripPattern({
 
           <div className={style.legs__lastLegLineContainer}>
             <div className={style.legs__lastLegLine} />
+            <div className={style.legs__lastLegLine} />
           </div>
         </div>
 
--- a/src/page-modules/assistant/trip/trip-pattern/index.tsx
+++ b/src/page-modules/assistant/trip/trip-pattern/index.tsx
@@ -20,6 +20,16 @@ type TripPatternProps = {
   index: number;
 };
 
+const isEqualOnMinuteLevel = (
+  aimedStartTime: string,
+  expectedStartTime: string,
+): boolean => {
+  return (
+    new Date(aimedStartTime).getMinutes() ===
+    new Date(expectedStartTime).getMinutes()
+  );
+};
+
 export default function TripPattern({
   tripPattern,
   delay,
@@ -108,14 +118,27 @@ export default function TripPattern({
                     )}
                   </div>
 
-                  <Typo.span textType="body__tertiary">
-                    {formatToClock(leg.expectedStartTime, language, 'floor')}
-                  </Typo.span>
+                  <div className={style.timeStartContainer}>
+                    <Typo.span textType="body__tertiary">
+                      {formatToClock(leg.expectedStartTime, language, 'floor')}
+                    </Typo.span>
+                    {i === 0 &&
+                      !isEqualOnMinuteLevel(
+                        leg.aimedStartTime,
+                        leg.expectedStartTime,
+                      ) && (
+                        <Typo.span
+                          textType="body__tertiary"
+                          className={style.lineThrough}
+                        >
+                          {formatToClock(leg.aimedStartTime, language, 'floor')}
+                        </Typo.span>
+                      )}
+                  </div>
                 </div>
 
                 {(i < expandedLegs.length - 1 || collapsedLegs.length > 0) && (
                   <div className={style.legs__legLineContainer}>
-                    <div className={style.legs__legLine} />
                     <div className={style.legs__legLine} />
                   </div>
                 )}
--- a/src/page-modules/assistant/trip/trip-pattern/index.tsx
+++ b/src/page-modules/assistant/trip/trip-pattern/index.tsx
@@ -13,6 +13,7 @@ import { TransportIconWithLabel } from '@atb/modules/transport-mode';
 import { andIf } from '@atb/utils/css';
 
 const LAST_LEG_PADDING = 20;
+const DEFAULT_THRESHOLD_AIMED_EXPECTED_IN_SECONDS = 60;
 
 type TripPatternProps = {
   tripPattern: TripPatternType;
--- a/src/page-modules/assistant/trip/trip-pattern/index.tsx
+++ b/src/page-modules/assistant/trip/trip-pattern/index.tsx
@@ -114,10 +114,7 @@ export default function TripPattern({
                     </Typo.span>
                     {secondsBetween(leg.aimedStartTime, leg.expectedStartTime) >
                       60 && (
-                      <Typo.span
-                        textType="body__tertiary"
-                        className={style.lineThrough}
-                      >
+                      <Typo.span textType="body__tertiary--strike">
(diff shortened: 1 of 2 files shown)


src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -20,6 +20,16 @@ type TripPatternProps = {
   index: number;
 };
 
+const isEqualOnMinuteLevel = (
```
mortennordseth, line 23: Maybe we could use the `secondsBetween` function that already exists in `src/utils/date.ts` ?

src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -108,15 +118,27 @@ export default function TripPattern({
                     )}
                   </div>
 
-                  <Typo.span textType="body__tertiary">
-                    {formatToClock(leg.expectedStartTime, language, 'floor')}
-                  </Typo.span>
+                  <div className={style.timeStartContainer}>
+                    <Typo.span textType="body__tertiary">
+                      {formatToClock(leg.expectedStartTime, language, 'floor')}
```
mortennordseth, line 123: Not sure if this matters, to be honest, but I immediately think that we should use `leg.aimedStartTime` here. I get that if they are within the same minute it doesn't matter and the logic would be valid. But hey, just a thought :) It might just be me and this is fine, up to you

src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -108,15 +118,27 @@ export default function TripPattern({
...
+                        {formatToClock(leg.aimedStartTime, language, 'floor')}
+                      </Typo.span>
+                    )}
+                  </div>
                 </div>
 
                 {(i < expandedLegs.length - 1 || collapsedLegs.length > 0) && (
                   <div className={style.legs__legLineContainer}>
                     <div className={style.legs__legLine} />
-                    <div className={style.legs__legLine} />
```
mortennordseth, line 119: Was this removed accidentally? There should be two lines according to the figma sketches
image

jonasbrunvoll: Yes, that div was mistakenly deleted🤔

src/page-modules/assistant/trip/trip-pattern/index.tsx
jonasbrunvoll: I´ll check it out👍

src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -108,15 +118,27 @@ export default function TripPattern({
...
+                    <Typo.span textType="body__tertiary">
+                      {formatToClock(leg.expectedStartTime, language, 'floor')}
+                    </Typo.span>
+                    {!isEqualOnMinuteLevel(
+                      leg.aimedStartTime,
+                      leg.expectedStartTime,
+                    ) && (
+                      <Typo.span
+                        textType="body__tertiary"
+                        className={style.lineThrough}
```
adriansberg, line 131: Isn't there a `textType` with strike-through?

src/page-modules/assistant/trip/trip-pattern/index.tsx
adriansberg: Should see how this affects the duration of a trip and corresponding start/end times too probably.

src/page-modules/assistant/trip/trip-pattern/trip-pattern.module.css
```diff
@@ -140,3 +140,14 @@
   gap: var(--spacings-small);
   align-items: center;
 }
+
+.timeStartContainer {
+  display: flex;
+  direction: row;
+  gap: 4px;
```
adriansberg, line 147: Should use `var(--spacings-xSmall)` instead. Should be the same size.

src/page-modules/assistant/trip/trip-pattern/index.tsx
jonasbrunvoll: I assume we want the updated departure to appear first? In that case, I dont think it will display properly if I changed to initially display aimedStartTime instead of expectedStartTime. AimedStartTime is always equal or earlier than expectedStartTime😃

src/page-modules/assistant/trip/trip-pattern/index.tsx
jonasbrunvoll: Yes, but the font-size is of that text type is larger than  body__tertiarys font-size.

src/page-modules/assistant/trip/trip-pattern/trip-pattern.module.css
jonasbrunvoll: Will fix👍

src/page-modules/assistant/trip/trip-pattern/index.tsx
adriansberg: That shouldn't be the case? 

`textType="body__tertiary"` and `textType="body__tertiary--strike"` should be the same size.

src/page-modules/assistant/trip/trip-pattern/index.tsx
mortennordseth: It would have to be something in the line of

```tsx 
{ realtime ? 
<p> { leg.expectedStartTime} </p>
<p strike-through> { leg.aimedStartTime } </p>
:
<p> { leg.aimedStartTime } </p>
}
```

As I said, might not matter or be better, but I would think that it would be more clear what times we use and why

adriansberg: Does matter more when we now use `secondsBetween()` instead of the `isEqualOnMinuteLevel()`. The former might return true for 13:39:59 and 13:40:58, but are two different minutes.

src/page-modules/assistant/trip/trip-pattern/index.tsx
jonasbrunvoll: ah, my mistake. I mixed body__primary--strike and body__tertiary--strike. You are right.

src/page-modules/assistant/trip/trip-pattern/index.tsx
mortennordseth: Good point

src/page-modules/assistant/trip/trip-pattern/index.tsx
adriansberg: Still need the color change though, right?

src/page-modules/assistant/trip/trip-pattern/index.tsx
jonasbrunvoll: True, added it back now.

src/page-modules/assistant/trip/trip-pattern/index.tsx
jonasbrunvoll: True, added it back now.

src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -108,9 +108,30 @@ export default function TripPattern({
                     )}
                   </div>
 
-                  <Typo.span textType="body__tertiary">
-                    {formatToClock(leg.expectedStartTime, language, 'floor')}
-                  </Typo.span>
+                  <div className={style.timeStartContainer}>
+                    {secondsBetween(leg.aimedStartTime, leg.expectedStartTime) >
+                    60 ? (
```
mortennordseth, line 113: Could add the `60` as a const at the top of the file 👍 

```suggestion
                    DEFAULT_THRESHOLD_AIMED_EXPECTED_IN_SECONDS ? (
```

mortennordseth (approved):
One comment about changing the 60 seconds to a const, other than that I think it looks good 👍

PR #180, opened 2023-12-19
feat: link to service journey details from a trip suggestion

mortennordseth:
This PR adds the functionality to see details about the intermediate stops of a service journey from a trip pattern. 

I've added the headers referer to the global data so that we have the correct back link in the departures details page. If a user navigates to the departures details page from a trip suggestion one should be redirected back to the trip suggestion. In other cases (from departures or opening the link from someone) one should be navigated to the departures page. 

Closes https://github.com/AtB-AS/kundevendt/issues/15872

### DIFF
--- a/src/modules/cookies/index.tsx
+++ b/src/modules/cookies/index.tsx
@@ -15,6 +15,7 @@ export type { InitialCookieData };
 export type GlobalCookiesData = {
   initialCookies: InitialCookieData;
   headersAcceptLanguage: string;
+  referer: string;
 };
 export function getGlobalCookies(req?: {
   cookies: NextApiRequestCookies;
@@ -29,5 +30,6 @@ export function getGlobalCookies(req?: {
       language: req?.cookies[LANGUAGE_COOKIE_NAME] ?? null,
     },
     headersAcceptLanguage: req?.headers['accept-language'] ?? '',
+    referer: req?.headers.referer ?? '',
   };
 }
--- a/src/page-modules/assistant/details/trip-section/estimated-calls-section.tsx
+++ b/src/page-modules/assistant/details/trip-section/estimated-calls-section.tsx
@@ -2,30 +2,52 @@ import { Typo } from '@atb/components/typography';
 import { TripRow } from '@atb/modules/trip-details';
 import { PageText, useTranslation } from '@atb/translations';
 import { secondsToDuration } from '@atb/utils/date';
+import Link from 'next/link';
 
 import style from './trip-section.module.css';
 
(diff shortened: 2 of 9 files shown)


src/modules/cookies/index.tsx
```diff
@@ -29,5 +30,6 @@ export function getGlobalCookies(req?: {
       language: req?.cookies[LANGUAGE_COOKIE_NAME] ?? null,
     },
     headersAcceptLanguage: req?.headers['accept-language'] ?? '',
+    referer: req?.headers.referer ?? '',
```
adriansberg, line 33: Do we need to set it as a cookie?

mortennordseth: I might be reading the code wrong here, but this shouldn't set it as a cookie. A bit weird naming of the function maybe. Should probably be "getInitialData" or something

src/page-modules/assistant/details/trip-section/estimated-calls-section.tsx
```diff
@@ -2,30 +2,52 @@ import { Typo } from '@atb/components/typography';
...
+  serviceJourneyId,
+  date,
+  fromQuayId,
 }: EstimatedCallsSectionProps) {
   const { t, language } = useTranslation();
 
+  const shouldShowLinkToServiceJourney = serviceJourneyId ? true : false;
+  const serviceJourneyHref = fromQuayId
+    ? `/departures/details/${serviceJourneyId}?date=${date}&fromQuayId=${fromQuayId}`
+    : `/departures/details/${serviceJourneyId}?date=${date}`;
```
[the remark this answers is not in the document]
mortennordseth: True, I'll fix 👍

src/modules/cookies/index.tsx
adriansberg: Nevermind me, they aren't set as a cookie, only read from cookie of course. Confused by the type and naming of function as well.

src/page-modules/departures/details/index.tsx
```diff
@@ -49,13 +51,22 @@ export function DeparturesDetails({
     .map((s) => s.situationNumber)
     .filter((s): s is string => !!s);
 
+  const backLink =
+    referer && referer.includes('assistant/')
+      ? referer
+      : `/departures/${focusedCall.quay.stopPlace.id}`;
```
adriansberg, line 57:
```suggestion
  const backLink = referer?.includes('assistant/')
    ? referer
    : `/departures/${focusedCall.quay.stopPlace.id}`;
```

adriansberg (approved):

PR #185, opened 2023-12-20
fix: ensure that live icon and description is not rendered when displaying canceled departures.

jonasbrunvoll:
Fixes https://github.com/AtB-AS/kundevendt/issues/15983

### Background:

Currently, both the live icon and the description "I rute" is displayed when routes are cancelled in realtime. When cancelled, neither th icon or the description should be displayed. 
 
#### Illustrations
<details>
<summary>screenshot</summary>

Screenshot 2023-12-20 at 12 28 43

</details>

### Proposed solution:

Added a condition to verify that the departure is not cancelled. If cancelled, the icon and description are not included.  

### Acceptance criteria 
- [ ] The live icon and description should never display for cancelled departures.

### DIFF
--- a/src/page-modules/departures/details/index.tsx
+++ b/src/page-modules/departures/details/index.tsx
@@ -49,6 +49,10 @@ export function DeparturesDetails({
     .map((s) => s.situationNumber)
     .filter((s): s is string => !!s);
 
+  const isJourneyCancelled = situations.some((item) =>
+    item.summary.some((i) => i.value === 'Journey cancelled'),
+  );
+
   return (
     <section className={style.container}>
       <div className={style.headerContainer}>
@@ -66,7 +70,7 @@ export function DeparturesDetails({
           />
           <Typo.h2 textType="heading--big">{title}</Typo.h2>
         </div>
-        {realtimeText && (
+        {realtimeText && !isJourneyCancelled && (
           <div className={style.realtimeText}>
             <ColorIcon icon="status/Realtime" size="small" />
             <Typo.h3 textType="body__secondary">{realtimeText}</Typo.h3>
--- a/src/page-modules/departures/details/index.tsx
+++ b/src/page-modules/departures/details/index.tsx
@@ -35,6 +35,7 @@ export function DeparturesDetails({
       realtime: c.realtime,
     })),
   );
+
   const estimatedCallsWithMetadata = addMetadataToEstimatedCalls(
     serviceJourney.estimatedCalls,
     fromQuayId,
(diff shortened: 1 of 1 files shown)


adriansberg:
I guess this text is from Entur?

2023-12-20 at 13 13 16@2x

jonasbrunvoll:
I belive so. I could not find any text in the repo that said 'Journey cancelled'. If you have a suggestion to another way to detect if a departure is cancelled, I would love to hear about it.

src/page-modules/departures/details/index.tsx
```diff
@@ -49,6 +49,10 @@ export function DeparturesDetails({
     .map((s) => s.situationNumber)
     .filter((s): s is string => !!s);
 
+  const isJourneyCancelled = situations.some((item) =>
+    item.summary.some((i) => i.value === 'Journey cancelled'),
```
adriansberg, line 53: Is this the only way to know if an entire journey is cancelled? Just feels to me a bit fragile checking a string instead of a property like the estimated calls have. 

And if so, shouldn't the `situations` for the service journey be checked instead of the estimated call? (might not matter which one is checked since they both will have "Journey cancelled" possibly?)

jonasbrunvoll: I agree that the solution is a bit fragile. I can search a bit further but I haven't found a better solution yet.

jonasbrunvoll: I am viewing a departure that was cancelled earlier today at 06.45. The prop focusedCall.cancellation : false. Should´t this prop be true, since the departure was cancelled?    
  
Screenshot 2023-12-20 at 14 19 11

mortennordseth: I would assume so 🤔 Actual departure time is null as well, which should be updated with the actual departure time after it has passed a stop, so I guess it is true that this was cancelled. Not sure why the cancellation is false though

jonasbrunvoll: Interesting. I also checked a departure that was carried out as usual and both actualArrivalTime and actualDepartureTime are set to null here also.

Screenshot 2023-12-20 at 14 43 43

jonasbrunvoll: Tried some routes that have been canceled recently. Here cancellation is set to true. Is there a difference between two-digit routes and three-digit routes when it comes to reporting? 

Screenshot 2023-12-20 at 16 28 12
Screenshot 2023-12-20 at 16 28 26

jonasbrunvoll: Anyway, I´ll update the condition to check for the cancellation prop. Exploring whether cancellation in some cases is not set when it should is a separate task that in my opinion should be solved in a separate branch.

mortennordseth (approved):

PR #186, opened 2023-12-20
feat: reset search when switching tab

adriansberg:
## Summary

This PR introduces a couple of features and bug fixes:

-  **Features:**
    - Added functionality to reset search when switching between tabs.
    - Make it possible to click the search button at all times, routing to the travel planner even though from/to might not be set. Some users experienced trouble showing the autocomplete box, and wasn't able to click the search button because they couldn't set a location.

-  **Fixes:**
    - Fixed proper use of ARIA property.
    - Added missing translation for "From" label.

# ~~*Pull Request Description:*~~ 

### ~~*Overview*~~
~~*This PR addresses the issue with the widget failing due to being nested within a `form` element in the FRAM page. The problem arises from the HTML specification restriction of not allowing a form inside another form.*~~

### ~~*Changes*~~
~~*1. The widget has been rewritten to remove the `form` dependency. Event handlers have now been added to the input fields and the search button.*~~

### ~~*Impact*~~
~~*Here's how the modified widget fulfils the acceptance criteria:*~~

-    ~~*The reformatted widget performs in the same manner as its previous implementation. This ensures seamless user experience, with no change detected on the frontend.*~~

-    ~~*The widget can now function efficiently inside a `form` block without necessitating any excessive workarounds.*~~

-    ~~*Clicking on the 'search' button triggers a search action, signifying successful event binding.*~~

-    ~~*Pressing 'Enter' while focusing on an input field initiates a search, indicating successful Key Event handling.*~~

-    ~~*If the 'From' field is left blank, the search operation is not executed, maintaining the original search logic.*~~

-    ~~*The proper state of date/time selectors is sustained while toggling between different tabs.*~~

~~*Revising the widget to leverage event handlers for inputs and buttons, rather than relying on a `form`, has rectified the nesting issues without compromising on functionality or user experience.*~~

~~*Please review the changes and give your feedback.*~~

### DIFF
--- a/src/widget/widget.module.css
+++ b/src/widget/widget.module.css
@@ -223,10 +223,12 @@
 .button.button--disabled {
   background-color: var(--interactive-interactive_0-disabled-background);
   color: var(--interactive-interactive_0-disabled-text);
+  cursor: default;
 }
 .button:focus {
   outline: 0;
-  box-shadow: inset 0 0 0 var(--border-width-medium) var(--interactive-interactive_0-outline-background);
+  box-shadow: inset 0 0 0 var(--border-width-medium)
+    var(--interactive-interactive_0-outline-background);
 }
 
 /* Autocomplete */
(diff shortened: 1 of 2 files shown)


mikaelbr:
Should have said this before, but I actually think this shouldn't be implemented. It was more like a short time fix for FRAM specific stuff. But removing form here will reduce a11y and generally make it less semantic.

mikaelbr (approved):

PR #187, opened 2023-12-21
feat: intermediate stops link from trip details

mortennordseth:
This PR adds the functionality to see details about the intermediate stops of a service journey from a trip pattern.

I've added the headers referer to the global data so that we have the correct back link in the departures details page. If a user navigates to the departures details page from a trip suggestion one should be redirected back to the trip suggestion. In other cases (from departures or opening the link from someone) one should be navigated to the departures page.

I also extended the compressed query string so that the user can navigate back to the trip search results with the correct search parameters.

Closes https://github.com/AtB-AS/kundevendt/issues/15872

### DIFF
--- a/src/page-modules/assistant/details/trip-section/estimated-calls-section.tsx
+++ b/src/page-modules/assistant/details/trip-section/estimated-calls-section.tsx
@@ -23,7 +23,7 @@ export function EstimatedCallsSection({
 }: EstimatedCallsSectionProps) {
   const { t, language } = useTranslation();
 
-  const shouldShowLinkToServiceJourney = serviceJourneyId ? true : false;
+  const shouldShowLinkToServiceJourney = !!serviceJourneyId;
   let serviceJourneyHref = `/departures/details/${serviceJourneyId}?date=${date}`;
   if (fromQuayId) {
     serviceJourneyHref += `&fromQuayId=${fromQuayId}`;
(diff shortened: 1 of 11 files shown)


src/page-modules/assistant/details/trip-section/estimated-calls-section.tsx
```diff
@@ -2,30 +2,53 @@ import { Typo } from '@atb/components/typography';
...
 export function EstimatedCallsSection({
   numberOfIntermediateEstimatedCalls,
   duration,
+  serviceJourneyId,
+  date,
+  fromQuayId,
 }: EstimatedCallsSectionProps) {
   const { t, language } = useTranslation();
 
+  const shouldShowLinkToServiceJourney = serviceJourneyId ? true : false;
```
adriansberg, line 26: This can be simplified, but maybe it's not cognitively simpler.

```suggestion
  const shouldShowLinkToServiceJourney = !!serviceJourneyId;
```

mortennordseth: I think this is better 👍

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -227,7 +227,10 @@ export function createJourneyApi(
         TripsWithDetailsQueryVariables
       >({
         query: TripsWithDetailsDocument,
-        variables: tripQuery.query,
+        variables: {
+          ...tripQuery.query,
+          arriveBy: false,
```
adriansberg, line 232: Never makes sense that this can be true?

mortennordseth: No, because we are looking up a specific trip for a specific departure time. This means we want to find a trip that departs by time xx, if that makes sense?

mortennordseth: But to have the correct search mode when navigating back to the trip search results from trip details we have to know if the search was done as `arriveBy` or `departBy` and that's why I modified the compressed query

adriansberg: 👍

adriansberg (approved):
Works as expected 👏

PR #188, opened 2023-12-21
feat: get default focus point from Firebase

adriansberg:
This PR adds functionality that fetches the default focus point from Firebase and uses it for autocomplete, if it exists. If it does not exist in Firebase, a fallback is used. This fallback is currently hardcoded, not sure if we should handle it a different way. If we do, it means we have several ways of configuring the default focus point, which I mean is unnecessary. I think it's fine to have a hardcoded fallback, to e.g. Trondheim as it is now, and if it's needed/wanted to have it somewhere else Firebase should be used.

Fixes https://github.com/AtB-AS/kundevendt/issues/15997

### DIFF
--- a/src/modules/firebase/default-focus-point.ts
+++ b/src/modules/firebase/default-focus-point.ts
@@ -1,0 +1,26 @@
+import { getFirestore, getDoc, doc } from 'firebase/firestore';
+import app from './firebase';
+import { DefaultFocusPoint, defaultFocusPointSchema } from './types';
+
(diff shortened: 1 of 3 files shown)


adriansberg:
Do we need to add something here as well maybe? https://github.com/AtB-AS/firestore-configuration/blob/main/schema-definitions/other.json

mortennordseth (approved):
Looks very good 👍 I think we should update the firestore configuration schema as you mentioned

PR #191, opened 2023-12-22
fix: includes message from message box into buttons aria-label, to make message accessible for voiceover.

jonasbrunvoll:
Fixes https://github.com/AtB-AS/kundevendt/issues/15810

### Background

When the voiceover reads message boxes, only the button text is read. It is desirable that both the message and the button text is read by the voiceover.  

#### Illustrations
<details>
<summary>screenrecording</summary>

https://github.com/AtB-AS/planner-web/assets/59939294/7be3d20b-4f43-4845-81ac-a05bae115605

</details>

### Proposed solution

The proposed solution passes the message as prop to the button component, where it is included in the buttons arial-label, which is read by the voiceover. 

### Acceptance Criteria

- [ ] Both the massage and the button text is read by the voiceover.

### DIFF
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -26,7 +26,6 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
         disabled={props.disabled || props.state === 'loading'}
         aria-disabled={props.disabled || props.state === 'loading'}
         data-testid={testID}
-        aria-label={buttonProps?.['aria-label']}
         {...buttonProps}
       >
         <ButtonBase {...props} />
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -12,10 +12,14 @@ export type ButtonProps = {
    * Pass properties to button element directly
    */
   buttonProps?: JSX.IntrinsicElements['button'];
+  /**
+   * Define aria-label for voiceover to read.
+   */
+  ariaLabel?: string;
 } & ButtonBaseProps;
 
 export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
-  function Button({ onClick, testID, buttonProps, ...props }, ref) {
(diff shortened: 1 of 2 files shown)


src/components/button/button.tsx
```diff
@@ -12,10 +12,14 @@ export type ButtonProps = {
    * Pass properties to button element directly
    */
   buttonProps?: JSX.IntrinsicElements['button'];
+  /**
+   * Define aria-label for voiceover to read.
+   */
+  ariaLabel?: string;
```
mortennordseth, line 18: This already exists in the `buttonProps` I learned a week ago 😄

jonasbrunvoll: Aah! I was not aware of that either. Good to know!

src/components/button/button.tsx
```diff
@@ -26,6 +26,7 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
         disabled={props.disabled || props.state === 'loading'}
         aria-disabled={props.disabled || props.state === 'loading'}
         data-testid={testID}
+        aria-label={buttonProps?.['aria-label']}
         {...buttonProps}
```
mortennordseth, line 30: It's unnecessary to type `aria-label` explicitly, as it is passed with `{...buttonProps}`

src/components/message-box/index.tsx
```diff
@@ -10,6 +10,7 @@ import { Typo } from '@atb/components/typography';
 
 import style from './message-box.module.css';
 import { colorToOverrideMode } from '@atb/utils/color';
+import { screenReaderPause } from '../typography/utils';
```
mortennordseth, line 13:
```suggestion
import { screenReaderPause } from '@atb/components/typography/utils';
```

mortennordseth (approved):
Looks good!

PR #192, opened 2023-12-24
fix: show whole trip on map load

mortennordseth:
This PR makes the whole trip visible on the map on load instead of the first stop place. 

image

Closes https://github.com/AtB-AS/kundevendt/issues/16058

### DIFF
--- a/src/components/map/map.tsx
+++ b/src/components/map/map.tsx
@@ -21,6 +21,7 @@ export type MapProps = {
   onSelectStopPlace?: (id: string) => void;
   mapLegs?: MapLegType[];
   initialZoom?: number;
+  bounds?: [[number, number], [number, number]];
 };
 
 export function Map({
@@ -29,6 +30,7 @@ export function Map({
   onSelectStopPlace,
   mapLegs,
   initialZoom = ZOOM_LEVEL,
+  bounds,
 }: MapProps) {
   const mapWrapper = useRef<HTMLDivElement>(null);
   const mapContainer = useRef<HTMLDivElement>(null);
@@ -47,6 +49,7 @@ export function Map({
       style: mapboxData.style,
       center: position,
       zoom: initialZoom,
+      bounds: bounds, // If bounds is specified, it overrides center and zoom constructor options.
     });
 
     return () => map.current?.remove();
@@ -59,16 +62,6 @@ export function Map({
   );
   useMapPin(map, position, layer);
   useMapLegs(map, mapLegs);
-
-  useEffect(() => {
-    if (map.current) {
-      map.current.flyTo({
-        center: position,
-        zoom: initialZoom,
-        speed: 2,
-      });
-    }
-  }, [position, initialZoom]);
 
   return (
     <div className={style.map} aria-hidden="true">
--- a/src/components/map/map.tsx
+++ b/src/components/map/map.tsx
@@ -49,7 +49,7 @@ export function Map({
       style: mapboxData.style,
       center: position,
       zoom: initialZoom,
-      bounds: bounds ? [bounds.sw, bounds.ne] : undefined, // If bounds is specified, it overrides center and zoom constructor options.
+      bounds, // If bounds is specified, it overrides center and zoom constructor options.
     });
 
     return () => map.current?.remove();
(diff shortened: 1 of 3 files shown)


src/components/map/map.tsx
```diff
@@ -60,16 +63,6 @@ export function Map({
...
-  useEffect(() => {
-    if (map.current) {
-      map.current.flyTo({
-        center: position,
-        zoom: initialZoom,
-        speed: 2,
-      });
-    }
-  }, [position, initialZoom]);
-
```
mortennordseth, line 72: This is not needed as the position or initial zoom is not changing unless we show a new route / stop and the map is rendered.

jonasbrunvoll:
No images/videos are provided, but I assume the task is solved. The code looks good😊

jonasbrunvoll (approved):

mortennordseth:
> No images/videos are provided, but I assume the task is solved. The code looks good😊

I've updated the description with a screenshot 👍 Also, each PR is deployed to Vercel, making the changes available to test if you want to do so. Check out the "preview" link in the screenshot below 😃 

image

src/page-modules/assistant/details/index.tsx
```diff
@@ -47,14 +48,7 @@ export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
...
       <div className={style.mapContainer}>
-        <Map
-          position={{
-            lon: tripPattern.legs[0].fromPlace.longitude,
-            lat: tripPattern.legs[0].fromPlace.latitude,
-          }}
-          initialZoom={13.5}
-          mapLegs={mapLegs}
-        />
+        <Map mapLegs={mapLegs} bounds={[bounds.sw, bounds.ne]} />
```
mikaelbr, line 51: Could this be baked in to the component instead where as it always show the entire trip? Or atleast something like `showAllLegsInView` as a property and move invocation of `getMapBounds(mapLegs)` encapsulated by `Map` component?

src/components/map/map.tsx
```diff
@@ -21,6 +21,7 @@ export type MapProps = {
   onSelectStopPlace?: (id: string) => void;
   mapLegs?: MapLegType[];
   initialZoom?: number;
+  bounds?: [[number, number], [number, number]];
 };
```
adriansberg, line 25: All of these props are optional now, how can I know which ones cancel others and what not? I also support Mikael's thought in the comment regarding moving bounds.

src/page-modules/assistant/details/index.tsx
mortennordseth: Updated to reflect this now. When `mapLegs` is defined we want to show the whole trip, which makes your suggestion make much sense 👍

src/components/map/map.tsx
```diff
@@ -47,6 +49,7 @@ export function Map({
       style: mapboxData.style,
       center: position,
       zoom: initialZoom,
+      bounds: bounds ? [bounds.sw, bounds.ne] : undefined, // If bounds is specified, it overrides center and zoom constructor options.
```
adriansberg, line 52: Maybe `getMapBounds` can return the tuple we want here?

adriansberg: Then it would be just:
```
  center: position,
  zoom: initialZoom,
  bounds
```

src/components/map/map.tsx
mortennordseth: We could do something like this to indicate that `position` and `initialZoom` are redundant when `mapLegs` is defined, and check for what props are given, but I'm unsure if this is easier to understand. We also need to include that none is required somehow. Do you think moving the comment from line 52 is enough? 

```ts

export type MapProps = {
  layer?: string;
  onSelectStopPlace?: (id: string) => void;
} & (
  | {
      position?: Position;
      initialZoom?: number;
    }
  | {
      mapLegs?: MapLegType[];
    }
);

```

mikaelbr: Is this more accurate ?

```
export type MapProps = {
  layer?: string;
  onSelectStopPlace?: (id: string) => void;
} & (
  | {
      position: Position;
      initialZoom: number;
    }
  | {
      mapLegs: MapLegType[];
    }
  | {}
);
```

mortennordseth: Yes, that is more accurate 👍 I'll update

adriansberg (approved):
🚀

PR #193, opened 2024-01-02
feat: multimodal stops search results

mortennordseth:
This makes it so that the search returns children of a multimodal stop instead of the multimodal stop itself. This is done because returning only the multimodal stop would list "Dryna ferjekai" (and similar stops) with a bus icon, which also is a ferry stop.

In the future, we could consider improving this to show only one result for a multimodal stop but include multiple icons.

image

Closes https://github.com/AtB-AS/kundevendt/issues/16156


jonasbrunvoll (approved):
Looks good😊

PR #196, opened 2024-01-04
fix: fix issue with widget not selecting place

adriansberg:
The issue with the widget not selecting a place was due to the event listener on blur. The blur event was happening before the "combobox-commit" event which is the one that sets the place. A fix for this was to add event listener to the `mousedown` event instead of the `click` event in the combobox code. That was solved with a patch using `patch-package`. Also mentioned in this issue: https://github.com/github/combobox-nav/issues/54

Also fixed an issue that showed "undefined" if selected place did not have a locality.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
     "generate": "graphql-codegen",
     "prestart": "yarn generate",
     "predev": "yarn generate",
-    "generate-widget-version": "sh ./scripts/build-widget.sh"
+    "generate-widget-version": "sh ./scripts/build-widget.sh",
+    "postinstall": "patch-package"
   },
   "dependencies": {
     "@apollo/client": "^3.8.8",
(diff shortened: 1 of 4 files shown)


mortennordseth:
From what I can see, this changes the current version of the widget (v2.6.0), but should probably be a new version? 🤔

adriansberg:
Yeah, maybe it should? Thought that since it was a fix and not any breaking changes that it could be the same version so that it is updated, but I might've misunderstood how we treat the widget version.

adriansberg:
I can update to 2.6.1

mortennordseth:
> I can update to 2.6.1

It should be v2.8.0 I think, because this would be released with that version.

> Version is based on package.json. We should always keep this in sync with the generated version in Releases so the changelog is correct.

adriansberg:
Ah, mega confused.

What decides that the next release is 2.8.0 and not 2.7.1? I'd guess the trip in map changes.

---

So. If I change the version in package.json to 2.8.0, run the script again then make a draft release that matches the version in package.json?

mortennordseth:
Yes, since we are releasing with a new feature (add functionality) it will be 2.8.0

That sounds correct to me 👍 And unstage the v2.6.0 changes

adriansberg:
As long as we know that we'll release added changes and not a bug fix before the changes that is.

mortennordseth:
Yep, that's something to consider. Maybe we shouldn't generate new widget versions for PRs like this and do it as a part of the release flow instead.  

1. Draft a new release
2. Any changes to the widget? => generate a new widget version and push to main
3. Release

PR #198, opened 2024-01-08
fix(widget): fix search selection with keyboard

adriansberg:
Patching `@github/combobox-nav` resulted in keyboard selection not working. The function called when pressing Enter or Tab fired the `click` event, which we changed to be a `mousedown` event. Used `fireCommitEvent` instead.


patches/@github+combobox-nav+2.3.0.patch
```diff
@@ -20,3 +20,12 @@ index d86de71..3956688 100644
      }
      indicateDefaultOption() {
          var _a;
+@@ -155,7 +155,7 @@ function commit(input, list) {
+         return false;
+     if (target.getAttribute('aria-disabled') === 'true')
+         return true;
+-    target.click();
++    fireCommitEvent(target)
```
adriansberg, line 28: @mikaelbr, the `fireCommitEvent` also has `event` as a parameter, which is passed to `CustomEvent` like this:

```js
fireCommitEvent(target, { event })
---
function fireCommitEvent(target, detail) {
    target.dispatchEvent(new CustomEvent('combobox-commit', { bubbles: true, detail }));
}
```

Is it needed in this case as well? Seems to work as expected without it, but might be some information in it that should be passed.

mikaelbr: Might be needed for submitting things. It depends on what event is subscribed to and what not 🤔 Wouldn't really know without testing it out. But if it works it works? 🤷‍♂️

mikaelbr: Detail in custom event here could be used to access event details, but if we don't use it and retrieve that information i think it's not a problem to omit it

mikaelbr (approved):
Doing commit directly looks to be OK, and if it works with testing I think it is fine 👍

patches/@github+combobox-nav+2.3.0.patch
adriansberg: I couldn't see that we used it anywhere, and the combobox itself did not use it for anything. It works.

PR #199, opened 2024-01-08
Improve header

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/16123


src/layouts/shared/page-header.tsx
```diff
@@ -16,13 +20,20 @@ export default function PageHeader() {
...
-                role="none"
-                size="normal"
-                overrideMode="dark"
-              />
+              {fylkeskommune && (
+                <Image
+                  width={0}
+                  height={0}
+                  sizes="100vw"
+                  style={{ width: '100%', height: 'auto' }}
```
toretefre, line 28: This is insane. But apparently best practice: https://github.com/vercel/next.js/discussions/18474#discussioncomment-5501724

src/layouts/shared/page-header.tsx
```diff
@@ -1,10 +1,14 @@
...
 import Link from 'next/link';
 import style from './page-header.module.css';
-import { MonoIcon } from '@atb/components/icon';
+import { useDarkMode } from '@atb/modules/theme';
+import Image from 'next/image';
+import { getOrgData } from '@atb/modules/org-data';
 
 export default function PageHeader() {
   const { t } = useTranslation();
+  const [isDarkMode] = useDarkMode();
```
toretefre, line 10: Not sure why `isDarkMode` is wrapped in an array, but it seems to be the pattern used elsewhere.

src/layouts/shared/page-header.tsx
```diff
@@ -16,13 +20,20 @@ export default function PageHeader() {
...
+                  width={0}
+                  height={0}
+                  sizes="100vw"
+                  style={{ width: '100%', height: 'auto' }}
+                  src={
+                    isDarkMode
+                      ? fylkeskommune.logoSrcDark
+                      : fylkeskommune.logoSrc
+                  }
+                  alt={fylkeskommune.name}
```
toretefre, line 34: Should we append a "travel planner" at the end of the alt text here?

mortennordseth:
I'm not sure if AtB or NFK would like to have this logo in their header 🤔 Does it make sense to make it configurable? I think we could add a check that says `if !title use logo, else use title`. If they do want the change, we need to handle the logo color in light-mode. 

image

Also, I think we should keep the 3 columns to have the same look in the webshop and planner web. 

image

So the footer would look something like:

image

adriansberg:
> I'm not sure if AtB or NFK would like to have this logo in their header 🤔 Does it make sense to make it configurable? I think we could add a check that says if !title use logo, else use title. If they do want the change, we need to handle the logo color in light-mode.

Wouldn't this be solved by changing the `logo/Logo` in the design system assets to the full logo and allowing the width to be variable? As it is now the mono icon is always square, not sure if everything breaks if allowing the width to be variable (keeping the aspect ratio).

toretefre:
> I'm not sure if AtB or NFK would like to have this logo in their header 🤔 Does it make sense to make it configurable? I think we could add a check that says `if !title use logo, else use title`. If they do want the change, we need to handle the logo color in light-mode.

Agree. Added a check now. If it has a length, siteTitle is preferred. So fylkeslogo will only be shown if siteTitle for the present language is falsy (`''`).

toretefre:
Re-read the original text from FRAM now, and agree to revert to three-column footer. Should be like this then? (With possible help url appearing in the left column)

image

mortennordseth:
> Re-read the original text from FRAM now, and agree to revert to three-column footer. Should be like this then? (With possible help url appearing in the left column)

Yes, I think this looks good :)

src/layouts/shared/page-header.tsx
```diff
@@ -16,13 +20,22 @@ export default function PageHeader() {
...
+                      style={{ width: '100%', height: 'auto' }}
+                      src={
+                        isDarkMode
+                          ? fylkeskommune.logoSrcDark
+                          : fylkeskommune.logoSrc
+                      }
+                      alt={fylkeskommune.name}
+                    />
+                  )}
               <span>{t(CommonText.Titles.siteTitle)}</span>
```
mortennordseth, line 39: This makes the `siteTitle` appear twice if the `siteTitle` is defined

toretefre:
Fixed duplicate title now, and added the monoIcon back when it is needed. 

Also noticed that the siteTitle of course is used as the ... site title, so I will need to find a way around.

src/translations/common/titles.ts
```diff
@@ -10,6 +10,10 @@ export const Titles = orgSpecificTranslations(TitlesInternal, {
     siteTitle: _('Reis Reisesøk', 'Reis Travel Search', 'Reis Reisesøk'),
   },
   fram: {
-    siteTitle: _('FRAM Reisesøk', 'FRAM Travel Search', 'FRAM Reisesøk'),
```
toretefre, line 13: Are we preferring Reisesøk or Reiseplanlegger here?

mortennordseth: > Når vi skal bytte fra FRAM Reisesøk og FRAM Nettbutikk til bare FRAM-logoen i navbaren på nettsidene (trenger ikke gjøre logoen større enn plassen som blir brukt nå), kan vi fjerne at det står FRAM Reiseplanlegger og FRAM Nettbutikk i footeren også.
I i footeren kan det stå bare Reiseplanlegger og Nettbutikk i første spalten (hvor det står FRAM Reisesøk og FRAM Nettbutikk nå), og i andre spalten redigere Hjelp til reiseplanleggeren til Kontakt oss (skrive Kontakt oss i nettbutikken også)

I think PO would prefer Reiseplanlegger ☝️

toretefre: Fine as it is now then 👌 You can do the final review if you have time 🤝

mortennordseth (approved):

adriansberg:
Should update the PR title to include type (Conventional Commits).

PR #200, opened 2024-01-10
refactor: fetch trip patterns from the client

mortennordseth:
This changes the trip pattern fetching from being done on the server to the client. This is done to decrease the initial load time of the page. 

Closes https://github.com/AtB-AS/kundevendt/issues/16300

### DIFF
--- a/src/page-modules/assistant/client/journey-planner/index.ts
+++ b/src/page-modules/assistant/client/journey-planner/index.ts
@@ -1,9 +1,8 @@
-import { TripData } from '../../server/journey-planner/validators';
-import { TripQuery } from '../../types';
+import { NonTransitTripData, TripQuery, TripData } from '../../types';
 
 export type TripApiReturnType = TripData;
 
-export async function nextTripPatterns(
+export async function getTripPatterns(
   query: TripQuery,
 ): Promise<TripApiReturnType> {
   const queryString = tripQueryToQueryString(query);
@@ -22,4 +21,13 @@ function tripQueryToQueryString(input: TripQuery): string {
         encodeURIComponent(String(input[key as keyof TripQuery])),
     )
     .join('&');
+}
+
+export type NonTransitTripApiReturnType = NonTransitTripData;
+export async function getNonTransitTrip(
+  query: TripQuery,
+): Promise<NonTransitTripApiReturnType> {
+  const queryString = tripQueryToQueryString(query);
+  const result = await fetch(`/api/assistant/non-transit-trip?${queryString}`);
+  return await result.json();
 }
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -203,16 +203,15 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
             title={t(PageText.Assistant.search.buttons.find.title)}
             className={style.button}
             mode="interactive_0--bordered"
-            disabled={!tripQuery.from || !tripQuery.to}
+            disabled={!tripQuery.from || !tripQuery.to || isSearching}
             buttonProps={{ type: 'submit' }}
-            state={isSearching ? 'loading' : undefined}
           />
         </div>
       </form>
 
       <section className={style.contentContainer}>
         <EmptySearch
-          isSearching={isSearching}
+          isSearching={false}
           isGeolocationError={geolocationError !== null}
           type="trip"
         >
--- a/src/page-modules/assistant/trip/index.tsx
+++ b/src/page-modules/assistant/trip/index.tsx
@@ -3,16 +3,16 @@ import {
   formatToWeekday,
   parseIfNeeded,
 } from '@atb/utils/date';
-import {
-  TripData,
-  type TripPattern as TripPatternType,
-} from '../server/journey-planner/validators';
+import { type TripPattern as TripPatternType } from '../server/journey-planner/validators';
 import style from './trip.module.css';
 import { PageText, useTranslation } from '@atb/translations';
 import { Typo } from '@atb/components/typography';
-import { GeocoderFeature } from '@atb/page-modules/departures';
-import { nextTripPatterns } from '../client';
-import { FromToTripQuery, NonTransitTripData } from '../types';
+import { getNonTransitTrip, getTripPatterns } from '../client';
+import {
+  FromToTripQuery,
+  NonTransitTripData,
+  TripPatternWithTransitionDelay,
+} from '../types';
 import {
   createTripQuery,
   filterOutDuplicates,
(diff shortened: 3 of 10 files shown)


src/page-modules/assistant/trip/index.tsx
```diff
@@ -96,68 +105,87 @@ export default function Trip({ tripQuery, trip, nonTransitTrips }: TripProps) {
...
+      const trip = await getTripPatterns(tripQuery);
+      const newTripPatternsWithTransitionDelay =
+        tripPatternsWithTransitionDelay(trip.tripPatterns);
+
+      setTripPatterns(newTripPatternsWithTransitionDelay);
+      setCursor(getCursorBySearchMode(trip, query.searchTime.mode));
+      setIsFetchingTripPatterns(false);
+    };
+    fetch();
+  }, [query]);
```
mikaelbr, line 145: Maybe I should have thought of this earlier, but thinking about this. Maybe useSWR would be a better approach here than a clean fetch call? swr has Fallback-components which can support SSR. So if Fallback component has data from SSR it will be populated from server, if not it will do client cache. This way we can inject cached results from server when navigating back if that makes sense.

src/page-modules/assistant/client/journey-planner/index.ts
```diff
@@ -1,18 +1,55 @@
...
 
 export type TripApiReturnType = TripData;
 
-export async function nextTripPatterns(
-  query: TripQuery,
-): Promise<TripApiReturnType> {
-  const queryString = tripQueryToQueryString(query);
+const getKey =
+  (query: TripQuery) =>
+  (pageIndex: number, previousPageData: TripApiReturnType) => {
```
mikaelbr, line 10: Generally unsure if this is the most readable way to do "double functions". What you are doing here is essentially allowing for [partial application](https://en.wikipedia.org/wiki/Partial_application). But even without doing structured partial application i think maybe this would be easier to read without having two arrow functions? 🤔

```ts
function createKeyGetter(query: TripQuery) {
  return (pageIndex: number, previousPageData: TripApiReturnType) => {
```

src/page-modules/assistant/trip/index.tsx
```diff
@@ -3,48 +3,46 @@ import {
...
-  const { tripPatterns, fetchNextTripPatterns, isFetchingTripPatterns } =
-    useTripPatterns(trip, tripQuery.searchTime);
+  const { trips, isLoading, loadMore, isLoadingMore } = useTripPatterns(query);
+  const { nonTransitTrips } = useNonTransitTrip(query);
 
-  const nonTransits = Object.entries(nonTransitTrips);
+  useEffect(() => {
+    const q = createTripQuery(tripQuery, getInitialTransportModeFilter(null));
+    setQuery(q);
+  }, [tripQuery]);
```
mikaelbr, line 37: Is the intention here to avoid query changing as object causing an infinite fetch loop? If so I think maybe `useMemo` is a better option? 🤔

mortennordseth: This might be a remnant from several rounds of refactoring when transitioning from fetch to useSWR. That was the intention, and `useMemo` should be a better option to solve that. However, it might not be necessary to parse the `FromToTripQuery` object here at all 🤔 I guess the hook can handle that as well. This also makes the transport mode filter not work, so I'll fix that 👍

src/page-modules/assistant/client/journey-planner/index.ts
mortennordseth: I didn't think of that. I agree, more readable for sure

src/page-modules/assistant/client/journey-planner/index.ts
```diff
@@ -1,16 +1,59 @@
...
+    isLoading: isLoadingInitialData,
+    isError: error,
+    loadMore: () => setSize(size + 1),
+    isLoadingMore,
+  };
+}
+
+export function useNonTransitTrip(query: TripQuery) {
+  const queryString = tripQueryToQueryString(query);
+  const { data, error, isLoading } = useSWR<NonTransitTripApiReturnType>(
```
adriansberg, line 46: Not using `useSWRInfinite` here?

src/page-modules/assistant/client/journey-planner/index.ts
```diff
@@ -1,16 +1,59 @@
...
+      {},
+    );
 
-  const result = await fetch(`/api/assistant/trip?${queryString}`);
+  const isLoadingInitialData = !data && isLoading;
+  const isLoadingMore = isValidating && size > 1;
+  return {
+    trips: data,
+    isLoading: isLoadingInitialData,
+    isError: error,
```
adriansberg, line 49: `error` is of type `Error | undefined`, should there be any check so that the boolean value is correct or is it enough that it is defined of `isError` to be correct?

mortennordseth: Since it is named isError I think this should be a boolean, yes 👍 Updated now. Could return the error as `error` too, but I think this is fine.

src/page-modules/assistant/client/journey-planner/index.ts
mortennordseth: No, that is unnecessary since we are only after one example of the non-transit trip types (no pagination or infinite scroll here)

src/page-modules/assistant/trip/index.tsx
```diff
@@ -3,48 +3,36 @@ import {
...
+  const { trips, isLoading, loadMore, isLoadingMore } =
+    useTripPatterns(tripQuery);
+  const { nonTransitTrips } = useNonTransitTrip(tripQuery);
 
-  const nonTransits = Object.entries(nonTransitTrips);
+  const nonTransits = nonTransitTrips ? Object.entries(nonTransitTrips) : [];
 
-  if (tripPatterns.length === 0) {
+  if (isLoading) {
+    return <EmptySearch isSearching={isLoading} type="trip"></EmptySearch>;
```
adriansberg, line 32:
```suggestion
    return <EmptySearch isSearching={isLoading} type="trip" />;
```

src/page-modules/assistant/layout.tsx
```diff
@@ -203,16 +203,15 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
...
             buttonProps={{ type: 'submit' }}
-            state={isSearching ? 'loading' : undefined}
           />
         </div>
       </form>
 
       <section className={style.contentContainer}>
         <EmptySearch
-          isSearching={isSearching}
+          isSearching={false}
```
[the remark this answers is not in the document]
mortennordseth: Yes, with this change the layout does not know anything about the search state. This is now handled by the Trip component.

src/page-modules/assistant/client/journey-planner/index.ts
adriansberg: Won't revalidation on focus/mount etc come into play?

src/page-modules/assistant/layout.tsx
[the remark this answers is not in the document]
adriansberg: > the layout does not know anything about the search state
So essentially no need for an `isSearching` at all in this file?

src/page-modules/assistant/client/journey-planner/index.ts
mortennordseth: Not sure what you are thinking of 🤔 `useSWRInfinite` vs. `useSWR` should not behave differently regarding that

src/page-modules/assistant/layout.tsx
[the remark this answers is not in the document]
mortennordseth: Its only function currently is to disable the "Finn avganger" button while we use `router.push`. We could change the name to `isNavigating` or `isPerformingSearchNavigation` etc?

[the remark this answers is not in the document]
mortennordseth: I updated to `isPerformingSearchNavigation`, it makes more sense I think

adriansberg (approved):

PR #202, opened 2024-01-12
feat: adding the option to include via location when searching after a trip.

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/16239

### Background

FRAM's customer service has come up with a wish to be able to add a "via" field to make it easier to plan routes where travelers go via a location.

#### Illustrations
<details>
Screenshot 2024-02-06 at 09 55 40
Screenshot 2024-02-06 at 09 55 57
Screenshot 2024-02-06 at 09 56 08

</details>

### Proposed solution

Add the ability to specify one via location to a trip search.

### Acceptance Criteria
- [x] The user is able to both add and remove via location to the search. 
- [x] The trip result shown to the user are correct.
- [x] Both the initial overview page and the trip detail page shows the correct information when via location is added.


adriansberg:
Should the number of via locations be dynamic, and should they be located closer to the from/to?

E.g. how Google's implementation is:

2024-01-18 at 10 43 47@2x

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -496,23 +635,33 @@ function generateSingleTripQueryString(
...
 
 function isTripsQueryVariables(a: any): a is TripsQueryVariables {
   return a && 'from' in a && 'to' in a && 'when' in a && 'arriveBy' in a;
 }
 
+function isTripsViaQueryVariables(a: any): a is TripsQueryVariables {
+  return a && 'from' in a && 'to' in a && 'when' in a && 'via' in a;
+}
+
+function isTripsViaWithDetailsQueryVariables(a: any): a is TripsQueryVariables {
```
jonasbrunvoll, line 661: Remember to remove unused function.

jonasbrunvoll:
> Should the number of via locations be dynamic, and should they be located closer to the from/to?
> 
> E.g. how Google's implementation is:
> 
> 2024-01-18 at 10 43 47@2x

src/page-modules/assistant/server/journey-planner/journey-gql/via-trip-with-details.gql
```diff
@@ -0,0 +1,127 @@
+query ViaTripsWithDetails(
+  $from: Location!
+  $to: Location!
+  $when: DateTime
+  $via: [ViaLocationInput!]!
+) {
+  viaTrip(
+    from: $from
+    to: $to
+    searchWindow: "PT2H" # 1 hour
```
jonasbrunvoll, line 10: Should maybe be a param instead of a static value?

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -349,10 +480,22 @@ function inputToLocation(
...
 }
+function inputToViaLocation(input: TripViaInput, direction: 'via') {
+  return {
+    place: input[direction].id,
+    coordinates: {
+      latitude: input[direction].geometry.coordinates[1],
+      longitude: input[direction].geometry.coordinates[0],
+    },
+    name: input[direction].name,
+    minSlack: 'PT120S',
```
jonasbrunvoll, line 491: Should maybe be a params instead of a static values?

mortennordseth:
We should do something about the result list not being shown properly: 

image

src/page-modules/assistant/server/journey-planner/journey-gql/via-trip-with-details.gql
mortennordseth: I think it's fine that this is a static value, as this is probably not something we want to change between searches. I see that you write `searchWindow: "PT2H" # 9 hours` in another file, and 1 hour here. Looking at the value it looks to be 2h. Any reason not to use 9 hours as done for regular trips?

src/page-modules/assistant/details/utils.ts
```diff
@@ -30,7 +30,12 @@ export function tripQueryStringToQueryParams(
   const tripQueryVariables = parseTripQueryString(queryString);
   if (!tripQueryVariables) return undefined;
 
-  const { arriveBy, from, to } = tripQueryVariables.query;
+  const { from, to } = tripQueryVariables.query;
+
+  let arriveBy = undefined;
+  if ('arrivedBy' in tripQueryVariables.query)
```
mortennordseth, line 36:
```suggestion
  if ('arriveBy' in tripQueryVariables.query)
```

src/page-modules/assistant/details/utils.ts
```diff
@@ -30,7 +30,12 @@ export function tripQueryStringToQueryParams(
...
+    arriveBy = tripQueryVariables.query.arrivedBy;
```
mortennordseth, line 37:
```suggestion
    arriveBy = tripQueryVariables.query.arriveBy;
```

src/page-modules/assistant/server/journey-planner/journey-gql/via-trip-with-details.gql
jonasbrunvoll: So it should be "PT9H" to ensure a 9 hours search windows? The "PT2H" value is used for testing to limit the search window, and the comment comes simply from copy pasting from Shamash.

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -190,32 +202,85 @@ export function createJourneyApi(
...
+            ...tripQuery.query,
+            arriveBy: false,
+            ...DEFAULT_JOURNEY_CONFIG,
+          },
+        });
+
+        formatedResult = { trip: result.data.trip.tripPatterns };
       }
 
+      if (result.error || result.errors) throw result.error || result.errors;
```
mortennordseth, line 274: I think we should do this check earlier, right? In case of an error we don't have data to format

src/page-modules/assistant/server/journey-planner/journey-gql/via-trip-with-details.gql
mortennordseth: We saw that in some cases in Møre and Romsdal there is a while between the buses, yielding no results when searching with a small search window. To fix this we increased the search window to 9 hours, but keeping the number of trip patterns to 8. I don't think this affects the search time, but might be wrong. I see that we can specify `numTripPatterns` in the viaTrip query as well. You could compare the search time when searching with 9 hours and 8 number of trip patterns vs. 2 hours of search window while the number of trip patterns is not specified. 

I can't see any documentation of the `searchWindow` value, but `PT9H` sounds like 9 hours to me. For the trip query, we use 540 (minutes).

src/page-modules/assistant/server/journey-planner/journey-gql/via-trip.gql
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+}
+
+fragment notice on Notice {
+  id
+  text
+}
+
+fragment situation on PtSituationElement {
```
mortennordseth, line 97: It looks like these fragments are the same as the ones defined in trip.gql and can probably be removed

src/page-modules/assistant/server/journey-planner/journey-gql/via-trip.gql
```diff
@@ -0,0 +1,116 @@
...
+    via: $via
+  ) {
+    ...viaTrip
+  }
+}
+
+fragment viaTrip on ViaTrip {
+  routingErrors {
+    ...routingError
+  }
```
mortennordseth, line 21: Since we are not using this data it can be removed. Not quite sure how this data works and it might be something that we should look into, but we should add it when we need the data

src/page-modules/assistant/server/journey-planner/index.ts
jonasbrunvoll: I´ll ensure that the check is performed after receiving the result👍

src/page-modules/assistant/server/journey-planner/journey-gql/via-trip.gql
jonasbrunvoll: Removed👍

src/page-modules/assistant/server/journey-planner/journey-gql/via-trip.gql
jonasbrunvoll: Removed👍

src/pages/api/assistant/trip.ts
```diff
@@ -26,13 +26,22 @@ export default handlerWithAssistantClient<TripApiReturnType>({
...
+              transportModes,
+              cursor: tripQuery.cursor!,
+            })
+          : await client.trip({
+              from: tripQuery.from!,
+              to: tripQuery.to!,
+              searchTime: tripQuery.searchTime,
+              transportModes,
+              cursor: tripQuery.cursor!,
+            }),
```
mortennordseth, line 44: Is there a reason why we have `viaTrip` and `trip`, but we use the same `singleTrip` function for both via and regular trips? Would it be better if we have the functions `trip` and `singleTrip` and that those functions handled which query to use? Both `viaTrip` and `trip` return `Promise<TripData>`. In my opinion, the `via` parameter is more or less the same as any other optional parameter, such as the cursor.

src/page-modules/assistant/types.ts
```diff
@@ -15,12 +15,17 @@ export type TripInput = {
   cursor?: string;
 };
 
+export type TripViaInput = TripInput & {
+  via: GeocoderFeature;
+};
+
```
mortennordseth, line 21: If my other comment makes sense to you, we could change the `type TripInput` to 

```ts
type TripInput = {
from: GeocoderFeature;
to: GeocoderFeature;
via?: GeocoderFeature;
...
}
```

I think the question comes down to: is there a difference between a trip that has a via location and a trip that does not?

src/page-modules/assistant/types.ts
```diff
@@ -36,6 +41,10 @@ export const TripQuerySchema = z.object({
   searchMode: searchModeSchema.optional(),
   searchTime: z.number().optional(),
   cursor: z.string().optional(),
+  viaId: z.string().optional(),
+  viaLon: z.number().optional(),
+  viaLat: z.number().optional(),
+  viaLayer: z.union([z.literal('address'), z.literal('venue')]).optional(),
```
mortennordseth, line 47: This is another relevant example. We don't have a `ViaTripQuerySchema`, but the `TripQuery` **can** contain a via location. I think we should be consistent in the typings here.

mikaelbr:
I think maybe the via-field should have a headline in the filter section 🤔

src/page-modules/assistant/layout.tsx
```diff
@@ -187,6 +199,20 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
...
+                    label={t(PageText.Assistant.search.input.via)}
+                    placeholder={t(PageText.Assistant.search.input.placeholder)}
+                    onChange={onViaSelected}
+                    selectedItem={tripQuery.via ?? undefined}
+                    autocompleteFocusPoint={tripQuery.via ?? undefined}
+                    button={
+                      <ClearButton
+                        className={style.searchInputButton}
+                        onClear={onClear}
+                      />
```
mikaelbr, line 213: I think this should only be visible when there is a search active

jonasbrunvoll: Is the idea then that the via input field only becomes visible after the user has clicked on the headline in the filter section?

jonasbrunvoll:
Is the idea then that the via input field only becomes visible after the user has clicked on the headline in the filter section?

jonasbrunvoll:
> I think maybe the via-field should have a headline in the filter section 🤔

That the via input field only becomes visible after the user has clicked on the headline in the filter section?

mikaelbr:
See comment on specific location of where the headers should be IMO https://github.com/AtB-AS/planner-web/pull/202/files#r1469568821

src/page-modules/assistant/layout.tsx
```diff
@@ -187,6 +199,20 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
                     data={transportModeFilter}
                     onChange={onTransportFilterChanged}
                   />
+
+                  <Search
```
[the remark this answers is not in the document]
jonasbrunvoll: Okay, got it!

src/page-modules/assistant/layout.tsx
```diff
@@ -82,6 +86,11 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
     setIsSwapping(false);
   };
 
+  const onClear = () => {
```
mortennordseth, line 89: Naming this something like `onClearViaLocation` would make this more readable I think.

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -152,12 +162,85 @@ export function createJourneyApi(
...
+        const tripPatternsFromViaTo = findTripPatternsFromViaTo(
+          tripPatternCombinationList,
+          tripPatternsFromVia,
+          tripPatternsViaTo,
+        );
+
+        const data: RecursivePartial<TripData> = mapResultToTrips(
+          {
+            nextPageCursor: undefined,
+            previousPageCursor: undefined,
```
mortennordseth, line 216: The "load more trip patterns" function on the assistant page won't work without the cursors. I see that the viaTrip input can take a cursor, but I'm not sure how to retrieve one for the current search. Another approach is to modify the search time when fetching more trip patterns, but that could potentially result in duplicate results and/or missing patterns. Did you look into this?

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -189,36 +272,76 @@ export function createJourneyApi(
...
-        TripsWithDetailsQueryVariables
-      >({
-        query: TripsWithDetailsDocument,
-        variables: {
-          ...tripQuery.query,
-          arriveBy: false,
-          ...DEFAULT_JOURNEY_CONFIG,
-        },
-      });
+      let formatedResult;
```
mortennordseth, line 280:
```suggestion
      let formattedResult;
```

src/page-modules/assistant/server/journey-planner/index.ts
mortennordseth: If this is default and never changing I would move it to the graphql query document instead.

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -560,3 +697,58 @@ function mapAndFilterNotices(notices: GraphQlNotice[]): Notice[] {
...
+      tripPatterns.push({
+        expectedStartTime: tripPatternsFromVia[combo.from].expectedStartTime,
+        expectedEndTime: tripPatternsViaTo[combo.to].expectedEndTime,
+        legs: [
+          ...tripPatternsFromVia[combo.from].legs,
+          ...tripPatternsViaTo[combo.to].legs,
+        ],
+      }),
+    );
+  });
```
mortennordseth, line 752: Would these functions be a good candidate for some unit tests? I guess there could be made tests for other parts of this code as well, such as testing that the correct query is used when `via` is defined and so on.

mortennordseth (commented):
Starting to look good! Added a few suggestions :)

mortennordseth:
I see that the filters are not working properly, most likely because the input for transport modes is a bit different for the via-trip query than for the regular trip query

src/page-modules/assistant/layout.tsx
jonasbrunvoll: Agree🤔

src/page-modules/assistant/server/journey-planner/index.ts
jonasbrunvoll: I've looked into option number two a bit by investigating if the search time could be passed as a parameter to the query, instead of the current option where the search time is a stativ value. As you mentioned, this could potentially lead to duplicate results and/or missing patterns. I can also imagine that by increasing the search window, the query will take longer to complete. Of course, the query can be fast enough even with increased search time.

But I´ll take a look if it is possible to retrieve a cursor from the current search.

src/page-modules/assistant/server/journey-planner/index.ts
jonasbrunvoll: Yes, I should probably add a few unit tests to verify the behavior of the feature when via is defined. I´ll look into it👍

src/page-modules/assistant/layout.tsx
```diff
@@ -96,6 +105,9 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
       setValuesWithLoading({ transportModeFilter }, true),
     500,
   );
+  const onViaSelected = async (via: GeocoderFeature) => {
+    setValuesWithLoading({ via });
+  };
```
mikaelbr, line 110:
```suggestion
  const onViaSelected = (via: GeocoderFeature) => {
    setValuesWithLoading({ via });
  };
```

src/page-modules/assistant/layout.tsx
```diff
@@ -187,6 +199,32 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
                     data={transportModeFilter}
                     onChange={onTransportFilterChanged}
                   />
+
+                  <div>
+                    <Typo.p
```
mikaelbr, line 204: Maybe this should have a more semantically correct element like h3 🤔

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -157,12 +167,86 @@ export function createJourneyApi(
 
       const from = inputToLocation(input, 'from');
       const to = inputToLocation(input, 'to');
-
       const when =
         input.searchTime.mode !== 'now'
           ? new Date(input.searchTime.dateTime)
           : new Date();
 
+      if (input.via) {
```
mikaelbr, line 175: I think all of this could maybe be seperated into different functions? This function body is quite big now. Hard to separate

src/page-modules/assistant/layout.tsx
jonasbrunvoll: Are you sure? Both the filter title and the search assistant title are both <Typo.p textType="body__primary--bold" className={style.heading}>.

src/page-modules/assistant/server/journey-planner/index.ts
jonasbrunvoll: The fetchMore button is now disabled when tripQuery.via is defined. 

Screenshot 2024-02-05 at 11 18 56

src/page-modules/assistant/layout.tsx
adriansberg: Might make sense to change those to be `h3` as well possibly.

jonasbrunvoll: Yes, that might make sense. In any case, I think that all of these titles should have the same font.

src/page-modules/assistant/assistant.module.css
```diff
@@ -50,10 +50,12 @@
   grid-area: alternatives;
   width: 100%;
   background-color: var(--static-background-background_accent_1-background);
-  overflow: hidden;
```
adriansberg, line 53: This was here because of the animation. The collapsing/uncollapsing makes the content overflow the container and not look too great. Is it necessary to remove this?

src/page-modules/assistant/layout.tsx
```diff
@@ -82,6 +86,11 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
     setIsSwapping(false);
   };
 
+  const onClearViaLocation = () => {
+    tripQuery.via = null;
```
adriansberg, line 90: Instead of modifying the object and setting `via` to `null`, I would probably pass a new object that spreads `tripQuery` and sets `via: null`. 

Not sure if modifying the object in this case results in any side effects, 


```ts
  const onClearViaLocation = () => {
    setValuesWithLoading({ ...tripQuery, via: null });
  };
```

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -200,36 +212,75 @@ export function createJourneyApi(
...
+        // Create list to make all trip pattern combinations more accessable.
+        const tripPatternCombinationList = findTripPatternCombinationsList(
+          tripPatternCombinations,
+        );
+
+        // Find trip patterns from-via and via-to destination.
+        const tripPatternsFromVia = tripPatternsPerSegment[0].tripPatterns;
+        const tripPatternsViaTo = tripPatternsPerSegment[1].tripPatterns;
+
+        // Find all possible trip patterns where the legs from the from-via location and the via-to location are concatinated.
```
adriansberg, line 249:
```suggestion
        // Find all possible trip patterns where the legs from the from-via location and the via-to location are concatenated.
```

src/page-modules/assistant/server/journey-planner/index.ts
adriansberg: I agree with Morten.

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -571,3 +636,147 @@ function mapAndFilterNotices(notices: GraphQlNotice[]): Notice[] {
...
+    combinations.map((combination) => ({
+      from: combination.from!,
+      to: combination.to!,
+    })),
+  );
+}
+
+export function findTripPatternsFromViaTo(
+  tripPatternCombinations: { from: number; to: number }[][],
+  tripPatternsFromVia: ViaTripsQuery['viaTrip']['tripPatternsPerSegment'][0]['tripPatterns'],
```
adriansberg, line 664: Is there no separate type for these kind of trip patterns?

src/translations/pages/assistant.ts
```diff
@@ -14,6 +14,14 @@ export const Assistant = {
       ),
       from: _('Fra', 'From', 'Frå'),
       to: _('Til', 'To', 'Til'),
+      via: {
+        label: _(
+          'Vil du reise via en mellomdestinasjon?',
+          'Do you want to travel via an intermidiate destination?',
```
adriansberg, line 20:
```suggestion
          'Do you want to travel via an intermediate destination?',
```

src/page-modules/assistant/types.ts
adriansberg: Is this handled, @jonasbrunvoll ?

src/page-modules/assistant/assistant.module.css
jonasbrunvoll: overflow:hidden was removed because the content from the via search input field was overlaped by the trip results as shown in the image. When removed, the content overlaps the trip results instead. 

<details>

Screenshot 2024-02-07 at 20 57 58

Screenshot 2024-02-08 at 08 28 43

</details>

src/page-modules/assistant/layout.tsx
jonasbrunvoll: I will check it out👍

src/page-modules/assistant/server/journey-planner/index.ts
jonasbrunvoll: Will fix.

src/translations/pages/assistant.ts
jonasbrunvoll: Will fix👍

src/page-modules/assistant/server/journey-planner/index.ts
jonasbrunvoll: The types ViaTripsQuery['viaTrip']['tripPatternsPerSegment'][0]['tripPatterns'] and ViaTripsWithDeatilsQuery['viaTrip']['tripPatternsPerSegment'][0]['tripPatterns'] comes from via-trip-generated.ts and via-trips-with-details-generated.ts. It does not exist any type  of separate types for these trip patterns.

src/page-modules/assistant/assistant.module.css
adriansberg: I see. What we can do is to add temporary CSS attributes with framer-motion, so that when collapsing happens the container does have `overflow: hidden;` to avoid the contents spilling out of it. Can be achieved by using `transitionEnd`, like this: 

2024-02-08 at 14 13 27


But not sure if it's a big deal. I guess most users barely notice it anyway.

PR #203, opened 2024-01-15
fix: make aria-labels in the search time selector visible to the screen reader

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/16349

### Background

Received feedback that the screen reader was unable to read the search time radio buttons, due the attribute specification: 'aria-hidden:true'. 
  
### Proposed solution

Remove the attribute specification: 'aria-hidden:true'. 
  
### Acceptance Criteria

- [ ] The screen reader is able to read the aria label of the radio button.

### DIFF
--- a/public/widget/GYJwhgtkA/2.8.0/planner-web.mjs
+++ b/public/widget/GYJwhgtkA/2.8.0/planner-web.mjs
@@ -405,7 +405,7 @@ function Oe({ URL_BASE: t }, e) {
               value="now"
               checked=""
             />
-            <span aria-hidden="true" class="${o.selector_option__label}">
+            <span class="${o.selector_option__label}">
               <span class="${o.selector_option__text}" id="${r}-now">
                 ${e.searchTime.now}
               </span>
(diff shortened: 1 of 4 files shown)


mortennordseth:
Did you use the NVDA screen reader to test this? I am just curios if this does anything since VoiceOver reads the radio button text just fine. There might be some differences between the screen readers, but the span element with the actual text does not have the hidden label, so I was wondering if the hidden-label makes all sub-elements of that element hidden by screenreaders or not?

adriansberg:
@mikaelbr what was the original reason to add aria-hidden to these? Would removing them break anything unforeseen?

jonasbrunvoll:
> Did you use the NVDA screen reader to test this? I am just curios if this does anything since VoiceOver reads the radio button text just fine. There might be some differences between the screen readers, but the span element with the actual text does not have the hidden label, so I was wondering if the hidden-label makes all sub-elements of that element hidden by screenreaders or not?

I did not use the NVDA screen reader, but the default macos VoiceOver to test this. I tested with VoiceOver only after I had implemented the changes.

mikaelbr:
I belive these where just copied from the output of the original form 🤔

adriansberg (approved):
I see no point of having `aria-hidden="true"` of any of these, as long as screen readers work as expected without it. :shipit:

PR #204, opened 2024-01-16
refactor: refactor the transport mode filter to URL state

adriansberg:
Also fetching which filters to be shown from Firestore instead of having them hard coded. There is no constraints to what the ids can be for the filter configuration, so the types were simplified to just be string arrays and not a union type of the available ids.

> [!NOTE]
> There is no debouncing implemented yet, a search is made immediately when a filter is changed. Not quite sure how to solve the debouncing when the checkboxes are dependent on the route to change to display the correct state. Somehow updating the URL, but not performing a search. 🤷

Closes https://github.com/AtB-AS/kundevendt/issues/15895


src/modules/transport-mode/filter/filter.tsx
```diff
@@ -30,9 +38,13 @@ export default function TransportModeFilter({
...
             aria-label={t(ComponentText.TransportModeFilter.allA11y)}
-            checked={Object.values(filterState).every(Boolean)}
+            checked={!filterState || filterState.length === data.length}
             onChange={(event) => {
-              onFilterChange(setAllValues(filterState, event.target.checked));
+              if (event.target.checked) {
+                onChange(null);
+              } else {
+                onChange(['none']);
+              }
```
mikaelbr, line 47:
```suggestion
             onChange(event.target.checked ? null : ['none']);
```

src/page-modules/assistant/layout.tsx
```diff
@@ -35,23 +34,28 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
...
-        ...tripQuery,
-        ...override,
-        searchTime,
-      },
-      transportModeFilter,
-    );
+  // Loading the transport mode filter data here instead of in the component
+  // avoids the data loading when the filter is mounted which causes the
+  // height to be incorrect and the animation to be janky.
+  const { data } = useSWRImmutable(
```
mortennordseth, line 45: Could consider changing this to `transportModeFilter` to be more explicit

mortennordseth (approved):
Looks good 👍 One suggestion that you can consider

src/page-modules/assistant/layout.tsx
adriansberg: Updated

PR #205, opened 2024-01-16
fix: improve semantics in widget  

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/16351

### Background

Received feedback that the semantics of the widget could be improved.

### Proposed solution

The widget is updated to from using div - tag and p - tag to fieldset - tag and legend - tag where requested.

### Acceptance Criteria
- [] The behavior of the widget remains unaffected by the semantic change.

### DIFF
--- a/public/widget/GYJwhgtkA/2.8.0/planner-web.mjs
+++ b/public/widget/GYJwhgtkA/2.8.0/planner-web.mjs
@@ -390,8 +390,8 @@ function Oe({ URL_BASE: t }, e) {
       </button>
     </div>
   `, $ = (r, c = !0) => y`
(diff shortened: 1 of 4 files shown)


mortennordseth (changes requested):
This also requires some new css because the fieldset for example adds a border, changing the current design. 

With this change: 

image

Today:

image

Also, I see that the changes are applied to v2.8.0 that were released yesterday. I think we should only apply the changes to src/widget.ts, and then, before the next release, we generate a new version with these (and possibly other) changes. At that time we know the correct version number for these changes.

mortennordseth (approved):
Looks good :)

PR #206, opened 2024-01-17
refactor: optimize trip pattern fetching for improved performance

mortennordseth:
This PR addresses performance issues related to fetching trip patterns. Previously, we were fetching 8 trip patterns in a single request before presenting the results to the user. This approach was causing a noticeable delay. For example, a search from Volda to Trondheim would take approximately 6 seconds to complete. 

To improve this, I've modified our approach to initially return only the first page of results. This change significantly reduces the time to display the first trip patterns to about 1.8 seconds for the same search. Note that these measurements were done locally and will differ in production, but it gives an idea of how it can improve the time to first result for the user. 

The `useTripPatterns` hook has been updated to continuously load more data until we reach the desired number of trip patterns. This allows us to display results to the user more quickly while continuing to fetch additional data in the background.

(Partially) closes https://github.com/AtB-AS/kundevendt/issues/16346


mikaelbr:
Screenshot 2024-01-17 at 11 03 19

src/page-modules/assistant/client/journey-planner/index.ts
```diff
@@ -41,14 +46,38 @@ export function useTripPatterns(tripQuery: FromToTripQuery) {
...
+    if (size >= MAX_NUMBER_OF_INITIAL_SEARCH_ATTEMPTS) {
+      return;
+    }
+    if (numberOfTripPatterns >= INITIAL_NUMBER_OF_WANTED_TRIP_PATTERNS) {
+      return;
+    }
+    setSize(size + 1);
+  }, [numberOfTripPatterns, size, setSize, isLoading, isValidating]);
+
+  const isLoadingFirstPage = !data && isLoading;
```
adriansberg, line 70: This is set to false after the first trip is loaded? Would `isLoadingFirstTrip` be more fitting?

src/page-modules/assistant/client/journey-planner/index.ts
```diff
@@ -41,14 +46,38 @@ export function useTripPatterns(tripQuery: FromToTripQuery) {
       {},
     );
 
-  const isLoadingInitialData = !data && isLoading;
+  useEffect(() => {
+    if (data) {
+      setNumberOfTripPatterns(
+        data.reduce((acc, curr) => acc + curr.tripPatterns.length, 0),
```
adriansberg, line 52: Doing a `flatMap` of `data.tripPatterns` and taking the length of that would result in the same number? No problem having it like this either, just me that needs to take a few moments to remember how `reduce` works every time I see it 😅

src/page-modules/assistant/client/journey-planner/index.ts
mortennordseth: Yes, false because we have data at that point. I agree, updated 👍

src/page-modules/assistant/client/journey-planner/index.ts
mortennordseth: Yes, that would result in the same number 😃 I think both have the same time complexities, but `flatMap` might be easier to read, not sure 🤔 With reduce we don't have to create a new array though, but I guess we don't have a memory issue 😅

adriansberg (approved):
Seems to be working as intended 👍

PR #207, opened 2024-01-17
feat: debounce search when filter change

adriansberg:
Added a debounce to the function that updates the URL (performing a search) so that a search is not performed immediately after changing a filter, but after a set time. I set this time to be 500ms. In addition, with the debounce the function is not called for every single filter change, but only the last change within the last 500ms.

Closes https://github.com/AtB-AS/kundevendt/issues/16003


mortennordseth (approved):
Nice 👍

PR #208, opened 2024-01-18
feat: configurable journey planner api configurations

mortennordseth:
Because the journey planner API configurations differ for AtB, NFK, and FRAM we need to have them configurable. This makes that possible by using the org data.

Closes https://github.com/AtB-AS/kundevendt/issues/16416


adriansberg:
How is this configured for the app? In the bff?

mikaelbr (changes requested):
I think this should be part of org data instead. It isn't needed as a secret but just easily accessible information

mortennordseth:
I've moved the configuration to the org data and updated the values to match what is used in the app. 

> How is this configured for the app? In the bff?

This is configured in the app. Some are editable with preferences

orgs/schema-validation.json
```diff
@@ -75,6 +75,27 @@
             "supportUrl"
           ],
           "type": "object"
+        },
+        "journeyAPIConfigurations": {
```
adriansberg, line 79:
```suggestion
        "journeyApiConfigurations": {
```

orgs/schema-validation.json
```diff
@@ -75,6 +75,27 @@
...
+          "additionalProperties": false,
+          "properties": {
+            "WAIT_RELUCTANCE": {
```
adriansberg, line 82: Do these need to be capitalised when they're not defined as constants, but as properties in an object?

adriansberg (approved):
Looks good, other than some nitpicky comments.

PR #211, opened 2024-01-22
feat: adds cancelled trips

mikaelbr:


mortennordseth:
Verified that this includes cancelled trips: 

image

mortennordseth:
Screenshots of the implementation: 

image

image

mortennordseth:
Any suggestions for how to improve the mobile view is appreciated. Could maybe remove the left icon for smaller devices? 🤔 

image

toretefre:
Isn't a greyed out tile background colour used other places to make it even more explicit that something is cancelled?

mortennordseth:
> Isn't a greyed out tile background colour used other places to make it even more explicit that something is cancelled?

It's used for trips that have passed. That may be confusing. The warning icon makes it different though. The departure page shows canceled departures like this: 

image

The problem with strike-through is that it can be hard to read the text, so I went with the text (Kansellert) behind the header.

src/page-modules/assistant/details/details-header/index.tsx
```diff
@@ -36,27 +36,57 @@ export default function DetailsHeader({ tripPattern }: DetailsHeaderProps) {
...
+              PageText.Assistant.trip.tripPattern.isCancelled.title,
+            ).toUpperCase()})`}
+        </Typo.h2>
+      </div>
       <div className={style.tripDetails}>
         <div className={style.date}>
           <MonoIcon icon="time/Date" />
-          <Typo.p textType="body__primary">{weekdayAndDate}</Typo.p>
+          <Typo.p
+            textType={isCancelled ? 'body__primary--strike' : 'body__primary'}
```
adriansberg, line 73: IIRC, in other places where we use strike-through to indicate that it is cancelled, the colour of the text is also changed. Should it be here as well? (I don't know how the contrast ratio is with a colour change.)

adriansberg (approved):
LGTM

src/page-modules/assistant/details/details-header/index.tsx
mortennordseth: You are correct. The secondary text color is used on the departure page to indicate a canceled departure. I talked to the product owner and he agreed that we could iterate on this in the next release. We should also talk to the designers at AtB do look into this when implementing it in the app

PR #215, opened 2024-01-24
feat: update icon sizes

rosvik:
👋 Hello hello, first time contributor here! 

Don't mind me, I'm just updating some icon sizes.

- Changes small icon size to 16
- Adds a new icon size, xSmall = 12

fixes https://github.com/AtB-AS/kundevendt/issues/3834

### DIFF
--- a/src/components/icon/utils.ts
+++ b/src/components/icon/utils.ts
@@ -6,6 +6,8 @@ export function useSize(size: SizeProps) {
   const theme = useTheme();
 
   switch (size) {
+    case 'xSmall':
+      return theme.icon.size.xSmall;
     case 'small':
       return theme.icon.size.small;
     case 'normal':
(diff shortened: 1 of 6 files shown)


src/components/icon/utils.ts
```diff
@@ -1,11 +1,13 @@
-import { useTheme } from '@atb/modules/theme';
+import { useTheme, theme } from '@atb/modules/theme';
 
-export type SizeProps = 'small' | 'normal' | 'large' | 'x-large';
+export type SizeProps = keyof typeof theme.light.icon.size | 'x-large';
```
adriansberg, line 3: `'x-large'` not in the theme?

adriansberg: It's not in use in planner-web either, I see. And there's a TODO there that it should be in the design system at some point 😅

rosvik: Oh 😅 I'll just remove it, then!

adriansberg (approved):

PR #217, opened 2024-01-31
feat: trip cache

mortennordseth:
This PR adds a TTL cache for the trip search page. 

Closes https://github.com/AtB-AS/kundevendt/issues/16346

### DIFF
--- a/src/page-modules/assistant/__tests__/assistant.test.tsx
+++ b/src/page-modules/assistant/__tests__/assistant.test.tsx
@@ -178,10 +178,6 @@ describe('assistant page', function () {
   });
 });
 
-vi.mock('./', () => ({
-  getServerSideProps: vi.fn(),
-}));
-
 describe('assistant page', function () {
   it('should add trip to trip cache', () => {
     const fromToTripQuery: FromToTripQuery = {


src/page-modules/assistant/server/trip-cache.ts
```diff
@@ -0,0 +1,11 @@
...
+
+let tripCache: TTLCache<unknown, unknown> | null = null;
+
+export function getCache(): TTLCache<unknown, unknown> {
+  if (!tripCache) {
+    tripCache = new TTLCache({ ttl: 120000 });
+  }
+
+  return tripCache;
+}
```
mortennordseth, line 11: Is this the approach you would go for, @mikaelbr ? Importing this wherever it's needed. Seems that the cache object would be recreated at times, such as when navigating to a new site, but not sure if that is a problem. Another approach I was thinking of is to create this object in the `/src/page-modules/assistant/server/journey-planner/index.ts` file and the `getServerSideProps` fetch from the cache using a (new) endpoint.

mikaelbr: Not what you are asking, but I think trip-cache possibly should have functions for explicit caching. So instead of having this at consumer site:

```ts
if (tripCache.has(`/api/assistant/trip?${queryString}`)) {
        return tripCache.get(`/api/assistant/trip?${queryString}`) as TripData;
      }
```

You would have

```ts
const potentiallCached = getAssistantTripIfCached(query);
```

And thus hide the cache key generation and avoid error source where you don't match proper cache keys.

mikaelbr: Also I think maybe what you are saying about placing it in `getServerSideProps` is the approach I would go for.

mortennordseth: Good idea. And then it would make sense to create it as a module I think

jonasbrunvoll (approved):
Looks good😊

src/pages/assistant/index.tsx
```diff
@@ -51,6 +52,17 @@ export const getServerSideProps = withGlobalData(
         };
       }
 
+      const potential = getAssistantTripIfCached(tripQuery);
```
mikaelbr, line 55: Should we also just add http cache headers here? (potentially as a separate PR)

mortennordseth: Sure, I'll look into it

PR #221, opened 2024-02-01
Update trip parameters for AtB

tormoseng:
Update the query parameters for trip search to be equal as in the AtB app. The changes are as the default parameters from Entur. See also [Slack](https://mittatb.slack.com/archives/CHLDG8C30/p1706776408713929).

Maybe we should change the values, but this takes a bit of testing, and for now it's better to have the values equal.

<details>
<summary>From Shamash</summary>

Screenshot 2024-02-01 at 09 24 49

</details>

## Acceptance criteria

- [ ] Should test and see that the trip search results are the same in travel planner and app for AtB

### DIFF
--- a/orgs/atb.json
+++ b/orgs/atb.json
@@ -37,10 +37,10 @@
     }
   },
   "journeyApiConfigurations": {
-    "waitReluctance": 1.5,
-    "walkingReluctance": 1.5,
+    "waitReluctance": 1.0,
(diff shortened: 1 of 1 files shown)


mortennordseth (approved):
Aha, I didn't know shamash displayed Enturs default values. I will update for FRAM and NFK as well.

PR #223, opened 2024-02-01
chore: adds initial e2e case

mikaelbr:
Adds initial structure for running E2E tests on pull requests and all staging environments every 3 hours. At some point this could also be extended to running as smoke tests for the production environment.

Currently only includes one test, but want to merge this before adding more test cases

Part of https://github.com/AtB-AS/kundevendt/issues/16713

### DIFF
--- a/e2e-tests/search-test.spec.ts
+++ b/e2e-tests/search-test.spec.ts
@@ -1,6 +1,8 @@
 import { test, expect } from '@playwright/test';
 
-test('test', async ({ page }) => {
+test('Test fetching Kristiansund - Molde and loading more after first result', async ({
+  page,
+}) => {
   await page.goto(process.env.E2E_URL ?? 'http://localhost:3000');
   await page.getByRole('textbox', { name: 'From' }).click();
   await page.getByRole('textbox', { name: 'From' }).fill('Kristiansund');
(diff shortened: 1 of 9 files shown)


mortennordseth (approved):
Nice! Just a thought: a bit strange that we reference elements by their English text in e2e tests, and by their Norwegian text in our unit tests. I guess adding test IDs to all of these elements would be overkill. I don't mind that it is like this, just something I thought of.

tormoseng (approved):
Pent!

e2e-tests/search-test.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+test('Test fetching Kristiansund - Molde and loading more after first result', async ({
+  page,
+}) => {
+  await page.goto(process.env.E2E_URL ?? 'http://localhost:3000');
+  await page.getByRole('textbox', { name: 'From' }).click();
+  await page.getByRole('textbox', { name: 'From' }).fill('Kristiansund');
+  await page.getByRole('option', { name: 'Kristiansund Kristiansund' }).click();
+  await page.getByRole('textbox', { name: 'To' }).click();
+  await page.getByRole('textbox', { name: 'To' }).click();
```
jonasbrunvoll, line 11: Is there a reason for why the exact same line of code is repeated twice on line 10 and line 11?🤔

jonasbrunvoll (approved):

mikaelbr:
@mortennordseth Good point. This is automatically recorded by Playwright in default playwright browsers, so the setting is in English. COULD do scripts manually and using test IDs. But want to test out if recorders can be valuable

e2e-tests/search-test.spec.ts
mikaelbr: Probably hiccup with the recorder. Can remove

PR #224, opened 2024-02-02
fix: journey planner API config for FRAM and NFK

mortennordseth:
The current values that are used are Entur default values. This updates the values to be the same as in the app. 

Closes https://github.com/orgs/AtB-AS/projects/10/views/13?pane=issue&itemId=49668512 for FRAM and NFK.


jonasbrunvoll (approved):

tormoseng:
👍

PR #226, opened 2024-02-05
fix: filter out locations where locality is undefined from the travel planner's input field.

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/16688

### What was wrong, and what was expected behavior?

When clicking on the Ålesund geopoint in travel planner web, the locality attribute  was undefined in the input box.
This is probably due to missing municipality field in OSM. 

How to replicate:
Enter Ålesund in input. Click geopoint at top. See undefined in input box. If running locally, the application may crash. 

### Illustration problem
<details>
<Summary>screenshots/problem</Summary>

img2undefined

img1Undefined

</details>

### Illustration solution

<details>
<Summary>screenshot/solution</Summary>

Screenshot 2024-02-05 at 16 06 29

</details>

### Solution
- Set locality to null if noe defined
- Only display search locality if not it is defined."

### DIFF
--- a/src/components/search/search.tsx
+++ b/src/components/search/search.tsx
@@ -70,37 +70,35 @@ export default function Search({
 
           <ul className={style.menu} {...getMenuProps()}>
             {isOpen &&
-              data
-                ?.filter((item) => item.locality !== undefined)
-                .map((item, index) => (
-                  <li
-                    className={andIf({
-                      [style.item]: true,
-                      [style.itemHighlighted]: highlightedIndex === index,
-                    })}
-                    key={item.id}
-                    {...getItemProps({
-                      index,
-                      item,
-                    })}
-                  >
-                    <div className={style.itemIcon} aria-hidden>
-                      <VenueIcon categories={item.category} />
-                    </div>
-                    <span className={style.itemName}>
-                      {highlightSearchText(inputValue, item.name).map(
-                        ({ part, highlight }) => {
-                          if (!part) return null;
-                          if (highlight)
-                            return <span key={item.id + part}>{part}</span>;
-                          else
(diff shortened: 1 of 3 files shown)


adriansberg:
IMO we should not filter out locations without a locality, but instead don't add a locality to the item in the search field if it is not present, much like the widget.

For instance it can be handled in `geocoderFeatureToString`, check for locality and only show the feature name if not present.

jonasbrunvoll:
To me it looks like the undefined locality attribute in some cases makes the travel planner crash. For example when searching after 'Ålesund'.

For some reason, when searching after 'Campus Ålesund', 'Møre og Romsdal' is automatically added as locality attribute, even when 'Møre og Romsdal' was not originally listed as the location of Campus Ålesund.

adriansberg:
Looks like the autocomplete results have no locality, but when perform a search the page is loaded again (routing) the from place (and to place for that matter) is being reversed based on their coordinates, which then results in a locality.


> To me it looks like the undefined locality attribute in some cases makes the travel planner crash. For example when searching after 'Ålesund'.

How does it crash for you? I'm able to both select and perform a search without an issue.

jonasbrunvoll:
> Looks like the autocomplete results have no locality, but when perform a search the page is loaded again (routing) the from place (and to place for that matter) is being reversed based on their coordinates, which then results in a locality.
> 
> > To me it looks like the undefined locality attribute in some cases makes the travel planner crash. For example when searching after 'Ålesund'.
> 
> How does it crash for you? I'm able to both select and perform a search without an issue.

It crashed when the application was run locally due to the undefined locality attribute. This was not a problem in prod. After the last commit, this is not a problem in either prod, or if the application is run locally.

adriansberg (approved):

PR #227, opened 2024-02-06
feat: line filter in search

mortennordseth:
This adds the functionality to filter lines in the travel search. 

image

Closes https://github.com/AtB-AS/kundevendt/issues/16679

### DIFF
--- a/orgs/atb.json
+++ b/orgs/atb.json
@@ -1,6 +1,7 @@
 {
   "$schema": "./schema-validation.json",
   "orgId": "atb",
+  "orgLineIdPrefix": "ATB:Line:2_",
 
   "supportEmail": "kundeservice@atb.no",
 
--- a/orgs/fram.json
+++ b/orgs/fram.json
@@ -2,6 +2,7 @@
   "$schema": "./schema-validation.json",
 
   "orgId": "fram",
+  "orgLineIdPrefix": "MOR:Line:",
 
   "supportEmail": "fram@mrfylke.no",
 
--- a/orgs/nfk.json
+++ b/orgs/nfk.json
@@ -1,6 +1,8 @@
 {
   "$schema": "./schema-validation.json",
   "orgId": "nfk",
+  "orgLineIdPrefix": "NOR:Line:12_",
+
   "supportEmail": "sentrumsterminalen@boreal.no",
 
   "fylkeskommune": {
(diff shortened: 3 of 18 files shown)


jonasbrunvoll:
The trip planner does not work quite as it should when the trip has first been specified and then removed. The trip search does not work Initially when specifying the line number, the search returns trips from the specified line. Example image 1, where line 800 is specified. Then, when removing the specified line, the search does not    

### Illustration
<details>
<Summary>screenshots/video</Summary>

Screenshot 2024-02-06 at 14 59 45

Screenshot 2024-02-06 at 14 59 54


</details>

mortennordseth:
> The travel planner does not work quite as it should when the trip has first been specified and then removed.

> Example: When fist specifying the line 800, the travel planner returns trips from the line 800. Then, when removing the specified line 800, the search does not return any results.

Fixed :)

jonasbrunvoll:
> > The travel planner does not work quite as it should when the trip has first been specified and then removed.
> 
> > Example: When fist specifying the line 800, the travel planner returns trips from the line 800. Then, when removing the specified line 800, the search does not return any results.
> 
> Fixed :)

Nice🙌

jonasbrunvoll:
Another observation. When searching after trips from Ålesund to Molde, the result is empty both when line 100 and when line 905 is specified. When none line is specified, trips with both of these line numbers are returned. 


<details>
<Summary>screenshots/video</Summary>
Screenshot 2024-02-06 at 16 52 55
Screenshot 2024-02-06 at 16 53 11
Screenshot 2024-02-06 at 16 53 23
</details>

mortennordseth:
This is because you are using AtBs version of the planner web while filtering for a FRAM bus. If you do the same search in FRAMs version, it would look like this: 

image

We need to add a message for empty results though

jonasbrunvoll:
> This is because you are using AtBs version of the planner web while filtering for a FRAM bus. If you do the same search in FRAMs version, it would look like this:
> 
> image
> We need to add a message for empty results though

Ah, I see! Yes, adding a message for empty results would be helpful👍

jonasbrunvoll (approved):
Looks good🙌

mikaelbr:
Think the tests here fails due to not merged changes from main 🤔

mikaelbr:
I'll clean up and fix merge conflicts here

PR #229, opened 2024-02-09
fix: issue with caching when selecting transport filter

mikaelbr:
The cache doesn't bust on transport mode, causing it not to work as expected. This adds a test case (e2e) to verify fix and fixes the issue.


jonasbrunvoll (approved):
Looks good🚀

e2e-tests/assistant-view.spec.ts
```diff
@@ -0,0 +1,116 @@
+import { test, expect } from '@playwright/test';
+
+test('Test fetching Kristiansund - Molde and loading more after first result', async ({
```
adriansberg, line 3: What is the naming convention for these tests?

adriansberg:
Should the tests be working?

mikaelbr:
Problem here is that FRAM tests runs on AtB tests so search results are differently sorted 😅

e2e-tests/assistant-view.spec.ts
mikaelbr: Should start with should if we're following BDDish naming. I didn't start with that until next test tho and didn't update this one 😅

src/pages/assistant/index.tsx
```diff
@@ -53,6 +53,7 @@ export const getServerSideProps = withGlobalData(
       }
 
       const potential = getAssistantTripIfCached(tripQuery);
+      console.log('potential', potential);
```
adriansberg, line 56: ⌫

mikaelbr:
It all depends on what tags we use (conventional commits allows for defining your own). But that said, this will be a fix as it now fixes cache issue which is a real bug  😄 

Most cases adding tests is part of fix after either discovering a bug or when developing new things.

src/page-modules/assistant/server/trip-cache.ts
```diff
@@ -15,16 +14,27 @@ function getTripCacheInstance(): TTLCache<string, TripData> {
...
+  const keys = Object.keys(valuesToCreateCacheKey).sort();
+  const values = keys.map(
+    (key) => valuesToCreateCacheKey[key as keyof FromToTripQuery],
+  );
+  const cacheKey = JSON.stringify(values);
+
+  return 'assistant-trip-' + Buffer.from(cacheKey).toString('base64');
+}
+
+function createStableCacheKeyFromObject(object: Record<string, unknown>) {}
```
adriansberg, line 40: Removable?

adriansberg (approved):

PR #230, opened 2024-02-12
fix: improve button border contrast when organization is set to fram.

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/16871

### Background

Fram's color palette is such that certain buttons do not meet the requirements of UU when the buttons are in :focus

#### Illustrations
<details>

https://github.com/AtB-AS/planner-web/assets/59939294/e28796c8-8c87-4c6b-a082-b360728d9143

https://github.com/AtB-AS/planner-web/assets/59939294/8d84d41c-2332-424f-a1b8-0fe324254098

</details>

### Proposed solution

Apply a special .css class to the button component when the component is rendered on frammr.no.

### DIFF
--- a/src/components/button/utils.tsx
+++ b/src/components/button/utils.tsx
@@ -6,6 +6,7 @@ import { LoadingIcon } from '../loading';
 export type ButtonModes =
   | 'interactive_0'
   | 'interactive_0--bordered'
+  | 'fram_interactive_0--bordered'
   | 'interactive_1'
   | 'interactive_2'
   | 'interactive_3'
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -254,7 +254,11 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
           <Button
             title={t(PageText.Assistant.search.buttons.find.title)}
             className={style.button}
-            mode="interactive_0--bordered"
+            mode={
+              urls.homePageUrl.name === 'frammr.no'
+                ? 'fram_interactive_0--bordered'
+                : 'interactive_0--bordered'
+            }
             disabled={
               !tripQuery.from || !tripQuery.to || isPerformingSearchNavigation
             }
--- a/src/components/button/button.module.css
+++ b/src/components/button/button.module.css
@@ -101,30 +101,46 @@
 
 /* Interactive button 0 bordered */
 .button--interactive_0--bordered,
-.button--interactive_0--bordered:visited {
+.button--interactive_0--bordered:visited,
+.button--fram_interactive_0--bordered,
+.button--fram_interactive_0--bordered:visited {
   background-color: var(--interactive-interactive_0-default-background);
   color: var(--interactive-interactive_0-default-text);
   box-shadow: inset 0 0 0 var(--border-width-slim)
     var(--interactive-interactive_0-default-text);
 }
-.button--interactive_0--bordered:hover {
+
+.button--interactive_0--bordered:hover,
+.button--fram_interactive_0--bordered:hover {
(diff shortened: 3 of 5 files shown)


src/components/button/utils.tsx
```diff
@@ -6,8 +6,10 @@ import { LoadingIcon } from '../loading';
 export type ButtonModes =
   | 'interactive_0'
   | 'interactive_0--bordered'
+  | 'interactive_0--bordered_alternative'
   | 'interactive_1'
   | 'interactive_2'
+  | 'interactive_2_alternative'
```
adriansberg, line 12: Variations should be named by separating the base and variant with double hyphens; for example 'interactive_2--alternative'`.

Furthermore, I think "alternative" might be a bit vague and not descriptive enough, but at the same time I don't have any good proposed solution so it's not a big deal.

src/components/button/utils.tsx
```diff
@@ -6,8 +6,10 @@ import { LoadingIcon } from '../loading';
 export type ButtonModes =
   | 'interactive_0'
   | 'interactive_0--bordered'
+  | 'interactive_0--bordered_alternative'
```
adriansberg, line 9: I think we can use a hyphen instead of an underscore to separate "bordered" and "alternative" → 'interactive_0--bordered-alternative'

src/components/button/utils.tsx
jonasbrunvoll: Okey, I will change the naming, and keep that in minde when naming .css classes in the future. 

I'll also see if I can find a better name. Initially, I thought of a name to distinguish between light and dark background, but it didn't really fit with its color palette.

src/components/button/utils.tsx
jonasbrunvoll: Okey, I´ll fix👍

src/components/button/utils.tsx
```diff
@@ -6,8 +6,10 @@ import { LoadingIcon } from '../loading';
 export type ButtonModes =
   | 'interactive_0'
   | 'interactive_0--bordered'
+  | 'interactive_0--bordered--customized_off_palette'
```
adriansberg, line 9: Would something like `'interactive_0--bordered-light-outline'` suffice?

jonasbrunvoll: Yes, that works too.

src/components/button/utils.tsx
```diff
@@ -6,8 +6,10 @@ import { LoadingIcon } from '../loading';
 export type ButtonModes =
   | 'interactive_0'
   | 'interactive_0--bordered'
+  | 'interactive_0--bordered-light-outline'
   | 'interactive_1'
   | 'interactive_2'
+  | 'interactive_2--customized_off_palette'
```
adriansberg, line 12:
```suggestion
  | 'interactive_2--light-outline'
```

And the occurrences in the css file that is.

adriansberg (approved):

src/page-modules/assistant/layout.tsx
```diff
@@ -246,15 +248,22 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
                 : t(PageText.Assistant.search.buttons.alternatives.more)
             }
             className={style.button}
-            mode="interactive_2"
+            mode={
+              orgId === 'fram' && !isDarkMode
```
adriansberg, line 252: Should only this one check for dark mode?

jonasbrunvoll: Yes. The interactive_2 button both looks good and meets the UU requirements in darkmode.  Therefore, the  interactive_2--light-outline should only be used in light mode.

adriansberg (approved):

PR #231, opened 2024-02-13
 feat: uncheck flight as default way of transport when organization is nfk 

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/16683

### Background

#### Illustrations
<details>

Screenshot 2024-02-13 at 14 01 16

</details>

### Proposed solution

At the initial render, if the organization is nfk, set the transport mode filter to all filters but the flight filter.

### Acceptance Criteria
- [ ] All filters but air is initially set when the organization is nfk.
- [ ] The user should be able to set the flight filter after the initial render.
- [ ] The trip planner works as it did before this update.

### DIFF
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -16,7 +16,12 @@ import { PageText, useTranslation } from '@atb/translations';
 import { FocusScope } from '@react-aria/focus';
 import { AnimatePresence, motion } from 'framer-motion';
 import { useRouter } from 'next/router';
-import { FormEventHandler, PropsWithChildren, useState } from 'react';
+import {
+  FormEventHandler,
+  PropsWithChildren,
+  useEffect,
+  useState,
+} from 'react';
 import style from './assistant.module.css';
 import { FromToTripQuery } from './types';
 import { createTripQuery } from './utils';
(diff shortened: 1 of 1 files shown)


src/page-modules/assistant/layout.tsx
```diff
@@ -53,6 +59,15 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
     getTransportModeFilter,
   );
 
+  useEffect(() => {
+    if (orgId === 'nfk') {
+      tripQuery.transportModeFilter =
```
adriansberg, line 64: Directly modifying the tripQuery object won't update the URL, will it? The rendering of the filter will work I guess, since `filterState={tripQuery.transportModeFilter}` is passed as a prop. 

I guess the URL is set correctly when the first search is performed.

> [!IMPORTANT]
> If the transport mode filter is set already, we should not overwrite it and change it. E.g. the URL is copied from somewhere with a set transport mode filter (which may intentionally include air transport), that filter should be untouched.

adriansberg:
Do they only want this for the planner and not the app? 

If it's a requested change in both the planner and the app, I think a better approach might be to expand on the data structure with a flag that says something about it being default selected or not.

jonasbrunvoll:
> Do they only want this for the planner and not the app?
> 
> If it's a requested change in both the planner and the app, I think a better approach might be to expand on the data structure with a flag that says something about it being default selected or not.

Although I can't see that it is specified, I assume that nfk wants this in both the app and web solution to achieve the same behavior.

adriansberg:
IIRC there are some types [here](https://github.com/AtB-AS/fare-product-type-configs/blob/main/schema-definitions/travelSearchFilters.json) that are used.

PR #232, opened 2024-02-19
fix: automatically resets to the current date if the current month or current year is reselected twice.

jonasbrunvoll:
### Background

When reselecting the current the year twice, the date is automatically set to current date. At the moment,  this is not the case when reselecting the current month twice. To make the user flow as predictable as possible, this function should be the same for both year selection and month selection     

### Proposed solution

Update to automatically rest to current date if reselecting the current month or current year twice.

### DIFF
--- a/src/modules/search-time/selector/index.tsx
+++ b/src/modules/search-time/selector/index.tsx
@@ -56,11 +56,14 @@ export default function SearchTimeSelector({
 
   const isPastDate = (selectedDate: string) => {
     const today = new Date().toISOString().split('T')[0];
+    const year = initialDate.getFullYear().toString();
+    const month = initialDate.getMonth().toString().padStart(2, '0');
+    const day = initialDate.getDate().toString().padStart(2, '0');
+    const formatedInitialDate = `${year}-${month}-${day}`;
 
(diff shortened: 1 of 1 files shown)


adriansberg:
But is reseting the selection to the current date the wanted behaviour, if clicking the same year/month twice?

jonasbrunvoll:
> But is reseting the selection to the current date the wanted behaviour, if clicking the same year/month twice?

We have not received any feedback that says that this is a wanted behaviour. I created this PR because when using Firefox with date selector on the mm/dd/yyyy format,  I often find my self trying to change month back to current month, but is restricted since the current month combined with the selected day, is a previous date. Initially, I often think that the date selector is broken, before remembering that I am not allowed to select a previous date.  By adding the reset behaviour, the user receive some type of feedback instead of nothing happening.

jonasbrunvoll:
> > but is restricted since the current month combined with the selected day
> 
> Oh! So if the month is changed to the current month, but the day is a previous day, _then_ the date is reset to the current date?

Yes! That is the meaning of this implementation. Instead of having to set the day to current or upcoming months manually before changing the month, the date is set to current date.

adriansberg (approved):

PR #233, opened 2024-02-22
fix: update styling to make the linefilter and the search component appear more similar.

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/16994

### Background

Currently, the via search input field and the line search input field have different appearances. These fields should look the same.

#### Illustrations
Current solution / proposed solution  
<details>

Screenshot 2024-02-22 at 09 33 08

Screenshot 2024-02-22 at 09 32 49
</details>

### Proposed solution

Use the same search component and styling to make sure that the input field for via search and the input field for line search appear similar.

### DIFF
--- a/src/page-modules/assistant/line-filter/index.tsx
+++ b/src/page-modules/assistant/line-filter/index.tsx
@@ -38,13 +38,20 @@ export default function LineFilter({ filterState, onChange }: LineFilterProps) {
       <Typo.h3 textType="body__primary--bold" className={style.heading}>
         {t(PageText.Assistant.search.lineFilter.label)}
       </Typo.h3>
-      <input
-        className={style.input}
-        type="text"
-        placeholder={t(PageText.Assistant.search.lineFilter.placeholder)}
-        value={localFilterState}
-        onChange={onChangeWrapper}
-      />
+
+      <div className={style.labelInputContainer}>
+        <label className={style.label}>
+          {t(PageText.Assistant.search.lineFilter.description)}
+        </label>
+
+        <input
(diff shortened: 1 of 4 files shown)


src/page-modules/assistant/line-filter/index.tsx
```diff
@@ -38,13 +38,20 @@ export default function LineFilter({ filterState, onChange }: LineFilterProps) {
...
+        </label>
+
+        <input
+          className={style.input}
+          type="text"
+          placeholder={t(PageText.Assistant.search.lineFilter.placeholder)}
+          value={localFilterState}
+          onChange={onChangeWrapper}
+        />
+      </div>
```
mikaelbr, line 54: This could also have been a component I think 🤔

src/components/line-search/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import style from './line-search.module.css';
+
+type LineSearchProps = {
+  label: string;
+  placeholder: string;
+  value: string;
+  onChange: (event: ChangeEvent<HTMLInputElement>) => void;
+};
+
+export default function LineSearch({
```
mikaelbr, line 11: Maybe this is more a reusable labeled input rather than LineSearch specific input? Meaning, I think it has value to think of this as more a fundamental component that can be reused by other components to ensure same design on labeled inputs.

jonasbrunvoll: ah I see! Yes I can change this to labeled input component.

mikaelbr (approved):

PR #236, opened 2024-02-23
fix: update button style in widget to match button in travel planner when fram is the organization.  

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/16996

### Background

Currently, the widget search button and the travel planner search button have dissimilar appearance when FRAM is the organisation. These buttons should have the same appearance.  

#### Illustrations
<details>
 

https://github.com/AtB-AS/planner-web/assets/59939294/ef5d2679-709b-4b78-9528-85cec452275d

https://github.com/AtB-AS/planner-web/assets/59939294/7204ecb0-dc9e-4ba3-a374-01a70f75ba4f

</details>

### Proposed solution

Style the widget search button the same way as the search button currently implemented in the travel planner. 

### Acceptance Criteria

- [ ] Button in widget and travel planner have the same appearance. 
- [ ] The new styling does not affect the behavior of the button in the widget button.

### DIFF
--- a/src/widget/widget.ts
+++ b/src/widget/widget.ts
@@ -19,7 +19,8 @@ const html = String.raw;
 
 const MODULE_VERSION = process.env.MODULE_VERSION;
 const COMPRESSED_ORG = process.env.COMPRESSED_ORG;
-const useDefaultButtonStyle = COMPRESSED_ORG !== 'GYJwhgtkA' ? true : false;
+const ORG_ID = process.env.ORG_ID;
+const useDefaultButtonStyle = ORG_ID !== 'fram' ? true : false;
 
 function createSettingsConstants(urlBase: string) {
   if (!urlBase?.startsWith('http')) {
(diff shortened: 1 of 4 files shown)


src/widget/widget.ts
```diff
@@ -19,6 +19,7 @@ const html = String.raw;
 
 const MODULE_VERSION = process.env.MODULE_VERSION;
 const COMPRESSED_ORG = process.env.COMPRESSED_ORG;
+const useDefaultButtonStyle = COMPRESSED_ORG !== 'GYJwhgtkA' ? true : false;
```
mortennordseth, line 22: Not a big fan of this 🤔 Are we sure that there are no other options to this, then having an org-specific class?

jonasbrunvoll: I am nota big fan of this approach either, but the style.buttonLightOutline should only be applied when the organization is set to FRAM to make the widget and the web travel planner appear similar. I am not aware of another approach, except maybe reorganising the fram colours to achieve the same style. But I am sure that is a lot of work, and will also lead to other unwanted changes in the styling🤷‍♂️

adriansberg: Not sure, tbh. How about using `getOrgData()`, is that an option?

adriansberg (approved):
👍

PR #238, opened 2024-02-23
fix: set org specific examples for line numbers to filter on

mikaelbr:
Fixes https://github.com/AtB-AS/kundevendt/issues/17037


src/translations/pages/assistant.ts
```diff
@@ -407,3 +404,28 @@ export const Assistant = {
...
+      },
+    },
+  },
+  fram: {
+    search: {
+      lineFilter: {
+        example: _(
+          'Eksempel: 901, 800',
+          'Example: 901, 800',
+          'Eksempel: 901, 800',
```
mortennordseth, line 426: When thinking about it, these line examples might not make much sense for FRAM. These were just added as an example when I added the feature. Not sure if FRAM has a line 800 at all 🤔

mikaelbr: I thought about that also, just accepted that you had thought of this as lazy as I was 😅

mikaelbr: Checking with Christoffer. 701 is also probably relevant for Molde

adriansberg (approved):

PR #240, opened 2024-02-26
feat: add functionality to specify whether a filter that should be initially checked.   

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/16683

### Background

- [ ] Uncheck flight as default way of transport in Nordland
- [ ] Uncheck flight FRAM
- [ ] Uncheck train FRAM

### Proposed solution
-  If the isUncheckedByDefault property of a filter is defined, set the filter to be initially unchecked.

### DIFF
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -16,7 +16,12 @@ import { PageText, useTranslation } from '@atb/translations';
 import { FocusScope } from '@react-aria/focus';
 import { AnimatePresence, motion } from 'framer-motion';
 import { useRouter } from 'next/router';
-import { FormEventHandler, PropsWithChildren, useState } from 'react';
+import {
+  FormEventHandler,
+  PropsWithChildren,
+  useEffect,
+  useState,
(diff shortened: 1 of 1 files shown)


src/page-modules/assistant/layout.tsx
```diff
@@ -54,6 +59,13 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
     getTransportModeFilter,
   );
 
+  useEffect(() => {
+    tripQuery.transportModeFilter =
```
adriansberg, line 63: Maybe `setValuesWithLoading()` should be used here as well? Like we looked at some time ago with some other changes. Or maybe not? I guess if the `setValuesWithLoading` is used we get an (possibly) unwanted route change.

And it's important to check whether `transportModeFilter` is set already, then the `isUncheckedByDefault` should not be taken into account and the filter should remain as is, since there's a chance the user has set the filters themselves. Haven't tested this code, but that might be how this works with the optional chaining and nullish coalescing.

PR #242, opened 2024-02-28
feat: disable travel planner flight filter by default 

jonasbrunvoll:
Temporary solution for task https://github.com/orgs/AtB-AS/projects/10/views/13?pane=issue&itemId=51482158

### Background

#### Illustrations

<details>
<summary>screenshot</summary>

Screenshot 2024-02-28 at 12 36 11

</details>

### Proposed solution

Disabling the travel planner flight filter for for all organisations until a more sophisticated solution where each organisation can configure filter preferences. 

### Acceptance Criteria

- [ ] Flight filter is unchecked by default.
- [ ] Searching after trips with flight can still be performed by checking the flight filter.

### DIFF
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -59,15 +59,6 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
     getTransportModeFilter,
   );
 
-  useEffect(() => {
-    if (tripQuery.transportModeFilter === null)
-      onTransportFilterChanged(
-        transportModeFilter
-          ?.filter((filter) => filter.id !== 'air')
-          .map((filter) => filter.id) ?? null,
-      );
-  }, [transportModeFilter]);
-
   const setValuesWithLoading = async (
     override: Partial<FromToTripQuery>,
     replace = false,
@@ -131,8 +122,23 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
   const { urls, orgId } = getOrgData();
   const { isDarkMode } = useTheme();
 
+  /*
+   * Temporary solution until firebase configuration is in place.
+   */
+  useEffect(() => {
+    if (tripQuery.transportModeFilter === null)
+      onTransportFilterChanged(
+        transportModeFilter
+          ?.filter(
+            (filter) =>
+              !filter.modes.some((mode) => mode.transportMode === 'air'),
+          )
+          .map((filter) => filter.id) ?? null,
+      );
+  }, [transportModeFilter]);
+
   /**
-   * Temprorary solution to disable line filter for some orgs until
+   * Temporary solution to disable line filter for some orgs until
    * we have a working solution for all orgs.
    */
   const disableLineFilter =
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -50,7 +50,6 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
   const [isPerformingSearchNavigation, setIsPerformingSearchNavigation] =
     useState(false);
   const [geolocationError, setGeolocationError] = useState<string | null>(null);
-  const [defaultFilters, setDefaultFilters] = useState<string[] | null>(null);
 
   // Loading the transport mode filter data here instead of in the component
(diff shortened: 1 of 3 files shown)


src/page-modules/assistant/layout.tsx
```diff
@@ -45,6 +50,7 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
   const [isPerformingSearchNavigation, setIsPerformingSearchNavigation] =
     useState(false);
   const [geolocationError, setGeolocationError] = useState<string | null>(null);
+  const [defaultFilters, setDefaultFilters] = useState<string[] | null>(null);
```
adriansberg, line 53: Do we need a separate state for this? Don't we just need to overwrite the state on the initial load, and can use the URL state to know whether it should be done or not?

jonasbrunvoll: No, I do not think we need it. I added it because I when i tried to find out why the flight filter did not appear in preview and forgot to remove it. Will fix now👍

src/page-modules/assistant/layout.tsx
```diff
@@ -54,6 +60,14 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
     getTransportModeFilter,
   );
 
+  useEffect(() => {
+    setDefaultFilters(
+      transportModeFilter
+        ?.filter((filter) => filter.icon.transportMode !== 'air')
```
adriansberg, line 66: Should base the filter on the transport mode itself and not the icon, I think.

adriansberg, line 66: Will this always filter out air, regardless if the user has manually enabled it to see flights?

jonasbrunvoll: No, the user can manually set check the air filter after the initial search.

adriansberg: Right, because the `transportModeFilter` never changes, which means that the `useEffect` only runs once. In theory at least.

mortennordseth: It sounds like the URL state is not prioritized and that this would override it 🤔

adriansberg: My idea was that something like [this](https://github.com/AtB-AS/planner-web/pull/243) would work.

src/page-modules/assistant/layout.tsx
```diff
@@ -54,6 +59,15 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
     getTransportModeFilter,
   );
 
+  useEffect(() => {
+    if (tripQuery.transportModeFilter === null)
+      onTransportFilterChanged(
+        transportModeFilter
```
adriansberg, line 65: Could this logic be added directly to the `onTransportFilterChanged` and this hook can be removed?

src/page-modules/assistant/layout.tsx
adriansberg: Scratch that, everything didn't work.

I think the approach with using `onTransportFilterChanged` might be a good solution.

Also need to make a change with "All", something like below. Setting the filter to all of the options instead of `null`. (Need to handle when `data` is undefined or null or whatever).

https://github.com/AtB-AS/planner-web/blob/2dec3a0de733d707520f6b406dfd81ce2d0f7c2d/src/modules/transport-mode/filter/filter.tsx#L53-L55

```ts
            onChange={(event) => {
              onChangeWrapper(
                event.target.checked ? data?.map((option) => option.id) : [],
              );
            }}
```

jonasbrunvoll: Okey👍

src/page-modules/assistant/layout.tsx
jonasbrunvoll: I am not sure that is possible, as the transportModeFilter initially is null before the useSWRImmutable() have returned the results.

adriansberg: There are also other reasons that I did not see before.

But this should be fixed:
`ESLint: React Hook useEffect has missing dependencies: 'onTransportFilterChanged' and 'tripQuery.transportModeFilter'. Either include them or remove the dependency array.(react-hooks/exhaustive-deps)`

PR #244, opened 2024-02-29
chore: trigger playwright on deployment state

mikaelbr:
Changes playwright to trigger on deployment state.

Makes it more stable when running on PR


jonasbrunvoll (approved):

adriansberg (approved):
I think this is a good change. Experiencing instability with timeouts etc.

PR #248, opened 2024-03-01
fix: improve minor styling errors in widget

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/17110
### Background

-  When nfk is set as the organization, the width of the search button is too wide. 
-  The text colour of the active tab is set to black. The colour should be withe, as in the travel planner on web.  should.
- The selected tab should have bold font, as in the travel planner on web.   

#### Illustrations

<details>
Before / after

Image

Skjermbilde 2024-03-01 kl  09 51 48

### Proposed solution

- Fix the mentioned bullet points.

### DIFF
--- a/src/widget/widget.module.css
+++ b/src/widget/widget.module.css
@@ -63,12 +63,13 @@
 .tabSelected {
   text-decoration: none;
   border-bottom: 3px solid var(--static-background-background_accent_0-text);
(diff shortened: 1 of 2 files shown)


adriansberg (approved):
:shipit:

PR #251, opened 2024-03-04
fix: enable line filter for FRAM

mortennordseth:
Currently, the NFK env variable `NEXT_PUBLIC_DISABLE_LINE_FILTER` is used for FRAM because FRAM hasn't defined one. I've added the env variable to all organisations to prevent this from happening. We could look into "resetting" the env variables between each build as well, but this will fix the issue for now.


mikaelbr (approved):

adriansberg:
Wouldn't it be easier to instead make sure that `disableLineFilter` is false when the env variable is not present? E.g. by using nullish coalescing.

mortennordseth:
> Wouldn't it be easier to instead make sure that `disableLineFilter` is false when the env variable is not present? E.g. by using nullish coalescing.

I think we could improve this by resetting every env variable before each build. Not sure what is the best approach to this, but I'll merge this for now and we can look into it.

PR #253, opened 2024-03-05
fix: conditional render assistant page when transport mode filters are defined.  

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/17141

Some users sometimes experience that flight is checked by default. Must investigate. Flight should always be unchecked by default until a more sofisticated solution for configurating transport mode filters for each organization is ready.

### How to replicate:
According to feedback from customer service, all transport mode filters (including flight) are checked if the navigating directly to the travel planner, without using the widget.    

After testing in Safari, Chrome, Arc, Firefox and Microsoft Edge, I have not been able to replicate the described bug myself. 

A few times I have experienced that the trips query have completed before the transportModeFilter from the useSWRImmutable() is defined, making some trip results appear on the screen in halv a second, before the transportModeFilter is defined, repeating the query due to the useEffect().   

### Proposed Solution
Add a condition to only render the AssistantLayout when the useSWRImmutable() have completed and transportModeFilter is defined.

### DIFF
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -147,6 +147,8 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
   const disableLineFilter =
     process.env.NEXT_PUBLIC_DISABLE_LINE_FILTER === 'true';
 
+  if (transportModeFilter === null) return null;
+
   return (
     <div>
       <div className={style.homeLink__container}>
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -50,6 +50,7 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
   const [isPerformingSearchNavigation, setIsPerformingSearchNavigation] =
     useState(false);
   const [geolocationError, setGeolocationError] = useState<string | null>(null);
+  const [noneFilters, setNoneFilters] = useState(false);
 
   // Loading the transport mode filter data here instead of in the component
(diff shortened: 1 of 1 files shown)


src/page-modules/assistant/layout.tsx
```diff
@@ -107,9 +108,11 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
   const onToSelected = async (to: GeocoderFeature) =>
     setValuesWithLoading({ to });
   const onTransportFilterChanged = debounce(
-    async (transportModeFilter: string[] | null) =>
-      setValuesWithLoading({ transportModeFilter }, true),
-    500,
+    async (transportModeFilter: string[] | null) => {
+      setNoneFilters(transportModeFilter?.length === 0 ? true : false);
```
adriansberg, line 112: Can't this be a derived state, from `transportModeFilter`, or is it necessary to have its own state?

jonasbrunvoll: I belive it is necessary to have its own state. This was the only way is was able to uncheck all travel mode filters and update the URL accordingly😅

adriansberg: Isn't possible to just check the length of `transportModeFilter` where `!noneFilters` is checked for? 

I get that the `noneFilters` is used to trigger the `useEffect`, but wouldn't it be sufficient to either use the `transportModeFilter` in the condition and set it to `null` in the else clause? Or just pass `null` in `onTransportFilterChanged` if `transportModeFilter` is not present?

PR #254, opened 2024-03-06
feat: extend planner-web to support Troms

mortennordseth:
This PR extends the planner-web to support Troms (staging). Troms hasn't set up their Mapbox account yet, and I've used FRAMs credentials for now. 

Closes https://github.com/AtB-AS/kundevendt/issues/17160


jonasbrunvoll (approved):
Looks good🙌

orgs/troms.json
```diff
@@ -0,0 +1,47 @@
+{
+  "$schema": "./schema-validation.json",
+  "orgId": "troms",
+  "orgLineIdPrefix": "TRO:Line:1_",
+
+  "supportEmail": "",
+
+  "fylkeskommune": {
+    "logoSrc": "/troms_fylkeskommune.svg",
```
adriansberg, line 9: IIRC this was a request from NFK as well, but Troms may want to have both the logo for fylkeskommune and the logo for Fylkestrafikk, like they have on their current page:
2024-03-06 at 12 29 48

adriansberg (approved):

orgs/troms.json
mortennordseth: True! Troms Fylkestrafikk is undergoing a rebranding which will result in changes to their name, logo, etc. I don't think we should put any work into this until the branding is done as this could change, but it might be something we should do later 👍

adriansberg: Agree!

PR #259, opened 2024-03-12
fix: adds validation to line filter

mikaelbr:
Fixes https://github.com/AtB-AS/kundevendt/issues/17245

### DIFF
--- a/src/components/labled-input/index.tsx
+++ b/src/components/labled-input/index.tsx
@@ -1,21 +1,36 @@
+import { andIf } from '@atb/utils/css';
 import { ChangeEvent } from 'react';
+import { MessageBox } from '../message-box';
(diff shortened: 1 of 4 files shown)


src/components/labled-input/index.tsx
```diff
@@ -1,29 +1,53 @@
+import { andIf } from '@atb/utils/css';
 import { ChangeEvent } from 'react';
+import { MessageBox } from '../message-box';
```
mortennordseth, line 3:
```suggestion
import { MessageBox } from '@atb/components/message-box';
```

mortennordseth: I think?

mortennordseth (approved):
Looks good and works as expected 👍

PR #261, opened 2024-03-12
feat: extend line filtering to support nfk   

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/17072

### Background

Hanne found an issue with the line filter for NFK: https://github.com/AtB-AS/kundevendt/issues/17037#issuecomment-1963487733

The planner web filters lines by ID. For FRAM and AtB this works as line ID = line number (aka publicCode in Journey Planner API). Customers can therefore search by line number.

```
  {
    "id": "MOR:Line:100",
    "publicCode": "100"
  },
  {
    "id": "MOR:Line:101",
    "publicCode": "101"
  },
```

 This is not the case for NFK. For example do line 1, 2 and 200 in Bodø have line ID 8001, 8002 and 8200. Also, some lines have the same line number, but different line IDs. For example:

```
 {
        "id": "NOR:Line:12_8371",
        "publicCode": "1"
  },
 {
        "id": "NOR:Line:12_8636",
        "publicCode": "1"
  },
 {
        "id": "NOR:Line:12_8001",
        "publicCode": "1"
  },
```

#### Illustrations
<details>

Skjermbilde 2024-03-12 kl  15 57 54

Skjermbilde 2024-03-12 kl  15 58 06

</details>

### Proposed solution
In the proposed solution, all lines are fetched at the initial render of the line-filter component. Then, when a valid line number is written in the line filter component, any corresponding line id is included to the line whitelist of the trip search.  

### Acceptance Criteria
- [x] Line filter works for all organizations (AtB, FRAM and NFK). 
- [x] Specified filters should not disappear when refreshing the page. 
- [x] Lines query should only be run initially and when refreshing the page. Otherwise, the lines should be fetched from cache.   

### Test input 
Trip suggestions for testing. 
- **AtB**: From *Rønningsbakken, Trondheim*, to *Nidarosdomen, Trondheim*.
- **FRAM**: From *Molde, Molde*, to *Batnfjordsøra, Gjemnes*.
- **NFK**: From *Bodø sentrum, Bodø*, to *Universitetet, Bodø*.


src/page-modules/assistant/line-filter/index.tsx
```diff
@@ -1,39 +1,61 @@
...
     setLocalFilterState(lineFilter);
-
     if (!lineFilter) {
       onChange(null);
     } else {
-      const linesWithPrefix = lineFilter
+      const publicCodes = lineFilter
         .split(',')
-        .map((line) => `${orgLineIdPrefix}${line.trim()}`);
+        .map((publicCode) => `${publicCode.trim()}`);
```
adriansberg, line 35:
```suggestion
        .map((publicCode) => publicCode.trim());
```

src/page-modules/assistant/line-filter/index.tsx
```diff
@@ -2,44 +2,69 @@ import LabeledInput, { LabeledInputProps } from '@atb/components/labled-input';
...
-import { ChangeEvent, useState } from 'react';
+import { ChangeEvent, useEffect, useState } from 'react';
 import style from './line-filter.module.css';
+import useSWR from 'swr';
 type LineFilterProps = {
   filterState: string[] | null;
   onChange: (lineFilter: string[] | null) => void;
 };
 
+const fetcher = (url: string) => fetch(url).then((res) => res.json());
```
adriansberg, line 13: Can be removed, I think.

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -784,3 +814,21 @@ async function getSortedViaTrips(
...
+      } else {
+        publicCodeLineList.set(line.publicCode, [line.id]);
+      }
+    }
+  });
+
+  // @ts-ignore
+  // @ts-expect-error: This is needed to allow iteration through Map objects
+  // @ts-nocheck
+  return JSON.stringify([...publicCodeLineList]);
```
adriansberg, line 833: Is it necessary to stringify this?

src/page-modules/assistant/line-filter/index.tsx
jonasbrunvoll: Mhm, the fetch operation does not work when I only remove the fetcher. Am I supposed to replace it with something?

src/page-modules/assistant/server/journey-planner/index.ts
jonasbrunvoll: Yes, I think it is necessary. As I understand it, Map objects are not directly serializable to the JSON format, as JSON only supports simple data types such as strings, numbers, booleans, arrays and objects. Therefore, the Map object must be converted to a serializable form such as a string before it is sent to the client. On the client side, the string is turned into the etmap object again.

mikaelbr: Does it need to be a map? Seems like this can just be a normal object without actually loosing any functionality.

src/page-modules/assistant/server/lines-cache.ts
```diff
@@ -0,0 +1,38 @@
+import TTLCache from '@isaacs/ttlcache';
```
mikaelbr, line 1: I think this entire file isn't needed. See comment on `fetchPolicy`

src/page-modules/assistant/server/journey-planner/index.ts
```diff
@@ -156,6 +166,26 @@ export function createJourneyApi(
       return nonTransits;
     },
 
+    async lines(input) {
+      const potential = getLinesIfCached(input as LineInput);
+      if (potential) return potential;
+
+      const result = await client.query<LinesQuery, LinesQueryVariables>({
+        query: LinesDocument,
+        variables: { authorities: input.authorities },
```
mikaelbr, line 175: This can be cached with using ApolloClient directly without having to do your own caching
```suggestion
        variables: { authorities: input.authorities },
        fetchPolicy: 'cache-first'
```

src/page-modules/assistant/server/journey-planner/index.ts
mikaelbr, line 833: I think this will be simplified by using just something like:
```suggestion
function createPublicCodeLineMap(lines: LineFragment[]) {
  const publicCodeLineList: Record<string, string[]> = {};
  for (const line of lines) {
    if (!line.publicCode) continue;
    if (!publicCodeLineList[line.publicCode]) {
      publicCodeLineList[line.publicCode] = [];
    }

    publicCodeLineList[line.publicCode].push(line.id);
  }
  return publicCodeLineList;
```

src/page-modules/assistant/server/journey-planner/validators.ts
```diff
@@ -146,6 +146,10 @@ export const tripPatternWithDetailsSchema = z.object({
   ),
 });
 
+export const linesSchema = z.object({
+  publicCodeLineMapString: z.string(),
```
mikaelbr, line 150: Given previous suggestion this needs to be updated to reflect actual type
```suggestion
  publicCodeLineMapString: z.record(z.string(), z.array(z.string())),
```

src/page-modules/assistant/line-filter/index.tsx
```diff
@@ -2,44 +2,69 @@ import LabeledInput, { LabeledInputProps } from '@atb/components/labled-input';
...
 
-      onChange(linesWithPrefix);
+      onChange(lines);
     }
   };
 
+  useEffect(() => {
+    if (!data) return;
+    setPublicCodeLineMap(new Map(JSON.parse(data.publicCodeLineMapString)));
+  }, [data]);
```
mikaelbr, line 57: If you don't use `Map` this isn't needed and you can just use data directly

src/page-modules/assistant/line-filter/index.tsx
```diff
@@ -2,44 +2,69 @@ import LabeledInput, { LabeledInputProps } from '@atb/components/labled-input';
...
+        .flat();
 
-      onChange(linesWithPrefix);
+      onChange(lines);
     }
   };
 
+  useEffect(() => {
+    if (!data) return;
+    setPublicCodeLineMap(new Map(JSON.parse(data.publicCodeLineMapString)));
```
adriansberg, line 56: Isn't the data we get from the API endpoint already on the format we are wanting to use, and therefore can be used just as a lookup without having to set a local state?

src/page-modules/assistant/line-filter/index.tsx
```diff
@@ -2,44 +2,69 @@ import LabeledInput, { LabeledInputProps } from '@atb/components/labled-input';
...
+  }, [data]);
+
+  useEffect(() => {
+    if (!publicCodeLineMap || localFilterState !== '') return;
+    const publicCodeString = Array.from(publicCodeLineMap.entries())
+      .filter(([_, line]) => filterState?.some((code) => line.includes(code)))
+      .map(([code]) => code)
+      .join(', ');
+    setLocalFilterState(publicCodeString);
+  }, [publicCodeLineMap, localFilterState, filterState]);
```
mikaelbr, line 66: I'm not sure this is the most robust way of solving this. 🤔 Think we need to do some testing, as I was experiencing that when removing filters they popped back in (impossible to remove)

src/page-modules/assistant/line-filter/index.tsx
adriansberg: Disregard this comment, Mikael's appeared when I did a reload.

mikaelbr: yep. But we should pack the useSwr in custom hook to do proper type mapping

adriansberg: See `useMapTariffZones` or `useNonTransitTrip` for examples.

src/page-modules/assistant/line-filter/index.tsx
adriansberg: See `useMapTariffZones` or `useNonTransitTrip` for examples.

adriansberg:
Let's have a huddle about this tomorrow, easier to point and explain various ideas instead of having several comments trying to convey the same information but resulting in potential misunderstandings. 

I put together some of the thoughts in this commit, so we can have a look at that together: https://github.com/AtB-AS/planner-web/commit/c24f555389ddb9433c692ced7d00a4a3a943b9ca

mikaelbr:
I think a very quick win would be to use https://www.downshift-js.com/use-multiple-selection here.

jonasbrunvoll:
Expanding filter search to use useMultipleSelection can perhaps be a separate task? This PR is getting pretty big already.

adriansberg (approved):
LGTM! Tested and works as expected, adding lines when they are entered and removing them when removed, after the debounce set to 750ms passes. 

I reckon that this feature will be worked on to possibly implement the multiple selection that we talked about. Would be nice to have some tests for this then.

PR #263, opened 2024-03-14
fix: set filter correctly when navigating back from details page

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/17194

### Background

When going into a details page and navigating back the filter for transport mode is not included in the query, causing the filter to be reset to default.

#### Illustrations
The recording displays how the same filters are active before and after navigating to the details page. This can be seen in  both the filter buttons and URL parameters.  

<details>

https://github.com/AtB-AS/planner-web/assets/59939294/2ccac1f5-bc8e-4f9c-bdf9-cd16b6798bb5

</details>

### Proposed solution

Including a string made up by the selected transport mode filters to generate the singleTripQueryString. Thus, the same string is available when the singleTripQueryString is decompressed, making it possible to set the parameters again when navigating back from details page.

### Acceptance Criteria

- [ ] The set transport mode filter remains the same when going to details page and navigating back.
- [ ] Going to details page from copied url and navigating back keeps the transport mode filter.

### DIFF
--- a/src/page-modules/assistant/details/index.tsx
+++ b/src/page-modules/assistant/details/index.tsx
@@ -28,8 +28,12 @@ export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
   );
 
   const tripSearchParams = router.query.id
-    ? tripQueryStringToQueryParams(router.query)
+    ? tripQueryStringToQueryParams(String(router.query.id))
     : undefined;
+
+  if (tripSearchParams && router.query.filter) {
+    tripSearchParams.append('filter', router.query.filter as string);
+  }
 
   return (
     <div className={style.container}>
--- a/src/page-modules/assistant/details/index.tsx
+++ b/src/page-modules/assistant/details/index.tsx
@@ -28,7 +28,7 @@ export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
   );
 
   const tripSearchParams = router.query.id
-    ? tripQueryStringToQueryParams(String(router.query.id))
+    ? tripQueryStringToQueryParams(router.query)
     : undefined;
 
   return (
--- a/src/page-modules/assistant/trip/trip-pattern/index.tsx
+++ b/src/page-modules/assistant/trip/trip-pattern/index.tsx
@@ -72,7 +72,7 @@ export default function TripPattern({
 
   return (
     <motion.a
-      href={`/assistant/${tripPattern.compressedQuery}___${router.query.filter}`}
+      href={`/assistant/${tripPattern.compressedQuery}&filter=${router.query.filter}`}
       className={className}
       data-testid={testId}
       initial={{ opacity: 0, x: -10 }}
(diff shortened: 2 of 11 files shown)


mortennordseth:
Notater fra felles gjennomgang av PR:
- Legge til filter som query param i stedet for å utvide compressed query

jonasbrunvoll:
I was unable to navigate to the details page when `/` was used as a delimiter. I therefore chose to use `___` instead.

mortennordseth:
> I was unable to navigate to the details page when / was used as a delimiter. I therefore chose to use ___ instead.


I think you should use query string instead (https://en.wikipedia.org/wiki/Query_string). For example `?filter=bus` or `?filter=bus,rail`. This is in line with what is done for the /assistant page as well

src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -11,6 +11,7 @@ import { MonoIcon } from '@atb/components/icon';
 import { Typo } from '@atb/components/typography';
 import { TransportIconWithLabel } from '@atb/modules/transport-mode';
 import { andIf } from '@atb/utils/css';
+import { useRouter } from 'next/router';
```
mortennordseth, line 14: I can't see any use of the router here. Did you forget to remove the import?

src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -70,7 +72,7 @@ export default function TripPattern({
 
   return (
     <motion.a
-      href={`/assistant/${tripPattern.compressedQuery}`}
+      href={`/assistant/${tripPattern.compressedQuery}&filter=${router.query.filter}`}
```
mortennordseth, line 75:
```suggestion
      href={`/assistant/${tripPattern.compressedQuery}?filter=${router.query.filter}`}
```

src/page-modules/assistant/details/index.tsx
```diff
@@ -28,7 +28,7 @@ export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
   );
 
   const tripSearchParams = router.query.id
-    ? tripQueryStringToQueryParams(String(router.query.id))
+    ? tripQueryStringToQueryParams(router.query)
     : undefined;
```
mortennordseth, line 32: You might disagree with me, but I don't think we should modify `tripQueryStringToQueryParams` without changing the `decompressedQueryString`. What about doing something like this? 

```tsx
  const tripSearchParams = router.query.id
    ? tripQueryStringToQueryParams(String(router.query.id))
    : undefined;

  if (tripSearchParams && router.query.filter) {
    tripSearchParams.append('filter', router.query.filter as string);
  }

```
This only requires us to add the filter as a query param to the href in `/trip-pattern/index.tsx` (already done) and to use it when redirecting back to the travel suggestions (the suggested code above)

mortennordseth (approved):
Nice

PR #264, opened 2024-03-18
feat: support for flexible transport

mortennordseth:
image

image

Closes https://github.com/AtB-AS/kundevendt/issues/16685


src/modules/flexible/utils.ts
```diff
@@ -0,0 +1,173 @@
...
+  bookingArrangements: BookingArrangement | null,
+  aimedStartTime: string,
+  flex_booking_number_of_days_available?: number,
+): BookingStatus => {
+  if (!bookingArrangements) return 'none';
+
+  const secondsToDeadline = getSecondsRemainingToBookingDeadline(
+    bookingArrangements,
+    aimedStartTime,
+    Date.now(),
```
jonasbrunvoll, line 32: To account for time zone differences:
```suggestion
    setTimezone(Date.now()),
```

jonasbrunvoll:
Have this been feature been tested with different timezone? I see that there are a couple places when new Date() objects are created (f. ex  in BookingSection). Might it be necessary to take into account time zone differences as there is a deadline for booking here?

src/modules/flexible/utils.ts
mortennordseth: Entur returns UTC and `Date.now()` is UTC so this should be correct 🤔 I'll do some testing

jonasbrunvoll (approved):
Looks good🙌

src/modules/flexible/utils.ts
```diff
@@ -0,0 +1,173 @@
...
+export const isLineFlexibleTransport = (
+  line?:
+    | TripPatternWithDetails['legs'][0]['line']
+    | TripPattern['legs'][0]['line'],
+) => !!line?.flexibleLineType;
+
+export const getBookingStatus = (
+  bookingArrangements: BookingArrangement | null,
+  aimedStartTime: string,
+  flex_booking_number_of_days_available?: number,
```
adriansberg, line 25: Is there a reason that this can't be camel cased?

mortennordseth: Fixed 👍

PR #268, opened 2024-03-19
feat: add sitemap to travel planner

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/17182

### Background

UUtilsynet requires that we have at least two ways to navigate between pages on the website. Ref: https://www.uutilsynet.no/wcag-standarden/245-flere-mater-niva-aa/107

We don't have multiple links to the assistant and departures page. An alternative is a search function, but I believe adding a sitemap is sufficient.

#### Illustrations
<details>
<summary>screenshot</summary>

Skjermbilde 2024-03-19 kl  12 42 01

</details>

### Proposed solution

Add sitemap displaying urls and when the date when they were last modified.


src/pages/sitemap.xml.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <lastmod>${site.dateModified}</lastmod>
+    </url>
+    `;
+  })}
+   </urlset>
+ `;
+};
+
+export const getServerSideProps: GetServerSideProps<{}> = async (ctx) => {
+  const siteMap = generateSitemap('src/pages');
```
jonasbrunvoll, line 62:
```suggestion
  const sitemap = generateSitemap('src/pages');
```

src/pages/sitemap.xml.tsx
```diff
@@ -0,0 +1,72 @@
...
+  ctx.res.setHeader('Content-Type', 'text/xml');
+  ctx.res.write(siteMap);
```
jonasbrunvoll, line 64:
```suggestion
  ctx.res.write(sitemap);
```

mortennordseth (commented):
I like this approach, but it seems like `sites` is empty: 

image

adriansberg:
2024-03-19 at 20 14 01

The string replacement does not take `www` into account. I see some of the orgs have it, and some don't.

.gitignore
```diff
@@ -49,3 +49,7 @@ src/components/icon/generated-icons.ts
 /playwright-report/
 /blob-report/
 /playwright/.cache/
+
+
+/public/robots.txt
+/public/sitemap.xml
```
mortennordseth, line 55: Why are these ignored?

next-sitemap.js
```diff
@@ -0,0 +1,26 @@
+/** @type {import('next-sitemap').IConfig} */
+
+function getHomepageUrl() {
+  const orgId = process.env.NEXT_PUBLIC_PLANNER_ORG_ID;
+  const org = require(`./orgs/${orgId}.json`);
+  return org.urls.homePageUrl.href;
+}
```
mortennordseth, line 7: I don't think that we should use the home page URL to define hostname of the planner web. We don't know that `reise.<homePageUrl>` is used by everyone. If it is needed I think we should add a new property, e.g. `hostname`, to the org file.

.gitignore
jonasbrunvoll: As i understood it, these files are automatically be created postbuild, and are therefore not necessary to push.

next-sitemap.js
jonasbrunvoll: Agreed, sounds like a more robust approach.

public/robots.txt
```diff
@@ -0,0 +1,9 @@
+# *
```
mikaelbr, line 1: Should these files be ignored?

jonasbrunvoll: As I understood it, site.xml and robots.txt are automatically created postbuild, and can therefore be added to .gitignore.. But, as I'm not sure how to set the siteUrl in staging and preview, As I could´t see how sitemap.xml and robots.xml looked in preview, I removed the files from .gitignore to including the files in the public folder. 

From preview, it looks like it works as it should, so I think I can add add the files to .gitignore again.

public/sitemap.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
jonasbrunvoll: As I understood it, site.xml and robots.txt are automatically created postbuild, and can therefore be added to .gitignore.. But, as I'm not sure how to set the siteUrl in staging and preview, As I could´t see how sitemap.xml and robots.xml looked in preview, I removed the files from .gitignore to including the files in the public folder. 

From preview, it looks like it works as it should, so I think I can add add the files to .gitignore again.

.github/workflows/docker.yml
```diff
@@ -19,27 +19,31 @@ jobs:
         ATB_NEXT_PUBLIC_FIREBASE_AUTH_DOMAIN=${{ github.event_name == 'release' && 'atb-mobility-platform.firebaseapp.com' || 'atb-mobility-platform-staging.firebaseapp.com' }}
         ATB_NEXT_PUBLIC_FIREBASE_APP_ID=${{ github.event_name == 'release' && '1:827196677776:web:c4b1d8924bb5778bf50215' || '1:939812594010:web:0308b2a9cdc80b0d069363' }}
         ATB_NEXT_PUBLIC_FIREBASE_MEASUREMENT_ID=${{ github.event_name == 'release' && '' || '' }}
+        ATB_NEXT_PUBLIC_IN_STAGING=${{ github.event_name == 'atb-mobility-platform-staging'}}
```
mortennordseth, line 22: I think it would be better if this env variable said which environment it is, rather than a boolean for being staging or not. Something like `'dev' | 'staging' | 'prod'`. 

Also, the environment should be the same for everyone, so do we need to prefix these and have multiple ones?

mortennordseth: I think this should do the trick

`NEXT_PUBLIC_ENVIRONMENT=${{ github.event_name == 'release' && 'prod' || 'staging' }}`

jonasbrunvoll: I´ll update the logic.

mikaelbr:
@jonasbrunvoll Think it is good to include a text on this and how it works in the README. meaning that `yarn build` now essentially requires `NEXT_PUBLIC_PLANNER_ORG_ID` and `NEXT_PUBLIC_ENVIRONMENT`. So just note on how to test locally, required envs. Doesn't need to be lots of text

mikaelbr:
Seems next-sitemap ignores dynamic routes (like `[[...id]].tsx` which we uses for departures). So need to add that route explicitly. Sending suggestion in comment

next-sitemap.js
```diff
@@ -0,0 +1,20 @@
...
+
+module.exports = {
+  siteUrl: environment ? environmentUrls[environment] : environmentUrls['dev'],
+  generateIndexSitemap: false,
+  sitemapSize: 7000,
+  generateRobotsTxt: true,
+  robosTxtOptions: {
+    policies: [{ userAgent: '*', allow: '/' }],
+  },
+};
```
mikaelbr, line 20:
```suggestion
/** @type {import('next-sitemap').IConfig} */
module.exports = {
  siteUrl: environment ? environmentUrls[environment] : environmentUrls['dev'],
  generateIndexSitemap: false,
  sitemapSize: 7000,
  generateRobotsTxt: true,
  robosTxtOptions: {
    policies: [{ userAgent: '*', allow: '/' }],
  },

  // Adds path as it doesn't support dynamic routes.
  additionalPaths(config) {
    const result = [];
    // @TODO Consider if we should prepopulate this for better SEO for
    // quays.
    result.push({
      loc: '/departures',
      changefreq: 'yearly',
      priority: 0.7,
      lastmod: new Date().toISOString(),
    });
    return result;
  },
};
```

mikaelbr (approved):
With README changes I think this is good!

PR #270, opened 2024-03-22
feat: extend line filter to include national organizations

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/17261

### Background

As of today, the travel planner can only filter lines that are planned by the relevant organization. Better explained, this means that the travel planner of FRAM can only filter on journeys planned by FRAM, while the travel planner of NFK can only filter on journeys planned by NFK.

But, there are other national organizations that organize trips across counties and regions. Example: Vy, Sj Nord, Boreal and Fjord 1. The line search should be extended to be able to filter on such organizations.

#### Illustrations
<details>

Skjermbilde 2024-03-22 kl  15 11 26

Skjermbilde 2024-03-22 kl  15 11 48

</details>

### Proposed solution
- [x]  Find a different pattern to check whether a line number is valid or not.  
- [x] Identify which organizations that should always be possible to preform line filter on.  
- [x] Update the endpoint `lines()` in `src/pages/api/assistant/lines/[authorityId].ts` to include all the  authorityIds as params.

### Acceptance Criteria
- [x] Can filter on lines containing characters other than numbers.
- [x] Make sure that the user is able to provide line number in both upper end lower case when filtering.  
- [x] Able to filter lines organized by the relevant organization.
- [x] Able to filter lines organized by the national organizations.

### DIFF
--- a/src/page-modules/assistant/line-filter/index.tsx
+++ b/src/page-modules/assistant/line-filter/index.tsx
@@ -11,6 +11,8 @@ type LineFilterProps = {
   onChange: (lineFilter: string[] | null) => void;
 };
 
+const regexValidChars = /^$|^[0-9a-zA-Z\s-]+$/;
+
 export default function LineFilter({ filterState, onChange }: LineFilterProps) {
   const { t } = useTranslation();
 
@@ -38,6 +40,7 @@ export default function LineFilter({ filterState, onChange }: LineFilterProps) {
       onChange(null);
     } else {
       const lines = input
+        .toUpperCase()
         .split(',')
         .flatMap((line) => data[line.trim()])
         .filter(Boolean);
@@ -91,6 +94,6 @@ export default function LineFilter({ filterState, onChange }: LineFilterProps) {
 
 function isValidFilter(filter: string): boolean {
   return filter.split(',').every((line) => {
-    return !isNaN(Number(line.trim()));
+    return regexValidChars.test(line);
   });
 }
--- a/src/page-modules/assistant/line-filter/index.tsx
+++ b/src/page-modules/assistant/line-filter/index.tsx
@@ -23,8 +23,6 @@ export default function LineFilter({ filterState, onChange }: LineFilterProps) {
   const [validationError, setValidationError] =
     useState<LabeledInputProps['validationError']>();
   const [localFilterState, setLocalFilterState] = useState('');
-  const [publicCodeLineMap, setPublicCodeLineMap] =
-    useState<Map<string, string[]>>();
 
   const onChangeWrapper = (event: ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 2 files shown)


jonasbrunvoll:
Come to think about it now. Are there any disadvantages or reason for why we should fetch all liners across all authorities? 
If not, I could update the lines() endpoint to retrieve all lines, thus not not having to specify authorities. This way enables users to filter trips consisting of legs organized from different authorities. For example, the trip Trondheim - Molde, consisting of express boat (800) from Trondheim to Kristiansund (Atb), and bus 100 from Kristiansund to Molde (FRAM).

mortennordseth:
> Come to think about it now. Are there any disadvantages or reason for why we should fetch all liners across all authorities? If not, I could update the lines() endpoint to retrieve all lines, thus not not having to specify authorities. This way enables users to filter trips consisting of legs organized from different authorities. For example, the trip Trondheim - Molde, consisting of express boat (800) from Trondheim to Kristiansund (Atb), and bus 100 from Kristiansund to Molde (FRAM).

A few possible disadvantages I can think of
- Data size (slow)
- Authorities use the same line numbers, so how do we decide which one to show or select? This could possibly be solved by having a drop-down where the items include line number and authority name

However, I think having the option to search for line numbers from different authorities makes sense and is something we should look into/discuss how to solve

jonasbrunvoll:
> > Come to think about it now. Are there any disadvantages or reason for why we should fetch all liners across all authorities? If not, I could update the lines() endpoint to retrieve all lines, thus not not having to specify authorities. This way enables users to filter trips consisting of legs organized from different authorities. For example, the trip Trondheim - Molde, consisting of express boat (800) from Trondheim to Kristiansund (Atb), and bus 100 from Kristiansund to Molde (FRAM).
> 
> A few possible disadvantages I can think of
> 
> * Data size (slow)
> * Authorities use the same line numbers, so how do we decide which one to show or select? This could possibly be solved by having a drop-down where the items include line number and authority name
> 
> However, I think having the option to search for line numbers from different authorities makes sense and is something we should look into/discuss how to solve

Yes, the size of the return data is a disadvantage and could possibly slow the performance of the trip planner. Line filter will still work if  authorities uses the same line numbers, as the lines are added to a whitelist when specified in line filter. F. example, if line "1" is specified, all trips across all authorities with public code "1" are included to the whitelist.

mikaelbr:
I think this can be too implicit for the end user and there are some technical challenges (unnecessary large transfer payloads). So my suggestion is to temporary let this be and add an issue to solve this more conceptually for planner-web v1.1

mikaelbr:
Closed in favour of extending features to be more intuitive for the user: https://github.com/AtB-AS/kundevendt/issues/17726

PR #272, opened 2024-04-02
feat: enable line filter for nfk

jonasbrunvoll:
### Background

Line filter does not appear in nfk staging. 

### Proposed solution

Enable line filter for nfk by updating the docker build script.
 
### Acceptance Criteria

- [ ] Line filter appears in nfk staging.


.github/workflows/docker.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         NFK_NEXT_PUBLIC_FIREBASE_AUTH_DOMAIN=${{ github.event_name == 'release' && 'nfk-prod.firebaseapp.com' || 'nfk-staging.firebaseapp.com' }}
         NFK_NEXT_PUBLIC_FIREBASE_APP_ID=${{ github.event_name == 'release' && '1:638315587956:web:51e95a11a03f27dc9ca421' || '1:793301954236:web:908723b8ded2824cadeab9' }}
         NFK_NEXT_PUBLIC_FIREBASE_MEASUREMENT_ID=${{ github.event_name == 'release' && '' || '' }}
-        NFK_NEXT_PUBLIC_DISABLE_LINE_FILTER=true
+        NFK_NEXT_PUBLIC_DISABLE_LINE_FILTER=false
```
mortennordseth, line 30: This was added as a temporary solution to disable the line filter for NFK. Now that we have a working solution for all organizations I think we should remove this env-variable and its use instead

mortennordseth (changes requested):

mortennordseth (approved):

PR #273, opened 2024-04-02
fix: adds better screen reader support for loading / results states

mikaelbr:
Fixes https://github.com/AtB-AS/kundevendt/issues/17360


mortennordseth (approved):
Seems to work as intended :) Failing tests seems to be unrelated to these changes

PR #282, opened 2024-04-08
fix: permanent solution to take into account winter and summer time.

jonasbrunvoll:


mikaelbr:
Is it possible to test this in some way? Extend the tests that is

jonasbrunvoll:
> Is it possible to test this in some way? Extend the tests that is

Have now added tests to check whether the adjustment for timezone differences are correct. The tests checks if the departure times when selecting "now" and when selecting "departure by", are within the same timezone (true). This works as trips fetched using departure "now" always display the departures times in CET. 

I created a couple of help functions to help with the testing. I was unsure of where to add them, so for now they are in the same file as the tests.

mikaelbr:
Come to think of this also, I think this is something that is very handy to test in unit tests also. 🤔

src/page-modules/assistant/__tests__/assistant.test.tsx
```diff
@@ -33,6 +33,11 @@ import {
   getAssistantTripIfCached,
 } from '../server/trip-cache';
 import { FromToTripQuery } from '../types';
+import {
+  getHoursDifferenceFromCET,
+  getLastSundayOfMonthAndSetTime,
+  isDaylightSavingTime,
+} from '@atb/utils/date';
```
mikaelbr, line 41: Not to be difficult but these tests should be located in `utils/date` directory 🤔

jonasbrunvoll: I suspected that these tests might be moved😅

jonasbrunvoll: Should also move date.ts into `utils/date`? In that case, quite a few import statements to retrieve functions from date.ts will change, but that's the structure of the repo, right?

mikaelbr: I didn't check where date.ts was located. But date.ts can remain inside `utils` but you can create `utils/__tests__/date.test.ts` and move the tests over there. The point is to co-locate tests with the implementation.

mikaelbr (approved):

PR #284, opened 2024-04-09
feat: expand the search time selector to enable retrieval of search results from yesterday.

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/17561

### Background

Match the behavior of the app with the departure filter time. Should be possible to see results back in time (two days). This changes from existing behavior where it is only allowed for 1 day.

#### Illustrations
Screenshots were the 9th of April.

<details>

Skjermbilde 2024-04-09 kl  14 59 56
Skjermbilde 2024-04-09 kl  15 00 26

</details>

### Proposed solution

Expand the search time selector to enable retrieval of search results from yesterday.

### Acceptance Criteria

- [x] The user is able to search after trips that had departure time yesterday.

### DIFF
--- a/src/modules/search-time/selector/index.tsx
+++ b/src/modules/search-time/selector/index.tsx
@@ -16,6 +16,10 @@ type SearchTimeSelectorProps = {
   options?: SearchMode[];
 };
 
+const today = new Date();
+today.setDate(today.getDate() - 1);
+const yesterdayString = today.toISOString().split('T')[0];
+
 export default function SearchTimeSelector({
   onChange,
   initialState = { mode: 'now' },
@@ -72,7 +76,6 @@ export default function SearchTimeSelector({
   };
 
   const isPastDate = (selectedDate: string) => {
-    const today = new Date().toISOString().split('T')[0];
     const year = initialDate.getFullYear().toString();
     const month = initialDate.getMonth().toString().padStart(2, '0');
     const day = initialDate.getDate().toString().padStart(2, '0');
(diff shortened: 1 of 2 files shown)


src/modules/search-time/selector/index.tsx
```diff
@@ -83,8 +84,8 @@ export default function SearchTimeSelector({
       resetToCurrentDate();
 
     // To ensure that the time is not past whenever reselecting back the current date.
-    if (selectedDate <= today) resetToCurrentTime();
-    return selectedDate < today;
+    if (selectedDate <= yesterday) resetToCurrentTime();
+    return selectedDate < yesterday;
   };
```
mortennordseth, line 89: I guess this kind of belongs the original PR for this function, but this function seems a bit odd to me. It returns a boolean for whether the date is in the past or not, but also calls the `resetToCurrentTime`. I think it would make more sense if this function returns whether the date is in the past, or with this change, if it is past yesterday, and if so the calling function calls `resetToCurrentTime`.

src/modules/search-time/selector/index.tsx
```diff
@@ -16,6 +16,8 @@ type SearchTimeSelectorProps = {
   options?: SearchMode[];
 };
 
+const yesterday = format(subDays(new Date(), 1), 'yyyy-MM-dd');
+
```
mortennordseth, line 20: I think this variable should be declared inside the `SearchTimeSelector`. Let me know if there is a reason why it should not and the reason for being created here 👍

mortennordseth (approved):

PR #287, opened 2024-04-18
feat: improves link information for departures with Open Graph (and minor SEO improvements)

mikaelbr:
- Sets proper title
- Adds Open Graph metadata
- Adds generated open graph image for departure stop place
- Adds better meta description
- Adds all departures to the sitemap to be able to crawl all pages by search engines.

This adds a large file checked into the repo, but I think as this is very static data that rarely changes it is better to have it checked in rather than generating it build time. We could move it outside as a separate package, but that feels excessive at this point. In any case... Lots of lines changed on this one.

Fixes https://github.com/AtB-AS/kundevendt/issues/17688

Example generated image for Open Graph / Social platforms:

image


mortennordseth (approved):
These 753,426 new lines LGTM :)

PR #294, opened 2024-05-13
feat: add note informing users about having to buy external ticket

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/18065

### Background
As mentioned in the conversation, the same note informing the users out the need to buy tickets from other sources than in the app, should also be shown in the planner web.

> The booking phone number and/or URL is only shown if the booking status is "bookable", see https://github.com/AtB-AS/planner-web/blob/main/src/page-modules/assistant/details/trip-section/booking-section.tsx#L57. 
> 
> This is the same as what the app does. However, the app also has a note saying "This journey requires a ticket that is not available from this app, or must be purchased from a provider other than AtB." This doesn't answer all of the questions a customer might have, but it's some more information. A better approach could be to show the booking URL / phone number at all times, but if so we should probably update the app as well.
> 
> I suggest that we add the note from the app to the planner-web and then discuss with the designers at AtB how we can improve this in both the app and planner-web. 
> 
> <details>
> <summary>Screenshot of the OMS-app</summary>
> 
> Image
> 
> </details>
>   

 _Originally posted by @mortennordseth in [#17560](https://github.com/AtB-AS/kundevendt/issues/17560#issuecomment-2044238052)_

#### Illustrations
<details>

Image

</details>

### Proposed solution
Display the same message in planner web as already shown in the app.
 
### Acceptance Criteria
- [ ] The info box should appear the same way in panner-web as in the app.

### DIFF
--- a/src/page-modules/assistant/details/index.tsx
+++ b/src/page-modules/assistant/details/index.tsx
@@ -12,6 +12,9 @@ import { getInterchangeDetails } from './trip-section/interchange-section';
 import { getLegWaitDetails } from './trip-section/wait-section';
 import { useRouter } from 'next/router';
 import { tripQueryStringToQueryParams } from './utils';
+import { MessageBox } from '@atb/components/message-box';
+import { getBookingStatus } from '@atb/modules/flexible/utils';
+import { getOrgData } from '@atb/modules/org-data';
 
 export type AssistantDetailsProps = {
   tripPattern: TripPatternWithDetails;
@@ -34,6 +37,12 @@ export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
   if (tripSearchParams && router.query.filter) {
     tripSearchParams.append('filter', router.query.filter as string);
   }
+
+  const requireTicketBooking = tripPattern.legs.some(
+    (leg) =>
+      getBookingStatus(leg.bookingArrangements, leg.aimedStartTime, 7) !==
+      'none',
+  );
 
   return (
     <div className={style.container}>
@@ -72,6 +81,12 @@ export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
         </div>
       </div>
       <div className={style.tripContainer}>
+        {requireTicketBooking && (
+          <MessageBox
+            type="info"
+            message={t(PageText.Assistant.ticketBooking.globalMessage)}
+          />
+        )}
         {tripPattern.legs.map((leg, index) => (
           <TripSection
             key={index}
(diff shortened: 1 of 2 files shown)


src/translations/pages/assistant.ts
```diff
@@ -70,6 +70,13 @@ const AssistantInternal = {
       },
     },
   },
+  ticketBooking: {
+    globalMessage: _(
+      'Reisen krever billett som må kjøpes fra et annet selskap enn AtB.',
+      'This journey requires a ticket that must be purchased from provider other than AtB.',
```
adriansberg, line 76: Should there be an "a" before "provider"? I.e. 'This journey requires a ticket that must be purchased from a provider other than AtB.'

src/page-modules/assistant/details/index.tsx
```diff
@@ -35,6 +37,12 @@ export function AssistantDetails({ tripPattern }: AssistantDetailsProps) {
     tripSearchParams.append('filter', router.query.filter as string);
   }
 
+  const requireTicketBooking = tripPattern.legs.some(
```
adriansberg, line 40: Is the same done in the app to know if booking is required?

src/translations/pages/assistant.ts
jonasbrunvoll: Fixed some minor errors now. The text should be the same as in the app.

src/translations/pages/assistant.ts
```diff
@@ -70,6 +70,13 @@ const AssistantInternal = {
       },
     },
   },
+  ticketBooking: {
```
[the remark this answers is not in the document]
jonasbrunvoll: Updatet the location👍

src/page-modules/assistant/details/index.tsx
jonasbrunvoll: In the app, there is a function to check whether a leg in the trip requires a ticket that can not be must be bought through the app. https://github.com/AtB-AS/mittatb-app/blob/eb680679c754e8e56848ffe7983779c0794468ba/src/travel-details-screens/components/Trip.tsx#L106-L109.

The solution I used in this PR is pretty much the same approach to determine whether the BookingSection should appear. https://github.com/AtB-AS/planner-web/blob/a4453ced8626eba2d075255fe06146e06e1424a8/src/page-modules/assistant/details/trip-section/index.tsx#L75-L79

adriansberg (approved):
LGTM!

PR #295, opened 2024-05-28
feat(widget): adds option to use inherited font in the widget

mikaelbr:
Added options to enforce use of inherited font for all elements in the widget. Will allow for parent container to dictate the use of font family. If not specified, the default behaviour is to use Roboto which is the default font for the hosted planner web solution (and webshop).

```ts
    const widget = window.PlannerWeb.createWidget({
      urlBase: 'https://reiseplanlegger.example.no/',
      language: 'nn', // supports 'nb', 'nn' and 'en'

      // Optional options
      outputOverrideOptions: {
        // Inherit font from page website.
        // By default it uses Roboto as the hosted planner web solution.
        inheritFont: false,
      },
    });
```

Fixes https://github.com/AtB-AS/kundevendt/issues/18111


jonasbrunvoll (approved):
🚀

PR #296, opened 2024-05-30
feat: extend widget with optional single column layout

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/18000

### Background

Add the ability to choose a single column type design (enforce mobile view) for the widget.

This is a need sent in by Troms.

#### Illustrations
<details>

Skjermbilde 2024-05-30 kl  13 08 17r
Skjermbilde 2024-05-30 kl  13 08 25

</details>

### Proposed solution

Include `singleColumnLayout` as an optional property to determine whether the widget should be shown in single column layout or not. By default set to false.  

### Acceptance Criteria

- [ ] By default, single column layout is set to false. 
- [ ] Should be possible to switch to single column layout when inspecting widget, by adding copy-pasting the css class `widget-module__singleColumnLayout` straight into the HTML with the element class widget-module__wrapper manually.

### DIFF
--- a/src/pages/widget/index.tsx
+++ b/src/pages/widget/index.tsx
@@ -204,8 +204,12 @@ function WidgetContent({
 
         <ul style={{ listStylePosition: 'inside', marginTop: '1rem' }}>
           <li>
-            <code>.widget-inheritFont</code>: Inherit font family from the
-            website
+            <code>.widget-module__inheritFont</code>: Inherit font family from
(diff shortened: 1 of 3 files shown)


mortennordseth:
I think we should document the single-column option here as well 👇 The rest looks good! 

image

mortennordseth (approved):
NIce

PR #299, opened 2024-06-05
fix: fix server side redirect error

adriansberg:
Removing the `useEffect` caused the code to run on the server side, causing an error by `next/router` which can only be run on the client side.

### DIFF
--- a/src/page-modules/assistant/layout.tsx
+++ b/src/page-modules/assistant/layout.tsx
@@ -16,7 +16,12 @@ import { PageText, useTranslation } from '@atb/translations';
 import { FocusScope } from '@react-aria/focus';
 import { AnimatePresence, motion } from 'framer-motion';
 import { useRouter } from 'next/router';
-import { FormEventHandler, PropsWithChildren, useState } from 'react';
+import {
+  FormEventHandler,
+  PropsWithChildren,
+  useEffect,
+  useState,
+} from 'react';
 import style from './assistant.module.css';
 import { FromToTripQuery } from './types';
 import { createTripQuery, setTransportModeFilters } from './utils';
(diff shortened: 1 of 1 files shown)


jonasbrunvoll (approved):
🚀

src/page-modules/assistant/layout.tsx
```diff
@@ -117,8 +122,14 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
...
 
-  if (tripQuery.transportModeFilter === null)
-    onTransportFilterChanged(setTransportModeFilters(transportModeFilter));
+  useEffect(() => {
+    if (tripQuery.transportModeFilter === null)
+      onTransportFilterChanged(setTransportModeFilters(transportModeFilter));
+  }, [
+    onTransportFilterChanged,
+    transportModeFilter,
+    tripQuery.transportModeFilter,
```
mikaelbr, line 131: We are sure that `tripQuery.transportModeFilter` is stable as to not change in reference every time? Won't have actual implications here other than doing unnecessary work.

adriansberg: Not sure. Can have just the transportModeFilter as a dependency and ignore the eslint warning?

mikaelbr: could technically have `tripQuery.transportModeFilter == null` as a dependency 🤔 Or stored as an variable before

adriansberg: Never used assertions in a dependency array 😲 I can try that.

adriansberg: Fine to have `onTransportFilterChanged` in the dep array here?

mikaelbr: Need to check if the effect triggers needlessly

adriansberg: Triggers twice on page load at the moment.

adriansberg: @jonasbrunvoll, could you check that the effect triggers when it needs to and that it doesn't trigger needlessly, according to when the transport filter should be updated?

jonasbrunvoll: @adriansberg, I checked in networks. To me it looks like useEffect is triggered once and only when there is a change in filter. I also see that queries fetching the icon SVGs are run every time the button "More choices / Less choices" is pressed, but this happens both in `adriansberg/fix-redirect-error` and in `main`. That could perhaps be improved, but that is probably a separate task of its own.

PR #301, opened 2024-06-05
refactor: parse global messages using zod

mortennordseth:
With this PR we start using zod for parsing global messages. I've also updated the `startDate` and `endDate` type to be Date instead of Firestore Timestamp. 

Part of solving https://github.com/AtB-AS/kundevendt/issues/17725

### DIFF
--- a/src/modules/global-messages/converters.ts
+++ b/src/modules/global-messages/converters.ts
@@ -17,7 +17,7 @@ export const globalMessageConverter = {
       type: data.type,
       subtle: data.subtle,
       context: data.context,
-      isDismissable: data.isDismissable ?? false,
+      isDismissable: data.isDismissable,
       startDate: firestoreTimestampToDate(data.startDate),
       endDate: firestoreTimestampToDate(data.endDate),
     };
--- a/src/modules/global-messages/converters.ts
+++ b/src/modules/global-messages/converters.ts
@@ -1,50 +1,5 @@
-import { isArray } from 'lodash';
-import { GlobalMessageContextEnum, GlobalMessageType } from './types';
-import { MessageMode } from '@atb/components/message-box';
-import { QueryDocumentSnapshot } from 'firebase/firestore';
-import { LocalizedString } from '@atb/translations/commons';
-
-export function mapToLocalizedStringArray(
-  data: any,
-): LocalizedString[] | undefined {
-  if (!data) return;
-  if (!isArray(data)) return;
-
-  return data
-    .map((ls: any) => mapToLocalizedString(ls))
-    .filter((lv): lv is LocalizedString => !!lv);
-}
-
-function mapToLocalizedString(data: any): LocalizedString | undefined {
-  if (!data) return;
-  if (data.lang != 'nob' && data.lang != 'eng' && data.lang != 'nno') return;
-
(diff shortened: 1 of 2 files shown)


jonasbrunvoll (approved):
Looks good to me👍

src/modules/global-messages/converters.ts
```diff
@@ -54,40 +9,31 @@ export const globalMessageConverter = {
     snapshot: QueryDocumentSnapshot<GlobalMessageType>,
   ): GlobalMessageType | undefined {
     const data = snapshot.data();
+    const potential = {
+      id: snapshot.id,
+      active: data.active,
+      title: data.title,
+      body: data.body,
```
adriansberg, line 16: The `mapToLocalizedString` is redundant with the zod parsing? Haven't looked too much into what that function did.

mortennordseth: We could be more strict with a function like that, e.g. saying that a body requires both bokmål and English translations, but currently, with the current zod implementation, we only require that the body has at least one translation in the format of

```json
{ lang: string, value: string }
```
or 
```json
{ language: string, value: string }
```

src/modules/global-messages/converters.ts
```diff
@@ -54,40 +9,31 @@ export const globalMessageConverter = {
...
 };
 
-function hasNecessaryGlobalMessageTypeFields(data: any) {
-  return (
-    'active' in data && 'body' in data && 'context' in data && 'type' in data
-  );
+function firestoreTimestampToDate(timestamp: any) {
+  if (!timestamp) return;
+  if (!(timestamp instanceof Timestamp)) return;
+  return timestamp.toDate();
```
mikaelbr, line 38: Maybe not cleaner, but for the record you could have done this

```suggestion
  return timestamp?.toDate?.();
```

mortennordseth: Neat! I didn't know one could use optional chaining for function calls 👀

src/modules/global-messages/types.ts
```diff
@@ -1,20 +1,28 @@
...
-  startDate?: Timestamp;
-  endDate?: Timestamp;
-};
+export const globalMessageTypeSchema = z.object({
+  id: z.string(),
+  active: z.boolean(),
+  title: z.array(languageAndTextSchema).optional(),
+  body: z.array(languageAndTextSchema),
+  type: messageModeSchema,
+  subtle: z.boolean().optional(),
```
mikaelbr, line 21: Should this `.default(false)`?

src/modules/global-messages/types.ts
```diff
@@ -1,20 +1,28 @@
...
+  context: z.array(z.nativeEnum(GlobalMessageContextEnum)),
+  isDismissable: z.boolean().optional(),
```
mikaelbr, line 23: Should this `.default(false)`?

mikaelbr (approved):
Good improvement! Some minor nitpicks/questions, but all in all good

src/modules/global-messages/types.ts
mortennordseth: Yes, that's better 👍

PR #303, opened 2024-06-11
feat: display global messages

mortennordseth:
This PR makes use of the global message context to display global messages on the landing page. The animations aren't perfect, so any suggestions to improve this are welcome. 

image

Closes https://github.com/AtB-AS/kundevendt/issues/17725


src/components/message-box/index.tsx
```diff
@@ -42,6 +42,8 @@ export const MessageBox = ({
   const overrideMode = useStatusThemeColor(type);
   const aria = modeToAria(type);
 
+  if (message === '') return null;
```
mikaelbr, line 45: Can message be undefined here? Or null? Maybe this is a broader check to use

```suggestion
  if (!message) return null;
```

adriansberg: The type asserts that it can't be undefined or null, doesn't it?

mortennordseth: Yes, the type should handle undefined and null values: 

```tsx
export type MessageBoxProps = {
  ...
  title?: string;
  message: string;
  noStatusIcon?: boolean;
  ...
};

```

jonasbrunvoll (approved):
Looks good to me!

adriansberg (approved):
LGTM2, but probably biased since we did this together 😂

src/page-modules/assistant/assistant.module.css
```diff
@@ -7,6 +7,9 @@
   padding: 0 var(--spacings-medium);
   margin: 0 auto;
 }
+.homeLink__container > :last-child:not(:only-child) {
+  margin-bottom: var(--spacings-medium);
+}
```
mikaelbr, line 12: Is this better handled by using `gap` in `homeLink__container`?

mikaelbr: But.. Is homeLink__container in use here now? Seems like it is removed from usage in assistant layout

mortennordseth: Not in use here, so I've removed it. I think using `gap` on the container would result in extra space between the button and the first global message, which we don't want.

src/page-modules/assistant/layout.tsx
```diff
@@ -114,22 +115,14 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
...
-    <div>
-      <div className={style.homeLink__container}>
-        <ButtonLink
-          mode="transparent"
-          href={urls.homePageUrl.href}
-          title={t(PageText.Assistant.homeLink(urls.homePageUrl.name))}
-          icon={{ left: <MonoIcon icon="navigation/ArrowLeft" /> }}
-        />{' '}
-      </div>
+    <SearchContainerLayout>
```
mikaelbr, line 125: Thought I wrote a comment here, but couldn't find it again, so probably didn't add it 😅. But I was thinking about the naming of this component, and I don't really find it intuitive. I wonder if maybe naming it by intention of use is better than what it is supposed to wrap? So what does this actually do "Search Container Layout" says nothing other than "this wrapped the search inputs", but in reality it shows the global messages.

src/modules/global-messages/global-messages.tsx
```diff
@@ -0,0 +1,53 @@
...
+export type GlobalMessagesProps = {
+  context: GlobalMessageContextEnum;
+  className?: string;
+};
+
+export function GlobalMessages({ context, className }: GlobalMessagesProps) {
+  const { language } = useTranslation();
+  const { activeGlobalMessages } = useActiveGlobalMessages();
+
+  if (!activeGlobalMessages || activeGlobalMessages.length === 0) return null;
```
mikaelbr, line 18: I don't really know if i prefer this or if it is clearer or more unclear 🤔 Just thinking out loud

```suggestion
  if (!activeGlobalMessages?.length) return null;
```

mikaelbr: That being said. Why would `activeGlobaleMessages` ever be `undefined`? I think that is a modeling fault. One of the biggest advantages of Arrays and Lists is that they can contain the same API interfaces if they have no content. So there should be very few cases where lists are undefined, but rather be instantiated as empty

src/modules/search-time/selector/index.tsx
```diff
@@ -126,10 +126,7 @@ export default function SearchTimeSelector({
 
             <span className={style.option__label}>
               {selectedMode.mode === state && (
-                <motion.span
-                  layoutId="searchTimeSelector"
-                  className={style.option__selected}
-                />
+                <span className={style.option__selected} />
```
mikaelbr, line 129: Don't know if this change is relevant/intended but probably fine! 😄

mikaelbr (approved):
Could name SearchContainer to be more descriptive, but other than that it looks good

src/modules/global-messages/global-messages.tsx
mortennordseth: `activeGlobalMessages` won't be `undefined` so this is an unnecessary check. I've updated and fixed this to `!activeGlobalMessages.length` 👍

src/modules/search-time/selector/index.tsx
mortennordseth: It's intended because the animation got very weird when the global messages were loaded and shown on the page

PR #305, opened 2024-06-24
feat: adds maximum interchange wait time to details page. Fixes #18410

mikaelbr:
Fixes https://github.com/AtB-AS/kundevendt/issues/18410

I'm unsure whether or not to actually handle the null case or just treat it as 0. But for now, if the max wait time is 0 we just treat it as not defined. 🤔 Open to actually handle it explicitly though.

Screenshot 2024-06-24 at 19 39 54

### DIFF
--- a/src/page-modules/assistant/details/trip-section/index.tsx
+++ b/src/page-modules/assistant/details/trip-section/index.tsx
@@ -225,6 +225,7 @@ export default function TripSection({
         <InterchangeSection
           interchangeDetails={interchangeDetails}
           publicCode={leg.line?.publicCode}
+          maximumWaitTime={leg.interchangeTo?.maximumWaitTime}
         />
       )}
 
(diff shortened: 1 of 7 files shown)


adriansberg:
2024-06-24 at 23 05 50@2x

Shouldn't the wait time in theory be about 2 minutes in this case? Is the reason it is showing that the `maximumWaitTime` is greater than 0, but less than 1?

mikaelbr:
If the time isn't shown it is defined as 0 in the API. Don't know if business logic says that 0 = 2, but can check with Christoffer.

mikaelbr:
But seeing here waiting 0 minutes is not good 😬

adriansberg:
> But seeing here waiting 0 minutes is not good 😬

That means it _is_ defined some how?

mikaelbr:
Yeah, and non-zero.

mikaelbr:
In FRAM at least the value will always be set and is Required by Hogia. So I think the best course of action is to hide it as intended for now. Will check why this is 0 but not caught by that check

mikaelbr:
In meråker case the maximum wait time was -1.

adriansberg:
Typical Meråker I tell you.

adriansberg (approved):
Looks good now 🎉

mortennordseth (approved):
🙌

PR #308, opened 2024-06-25
fix: show correct day label for trip results

mortennordseth:
For some trips, when the previous page has an empty trip pattern list, the day label is wrong (see screenshot). This PR updates the `getPreviousDepartureTime` function to take that into account and fix this bug. 

<details>

**Before**

image

**After**

image

</details>

### DIFF
--- a/src/page-modules/assistant/trip/index.tsx
+++ b/src/page-modules/assistant/trip/index.tsx
@@ -50,13 +50,13 @@ export default function Trip({ tripQuery, fallback }: TripProps) {
         details={
           tripQuery.transportModeFilter || tripQuery.lineFilter
             ? t(
-              PageText.Assistant.trip.emptySearchResults
-                .emptySearchResultsDetailsWithFilters,
-            )
+                PageText.Assistant.trip.emptySearchResults
+                  .emptySearchResultsDetailsWithFilters,
+              )
             : t(
-              PageText.Assistant.trip.emptySearchResults
-                .emptySearchResultsDetails,
-            )
+                PageText.Assistant.trip.emptySearchResults
+                  .emptySearchResultsDetails,
+              )
         }
       />
     );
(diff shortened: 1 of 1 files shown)


src/page-modules/assistant/trip/index.tsx
```diff
@@ -68,12 +68,22 @@ export default function Trip({ tripQuery, fallback }: TripProps) {
...
+      // Iterate backwards to find the first trip with a non-empty tripPatterns list
+      for (let i = tripIndex - 1; i >= 0; i--) {
+        const previousTrip = trips[i];
+        if (previousTrip.tripPatterns.length > 0) {
+          return previousTrip.tripPatterns[previousTrip.tripPatterns.length - 1]
+            ?.expectedStartTime;
+        }
+      }
+      // If no previous trip with non-empty tripPatterns is found
+      return undefined;
```
adriansberg, line 82:
```suggestion
      return;
```

This would be the same, right? IMO it is cleaner, but maybe less explicit.

mortennordseth: That should be the same, yes. I'm not sure if I agree with the change, but not having a strong opinion either 😄 . What do you think, @jonasbrunvoll ☝️

jonasbrunvoll: I prefer `returner undefined` over `return`, as I think it is easier to read when the return property is written explicitly.

mikaelbr: Today in "JavaScript History":
> In pre-ESM module times where strict mode (e.g. `"use strict";`) was not default JavaScript behaviour, `undefined` was not a reserved keyword. Making it so that any external dependency could override the value of undefined with for instance: `undefined = true;` and breaking things. Due to this it was recommended often to use either just `return;` which is implicitly `return undefined;`, or explicitly `return void 0;` as `void` is an operator in JavaScript making every operand return undefined. It could technically also have been `return void 'mikael';` or what ever as the operand, but `void 0` was common practise as it communicated abcense of anything.

mikaelbr (approved):

PR #310, opened 2024-06-28
fix: footer content for Svipper

mortennordseth:
Svipper do not want their fylkeskommune logo or social media links in the footer.

### DIFF
--- a/orgs/troms.json
+++ b/orgs/troms.json
@@ -5,11 +5,7 @@
 
   "supportEmail": "",
 
-  "fylkeskommune": {
-    "logoSrc": "/troms_fylkeskommune.svg",
-    "logoSrcDark": "/troms_fylkeskommune_dark.svg",
-    "name": "Troms fylkeskommune"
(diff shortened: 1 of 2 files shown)


mikaelbr (approved):

orgs/troms.json
```diff
@@ -5,11 +5,7 @@
 
   "supportEmail": "",
 
-  "fylkeskommune": {
-    "logoSrc": "/troms_fylkeskommune.svg",
-    "logoSrcDark": "/troms_fylkeskommune_dark.svg",
-    "name": "Troms fylkeskommune"
-  },
+  "fylkeskommune": {},
```
mikaelbr, line 8:
```suggestion
  "fylkeskommune": null,
```

mortennordseth: I took a look at the type and it should actually be undefined or defined, not null (and definitely not an empty object 😄 ). I've removed the properties from troms.json now.

PR #311, opened 2024-07-03
fix: bottom links layout in the footer

mortennordseth:
Troms won't have social media links or a logo in the footer, so I made some CSS changes to the bottom links so that:

- Center the element if we have one element
- Use space-evenly if we have two elements
- Use space-between if we have three or more elements

(same behavior as in the webshop https://github.com/AtB-AS/webshop2/pull/454)

### DIFF
--- a/src/layouts/shared/footer.tsx
+++ b/src/layouts/shared/footer.tsx
@@ -11,12 +11,37 @@ export type FooterProps = {
   withoutSettings?: boolean;
 };
 
+type SomeLink = {
+  href: string;
+  iconSrc: string;
+  title: string;
+};
+
 const { urls, fylkeskommune } = getOrgData();
 
 export default function Footer({ withoutSettings = false }: FooterProps) {
   const { isDarkMode, toggleDarkmode } = useTheme();
   const { t, language } = useTranslation();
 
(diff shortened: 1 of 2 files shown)


adriansberg:
This is the case for the webshop as well, or is it fixed there?

src/layouts/shared/footer.tsx
```diff
@@ -120,54 +120,55 @@ export default function Footer({ withoutSettings = false }: FooterProps) {
...
-                  rel: 'noreferrer',
-                  title: 'Twitter',
-                }}
-              />
-            )}
-          </div>
-
+          {(urls.facebookLink || urls.instagramLink || urls.twitterLink) && (
+            <div className={style.footer__some}>
+              {urls.facebookLink && (
```
adriansberg, line 125: I feel like it's possible to clean this up in a way that avoids repeating the markup for the button three times. E.g. create an array based on the existing urls with icon and href.

mortennordseth: Good suggestion. I've refactored this now

adriansberg: I like!

adriansberg (approved):

PR #312, opened 2024-07-12
fix: add condition to check whether passengers should stay seated between leg interchanges.   

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/18588

### Background
Currently, the way a bus line change a during a trip is displayed as there is also conducted a bus change. This can be confusing and might lead to customers to get of the bus when thei are supposed to stay seated.

Entur displays this differently, where only the waiting time is stated in the stop details so that there is no misunderstanding among customers who might jump off and wait for the next bus

#### Illustrations
<details>
<summary>screenshots/trip</summary>

Planner Web currently:
Skjermbilde 2024-07-12 kl  11 20 30

Entur:
Skjermbilde 2024-07-12 kl  11 52 42

After this PR:
Skjermbilde 2024-07-12 kl  11 17 09

</details>

<details>
<summary>screenshots/trip-details-view</summary>

Planner Web currently:
Skjermbilde 2024-07-12 kl  12 02 50

Entur:
Skjermbilde 2024-07-12 kl  12 02 02

After this PR:
Skjermbilde 2024-07-12 kl  15 38 35

</details>

### Proposed solution

Extend trip schema with with the property staySeated to check whether passengers should stay seated or not during interchanges between two legs.

### Acceptance Criteria

- [ ] This update should not affect how any trips are being displayed other than the one where `staySeated: true`.  

### Test input
- [ ] Some trips from Mosjøen sentrum to Søvik are lines where  `staySeated: true`.

### DIFF
--- a/src/page-modules/assistant/details/trip-section/interchange-section.tsx
+++ b/src/page-modules/assistant/details/trip-section/interchange-section.tsx
@@ -18,6 +18,7 @@ export type InterchangeSectionProps = {
   interchangeDetails: InterchangeDetails;
   publicCode?: string | null;
   maximumWaitTime?: number;
+  staySeated: boolean | undefined;
 };
 
 export function InterchangeSection(props: InterchangeSectionProps) {
@@ -43,6 +44,7 @@ function useInterchangeTextTranslation({
   publicCode,
   interchangeDetails,
   maximumWaitTime = 0,
+  staySeated,
 }: InterchangeSectionProps) {
   const { t, language } = useTranslation();
 
@@ -59,6 +61,16 @@ function useInterchangeTextTranslation({
         ].join(' ')
       : text;
 
+  if (publicCode && staySeated) {
+    return appendWaitTime(
+      t(
+        PageText.Assistant.details.tripSection.lineChangeStaySeated(
+          interchangeDetails.publicCode,
+          publicCode,
+        ),
+      ),
+    );
+  }
   if (publicCode) {
     return appendWaitTime(
       t(
--- a/src/page-modules/assistant/details/trip-section/interchange-section.tsx
+++ b/src/page-modules/assistant/details/trip-section/interchange-section.tsx
@@ -18,7 +18,7 @@ export type InterchangeSectionProps = {
   interchangeDetails: InterchangeDetails;
   publicCode?: string | null;
   maximumWaitTime?: number;
-  staySeated: boolean | undefined;
+  staySeated: boolean | undefined | null;
 };
 
 export function InterchangeSection(props: InterchangeSectionProps) {
--- a/src/page-modules/assistant/trip/trip-pattern/index.tsx
+++ b/src/page-modules/assistant/trip/trip-pattern/index.tsx
@@ -71,6 +71,18 @@ export default function TripPattern({
     [style['tripPattern--old']]: tripIsInPast,
   });
 
+  const staySeated = (idx: number) =>
+    expandedLegs[idx].interchangeTo?.staySeated === true;
+
+  const shouldHaveRemainedSeatedInPreviousLeg = (idx: number) => {
+    if (idx > 0) return staySeated(idx - 1);
+    return false;
+  };
+
+  const isNotLastLeg = (i: number) => {
+    return i < expandedLegs.length - 1 || collapsedLegs.length > 0;
+  };
+
   return (
     <motion.a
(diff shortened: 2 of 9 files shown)


jonasbrunvoll:
The way the travel planner web and Entur display trips in details view where staySeated is true is still different (images). Should we also update this?


| Entur      |  Planner Web |
| ----------- | ----------- |
|Skjermbilde 2024-07-12 kl  12 02 02| Skjermbilde 2024-07-12 kl  12 02 50
  |

src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -98,66 +98,79 @@ export default function TripPattern({
...
-                {(i < expandedLegs.length - 1 || collapsedLegs.length > 0) && (
-                  <div className={style.legs__legLineContainer}>
-                    <div className={style.legs__legLine} />
-                    <div className={style.legs__legLine} />
+                      )}
+                    </div>
                   </div>
                 )}
+
+                {(i < expandedLegs.length - 1 || collapsedLegs.length > 0) &&
```
adriansberg, line 167: Maybe this condition can be extracted to increase code legibility?

mikaelbr:
Looking at PR with details panel this is probably a good start. It explains why the number changes and can give information to cause less confusion for travellers

src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -71,6 +71,18 @@ export default function TripPattern({
     [style['tripPattern--old']]: tripIsInPast,
   });
 
+  const staySeated = (idx: number) =>
+    expandedLegs[idx].interchangeTo?.staySeated === true;
+
+  const shouldHaveRemainedSeatedInPreviousLeg = (idx: number) => {
+    if (idx > 0) return staySeated(idx - 1);
+    return false;
+  };
```
mikaelbr, line 80: With this usage (further down) you won't have cases where expandedLegs[idx] is undefined, due to checking for it a different place (in shouldHaveRemainedSeatedInPreviousLeg). But the check in shouldHaveRemainedSeatedInPreviousLeg is dependent on internal knowledge of staySeated. So you are kind of mixing knowledge of implementations and spreading it across different functions. No real practical consequences in this case, but it can be seen as a little "un clean" in a sense. And over time these functions could be moved, changed, etc, which introduces this as a problem. It can be better to be somewhat more paranoid when implementing functions like these. As to assume that no consumer has done checks/guards for you. So you would have to do null checks in staySeated also

src/page-modules/assistant/details/trip-section/interchange-section.tsx
```diff
@@ -18,6 +18,7 @@ export type InterchangeSectionProps = {
   interchangeDetails: InterchangeDetails;
   publicCode?: string | null;
   maximumWaitTime?: number;
+  staySeated: boolean | undefined;
```
mikaelbr, line 21: Very much not happy with this where null | undefined | value etc. But not your fault and not specifically with this PR. But wouldn't there also be a case where this is null now?

mikaelbr (approved):

src/page-modules/assistant/trip/trip-pattern/index.tsx
jonasbrunvoll: Makes sense! I agree that splitting the logic between several functions can make the logic a bit unclear. I also now see that by adding  a null checker to `staySeated`, `shouldHaveRemainedSeatedInPreviousLeg` becomes redundant as `staySeated` will return false as a result of the null checker, in case the leg is undefined.

src/page-modules/assistant/details/trip-section/interchange-section.tsx
jonasbrunvoll: Is thath due to that both leg and leg.interchangeTo can be null? In that case, should also the null type be added to `maximumWaitTime` as `leg.interchangeTo.maximumWaitTime` can possibly also be null?

mikaelbr: It is due to serializing "over the wire" for server side communication in Next.js does not support undefined values, but marked as "null". so when mapping data it isn't undefined but null. In practise this is the same in most cases (null and undefined), but not for types.

PR #314, opened 2024-08-01
chore: upgrades dependencies (2024-08-01)

mikaelbr:
Shouldn't break anything. Tests work, solution work, linting works.

eslint and satori can't be updated (still). So keeping them as is.


jonasbrunvoll (approved):
Looks good to me🚀

PR #318, opened 2024-08-02
feat: changes date input to use react-aria

mikaelbr:
Add a custom calendar component. This way we can control the format and not rely on browser implementation. This is apparently extremely important for some people (is there a passive-aggressive emoji somewhere?)

Fixes https://github.com/AtB-AS/kundevendt/issues/18791

<details><summary>Screenshot english</summary>
<p>

Screenshot 2024-08-12 at 11 39 27

</p>
</details> 

<details><summary>Screenshot norwegian</summary>
<p>

Screenshot 2024-08-12 at 11 39 45

</p>
</details> 

<details><summary>Screenshot Firefox English settings, norwegian lang in site</summary>
<p>

Screenshot 2024-08-12 at 11 41 18

</p>
</details>


src/layouts/base/index.tsx
```diff
@@ -19,32 +21,37 @@ export type BaseLayoutProps = PropsWithChildren<{
 export function BaseLayout({ children, title }: BaseLayoutProps) {
   useHtmlDarkMode();
   const theme = useTheme();
-  const { t } = useTranslation();
+  const { t, language } = useTranslation();
 
   const siteTitle = usePageTitle(title);
 
+  // used for calendards and date pickers, transform to locale supported by react-aria
```
adriansberg, line 28:
```suggestion
  // Used for calendars and date pickers, transform to locale supported by react-aria.
```

src/layouts/base/index.tsx
```diff
@@ -19,32 +21,37 @@ export type BaseLayoutProps = PropsWithChildren<{
...
-
-      <OpenGraphBase title={siteTitle} />
-
-      <PageHeader />
-
-      <main className={style.main}>{children}</main>
-
-      <Footer />
-    </div>
+    <I18nProvider locale={i18nLocale}>
```
adriansberg, line 32: Can't this provider be colocated with the other context providers? I guess it's because of `useTranslation` and how it works?

mikaelbr:
I've been trying to get react-testing-library to play nice with `contenteditable` and triggering onchanges, but no luck this far. Too much sunk time into looking into it vs the value it brings right now I think. 🤔

src/layouts/base/index.tsx
mikaelbr: It cannot because it needs to access the outer scope of the contexts yes. So in that case we'd need to introduce another layer of components, which would be for that sole purpose, and then I think this solution is as good as any

adriansberg (approved):

PR #323, opened 2024-08-21
feat: create page for contact schema with feature flag

jonasbrunvoll:
Create feature flag for contact schema.

### DIFF
--- a/orgs/atb.json
+++ b/orgs/atb.json
@@ -40,6 +40,11 @@
       "dev": "http://localhost:3000",
       "staging": "https://atb-staging.planner-web.mittatb.no/",
       "prod": "https://reise.atb.no/"
+    },
+    "contactPageUrls": {
+      "dev": "http://localhost:3000/contact",
+      "staging": "https://atb-staging.planner-web.mittatb.no/contact",
+      "prod": "https://reise.atb.no/contact"
     }
   },
   "journeyApiConfigurations": {
--- a/orgs/atb.json
+++ b/orgs/atb.json
@@ -40,11 +40,6 @@
       "dev": "http://localhost:3000",
       "staging": "https://atb-staging.planner-web.mittatb.no/",
       "prod": "https://reise.atb.no/"
-    },
-    "contactPageUrls": {
-      "dev": "http://localhost:3000/contact",
-      "staging": "https://atb-staging.planner-web.mittatb.no/contact",
-      "prod": "https://reise.atb.no/contact"
     }
   },
   "journeyApiConfigurations": {
--- a/src/layouts/shared/page-header.module.css
+++ b/src/layouts/shared/page-header.module.css
@@ -11,8 +11,10 @@
 }
 
 .pageHeader__inner {
+  width: 100%;
   display: flex;
   align-items: center;
+  justify-content: space-between;
   background-color: var(--static-background-background_accent_0-background);
   color: var(--static-background-background_accent_0-text);
   position: relative;
@@ -59,6 +61,28 @@
 }
 
 .pageHeader__link {
-  color: var(--static-background-background_accent_0-text);
+  composes: typo-body__secondary from '@atb/theme/typography.module.css';
+  transition:
+    background-color 150ms ease,
+    opacity 200ms ease-in-out;
+  display: block;
+}
+.pageHeader__link,
+.pageHeader__link:visited {
+  padding: var(--spacings-medium);
   text-decoration: none;
+  color: var(--static-background-background_accent_0-text);
+}
+.pageHeader__link:hover {
+  background-color: var(--static-background-background_accent_1-background);
+  color: var(--static-background-background_accent_1-text);
+}
+.pageHeader__link:active {
+  opacity: 0.8;
+}
+.pageHeader__link--active {
+  padding-bottom: calc(var(--spacings-medium) - var(--border-width-medium));
+  border-bottom: var(--border-width-medium) solid
+    var(--ws-colors-header-border-bottom);
+  composes: typo-body__secondary--bold from '@atb/theme/typography.module.css';
 }
--- a/src/layouts/shared/page-header.module.css
+++ b/src/layouts/shared/page-header.module.css
@@ -12,6 +12,7 @@
 
 .pageHeader__inner {
   display: flex;
+  align-items: center;
   background-color: var(--static-background-background_accent_0-background);
   color: var(--static-background-background_accent_0-text);
   position: relative;
@@ -55,4 +56,9 @@
 }
 .pageHeader__logoLink:active {
   opacity: 0.8;
+}
+
+.pageHeader__link {
+  color: var(--static-background-background_accent_0-text);
+  text-decoration: none;
 }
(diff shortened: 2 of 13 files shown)


orgs/atb.json
```diff
@@ -40,6 +40,11 @@
       "dev": "http://localhost:3000",
       "staging": "https://atb-staging.planner-web.mittatb.no/",
       "prod": "https://reise.atb.no/"
+    },
+    "contactPageUrls": {
```
mortennordseth, line 44: This should be optional

src/layouts/shared/page-header.tsx
```diff
@@ -10,6 +10,8 @@ export default function PageHeader() {
   const { t } = useTranslation();
   const [isDarkMode] = useDarkMode();
   const { fylkeskommune, urls } = getOrgData();
+  const displayContactPage =
+    process.env.NEXT_PUBLIC_DISPLAY_CONTACT_PAGE === 'true';
```
mortennordseth, line 14: The env variable should be a URL to the API that handles the contact form. If this URL is defined, we show the link (not use true/false)

src/layouts/shared/page-header.tsx
```diff
@@ -10,6 +10,7 @@ export default function PageHeader() {
   const { t } = useTranslation();
   const [isDarkMode] = useDarkMode();
   const { fylkeskommune, urls } = getOrgData();
+  const displayContactPage = process.env.NEXT_PUBLIC_DISPLAY_CONTACT_PAGE;
```
mortennordseth, line 13: This makes me think we need a second env variable, i.e. the pureservice-connect API URL. I don't think that's necessary, and we could rather check "if the contact form API URL is defined, show a link to the contact page". So I suggest renaming this env variable. 

Also, is it necessary for it to be public (`NEXT_PUBLIC_`)?

src/layouts/shared/page-header.tsx
```diff
@@ -49,6 +50,14 @@ export default function PageHeader() {
...
             </Link>
           </h1>
+          {displayContactPage && urls.contactPageUrls && (
+            <Link
+              className={style.pageHeader__link}
+              href={urls.contactPageUrls.dev}
+            >
+              <h1>{t(CommonText.Layout.contactLink)}</h1>
+            </Link>
+          )}
```
mortennordseth, line 60:
```suggestion
          {displayContactPage && (
            <Link className={style.pageHeader__link} href="/contact">
              <h1>{t(CommonText.Layout.contactLink)}</h1>
            </Link>
          )}
```

src/modules/org-data/index.tsx
```diff
@@ -46,9 +46,14 @@ export type OrgData = {
 
     sitemapUrls: {
       dev: string;
-      staging: string,
-      prod: string
-    }
+      staging: string;
+      prod: string;
+    };
+    contactPageUrls?: {
```
mortennordseth, line 52: I'm pretty sure that we don't need to specify these URL's in the org-data 🤔

mortennordseth (changes requested):
I've added some change suggestions that should be looked into before merging 👍

src/pages/contact/index.tsx
```diff
@@ -0,0 +1,32 @@
+import DefaultLayout from '@atb/layouts/default';
```
mortennordseth, line 1: We need to check if the env variable for the contact page is set, or else return 404 not found

src/layouts/shared/page-header.module.css
```diff
@@ -56,3 +57,8 @@
 .pageHeader__logoLink:active {
   opacity: 0.8;
 }
+
+.pageHeader__link {
```
mortennordseth, line 61: Need to add a hover-state for this class. We don't have an "accent 0 hover" color, but try with `opacity: 0.75` or something similar to begin with (test in both dark and light mode for FRAM).

src/translations/common/layout.ts
```diff
@@ -5,6 +5,8 @@ const LayoutInternal = {
   homeLink: (name: string) =>
     _(`Tilbake til ${name}`, `Back to ${name}`, `Tilbake til ${name}`),
 
+  contactLink: _('Kontaktskjema', 'Contact schema', 'Kontaktskjema'),
```
mortennordseth, line 8:
```suggestion
  contactLink: _('Kontaktskjema', 'Contact form', 'Kontaktskjema'),
```

src/layouts/shared/page-header.tsx
mortennordseth: Might want to reuse the styling from the webshop here

src/layouts/shared/page-header.tsx
jonasbrunvoll: Updated the variable name to be less confusing✅

I also agree that the .env variable can be set on the server, and that it is therefore not necessary to make the variable available from the client side (NEXT_PUBLIC_). But this can be resolved later. At the moment, I prioritize getting the whole flow in place from being able to register feedback through the front-end and sending it in to pureService.

src/page-modules/contact/layout.tsx
```diff
@@ -0,0 +1,19 @@
...
+function ContactLayout({ children }: ContactLayoutProps) {
+  const contactPage = process.env.NEXT_PUBLIC_CONTACT_PAGE;
+  if (contactPage) {
+    return (
+      <HomeLinkWithGlobalMessageLayout>
+        {children}
+      </HomeLinkWithGlobalMessageLayout>
+    );
+  }
+  return <ErrorContent statusCode={404} />;
```
mortennordseth, line 16: I think this could be moved out of the layout and handled by the `getServerSideProps` function on the contact page (`src/pages/contact/index.tsx`). See suggested change below

src/pages/contact/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <ContactLayout {...props}>
+        <ContactContent {...props} />
+      </ContactLayout>
+    </DefaultLayout>
+  );
+};
+
+export default ContactPage;
+
+export const getServerSideProps = withGlobalData();
```
mortennordseth, line 32:
```suggestion

export const getServerSideProps = withGlobalData(async () => {
  const contactPage = process.env.NEXT_PUBLIC_CONTACT_PAGE;

  if (!contactPage) {
    return {
      notFound: true,
    };
  }

  return { props: {} };
});
```

src/layouts/shared/page-header.tsx
```diff
@@ -49,6 +51,15 @@ export default function PageHeader() {
               )}
             </Link>
           </h1>
+          {hasContactFormUrl && (
```
mortennordseth, line 54: Let us add a `nav` element around this link element 👍 Also, should this be moved out of the element with the `pageHeader__inner` class? The link looks a bit off when it's placed all the way to the right 🤔

jonasbrunvoll: Do you think it looks better when the link is placed to closer to the logo?

mortennordseth: I think it should look more like this: 

image

image

mortennordseth: Those screenshots are from the webshop, this is from the current implementation in the planner-web:

image

jonasbrunvoll: Skjermbilde 2024-08-26 kl  12 39 44

mortennordseth (approved):
Nice 💯

PR #324, opened 2024-08-22
refactor: initialize map on mobile devices when opened

mortennordseth:
While looking into the excessive Mapbox loads, I found that a significant majority of our users are accessing the site on mobile devices (over 65% for FRAM). On mobile devices, the map isn't displayed until the user clicks the "Open in Fullscreen" option. However, we're currently initializing the map as soon as the page loads, even though it's not visible. This PR optimizes the process by delaying the map's initialization until it is actually opened in fullscreen, which should help reduce our costs.

Part of https://github.com/AtB-AS/kundevendt/issues/18810

### DIFF
--- a/src/components/map/map.tsx
+++ b/src/components/map/map.tsx
@@ -21,6 +21,7 @@ import { useMapLegs } from './use-map-legs';
 import { and } from '@atb/utils/css';
 import { MapLegType, Position } from './types';
 import { useMapTariffZones } from './use-map-tariff-zones';
+import useMediaQuery from '@atb/utils/user-media-query';
 
 export type MapProps = {
   layer?: string;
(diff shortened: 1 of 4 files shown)


adriansberg (approved):
LGTM

adriansberg (commented):
Don't know if this is an issue or not, but if you load the page with a small screen (<650px wide) the map is not initialized and then increase the page width, the map won't be initialized and there is no way to initialize it other than reloading the page. Probably an edge case that is non-critical.

mortennordseth:
> Don't know if this is an issue or not, but if you load the page with a small screen (<650px wide) the map is not initialized and then increase the page width, the map won't be initialized and there is no way to initialize it other than reloading the page. Probably an edge case that is non-critical.

True, most likely an edge case, but I see no reason to not fix it so I did 👍

adriansberg (approved):

PR #320, opened 2024-08-13
feat: extend notices to inform users about the need to contact the relevant authorities to carry out certain trips

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/4203

### Background
It seems that the app is currently not handling notices on the EstimatedCall on which a trip ends: https://github.com/AtB-AS/mittatb-app/blob/51ea34a7b69f8aca381aac5ffc94fa36a88086ae/src/travel-details-screens/utils.ts#L44-L50

#### Illustrations
<details>
<summary>screenshots/video/figma</summary>
Skjermbilde 2024-08-13 kl  14 58 32
Skjermbilde 2024-08-13 kl  14 58 40
</details>

### Proposed solution
To solve, we should probably extend the code above to also include notices from ToEstimatedCall.

### Acceptance Criteria
- [ ] Notices from estimated call are displayed. 

### Test input
- [ ] Tromsø Prostneset hurtigbåtkai - Lauksundskaret ferjekai
- [ ] Terråk hurtigbåtkai - Skotnes hurtigbåtkai

PR #325, opened 2024-08-22
Revert "Extend notifications to include behovsanlop. (#320)"

mortennordseth:
This reverts commit 94660fd36fa5a6fb73490aa5e897d78d488104b4.

Reverting this because of changes to where we should display the messages (see PR in app) and we need to do a planner-web release (https://github.com/AtB-AS/kundevendt/issues/4203)


adriansberg (approved):
Some of these changes will probably be re-added when the updated implementation is made 👍

PR #326, opened 2024-08-23
feat: show destination display via information

mortennordseth:
We're currently not using the via information found in the destination display object and are therefore giving different information than in the app. This updates the line names for the departure and departure details pages. 

<details>

image

image

</details>

### DIFF
--- a/src/page-modules/departures/__tests__/service-journey-data.fixture.ts
+++ b/src/page-modules/departures/__tests__/service-journey-data.fixture.ts
@@ -20,8 +20,8 @@ export const serviceJourneyFixture: ServiceJourneyData = {
       cancellation: false,
       date: '2023-11-10',
       destinationDisplay: {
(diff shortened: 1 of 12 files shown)


src/page-modules/departures/__tests__/service-journey-data.fixture.ts
```diff
@@ -21,6 +21,7 @@ export const serviceJourneyFixture: ServiceJourneyData = {
       date: '2023-11-10',
       destinationDisplay: {
         frontText: 'Vestlia via sentrum-Othilienborg',
```
adriansberg, line 23: Shouldn't this via text be in the `via` array?

mortennordseth: Fixed

adriansberg (approved):

PR #327, opened 2024-08-26
feat: add logic to interact with with contact api.

jonasbrunvoll:
- Added logic to interact with contact API.
- `handlerWithContactFormClient` was created to more easily retrieve information through the jouneryPlanner api later.


mortennordseth (approved):
Looking good 🙌

PR #329, opened 2024-08-29
chore: add initialize map log event

mortennordseth:
Add new log event to further investigate excessive Mapbox loads.

Part of https://github.com/AtB-AS/kundevendt/issues/18810


adriansberg (approved):
:shipit:

PR #331, opened 2024-08-30
feat: add contact section card component

jonasbrunvoll:
Component to display the content of a form. The content is set through children props.


src/page-modules/contact/components/section-card/section-card.module.css
```diff
@@ -0,0 +1,6 @@
+.container {
+  background-color: var(--static-background-background_0-background);
```
mortennordseth, line 2: Could consider adding `color` to have the correct text color

mortennordseth (approved):

PR #339, opened 2024-09-11
feat: ticket-control api endpoint

mortennordseth:
Updated the contact API to have a ticket-control specific API endpoint and updated the data structure. This needs to be fine-tuned later, but works for now


jonasbrunvoll (approved):
Nice😎

PR #347, opened 2024-09-20
feat: file attachments

mortennordseth:
This includes a new input for files that is used by the complaint and feedback form.


jonasbrunvoll (approved):
Looks good🚀

PR #353, opened 2024-09-25
chore: rewrite ticket control to use one state machine for all forms

jonasbrunvoll:
This PR includes the following:
- Rewrite ticket control to use one state machine to handle fee complaint form, feedback form and postpone payment form. 
- Rewrite `commonInputValidation` to retrieve a list of errorMessages. 
  

### Tested:
- [ ] All three forms are posted successfully on submit.    
- [ ] Tested that the correct validation of the input fields appear on submit if missing.


src/page-modules/contact/validation/commonInputValidator.ts
```diff
@@ -139,11 +139,7 @@ export const commonInputValidator = (context: any) => {
     ),
   );
 
-  // Populate context.errorMessages
-  context.errorMessages = inputErrorMessages;
-
-  // Return false if any error
-  return Object.keys(context.errorMessages).length > 0 ? false : true;
+  return inputErrorMessages;
```
mortennordseth, line 142: I think you need to update the travel-guarantee form with this change?

mortennordseth (approved):

src/page-modules/contact/validation/commonInputValidator.ts
jonasbrunvoll: Yes, i will update it👍

PR #354, opened 2024-09-25
fix: page header styling and contact layout

mortennordseth:
I've refactored the header to not use bold typo, have the contact page link to the right and added a "back to travel search" button for the contact layout. 

image


jonasbrunvoll (approved):
Looks good🚀

PR #358, opened 2024-09-26
feat: success and error handling

mortennordseth:
When a form is successfully submitted, we navigate the user to /contact/success; if it fails, we navigate to /contact/error. While submitting we disable the submit button and add a loading indicator.


jonasbrunvoll (approved):
Looks good🚀

PR #359, opened 2024-09-26
fix: button display

jonasbrunvoll:
This PR includes:

- The logic inside `ON_INPUT_CHANGE` is update so that `agreesSecondAgreement` is automatically set to `false` whenever `agreesFirstAgreement` is set to `false`.

- A new condition is added to control the visibility of the submit button.

### DIFF
--- a/src/page-modules/contact/ticket-control/forms/feeComplaintForm.tsx
+++ b/src/page-modules/contact/ticket-control/forms/feeComplaintForm.tsx
@@ -16,65 +16,28 @@ type FeeComplaintFormProps = {
   send: (event: typeof ticketControlFormEvents) => void;
 };
 
-export const FeeComplaintForm = ({ state, send }: FeeComplaintFormProps) => {
-  const { t } = useTranslation();
+type FirstAgreementProps = FeeComplaintFormProps;
 
-  const FirstAgreement = () => {
-    return (
-      <SectionCard
-        title={t(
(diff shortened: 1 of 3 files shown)


src/page-modules/contact/ticket-control/forms/feeComplaintForm.tsx
```diff
@@ -16,423 +16,425 @@ type FeeComplaintFormProps = {
   send: (event: typeof ticketControlFormEvents) => void;
 };
 
-export const FeeComplaintForm = ({ state, send }: FeeComplaintFormProps) => {
-  const { t } = useTranslation();
+type FirstAgreementProps = FeeComplaintFormProps;
```
mortennordseth, line 19: I think you should use TypeScript Pick here so that it's possible to add properties to FeeComplaintForm and not FirstAgreementProps

src/page-modules/contact/ticket-control/index.tsx
```diff
@@ -61,7 +65,7 @@ const TicketControlPageContent = () => {
       {state.context.formType === 'postponePayment' && (
         <PostponePaymentForm state={state} send={send} />
       )}
-      {state.context.formType && (
+      {displayButton && (
```
mortennordseth, line 68:
```suggestion
      {displaySubmitButton && (
```

mortennordseth (approved):

PR #386, opened 2024-10-10
feat: extend support for vkt and farte

mortennordseth:
This PR extends the planner-web so that it can run for VKT and Farte. Includes update of the design-system, added environment variables, org-data and so on.


adriansberg (approved):
LGTM! Didn't verify all links etc., but that can always be revised at a later point.

adriansberg (approved):
LGTM! Didn't verify all links etc., but that can always be revised at a later point.

mortennordseth:
Still a lot of dummy data, but the goal here is to have the planner-web run with org ID vkt and farte 👍

PR #387, opened 2024-10-10
refactor:  minor update in the organization of translation objects

jonasbrunvoll:
Move `appPhoneNumber`, `customerNumber` and `travelCardNumber` into `input` in `contacts.ts`, as these inputs are used in other contexts in addition to ticketControl.feeComplaint.


mortennordseth (approved):
This PR should be prefixed with `refactor` and not `chore`.

PR #394, opened 2024-10-16
feat: searchable selector form component 

jonasbrunvoll:
### Background
In drop-down lists with many options, it is desirable to have a search field to filter the list alternatives.  To achieve this, the component `SearchableSelector` is created.  

#### Illustrations
<details>

| Light    | Dark |
| --------| ------- |
| Skjermbilde 2024-10-18 kl  08 18 36 | Skjermbilde 2024-10-18 kl  08 19 43 |
| Skjermbilde 2024-10-18 kl  08 18 43 | Skjermbilde 2024-10-18 kl  08 19 50 |
| Skjermbilde 2024-10-18 kl  08 19 04 | Skjermbilde 2024-10-18 kl  08 20 00 |

</details>

### Proposed solution
The `SearchableSelector` enables a user to do the following: 
- [x] Select an option from the as the existing `Select` component.
- [x] Filter the list of options by typing the search word into an input field. 

The `SearchableSelector` also ensures the following:
- [x] Updates the specified `state.context.values`. 
- [x] searchable selector´s that depends on other values to be activated, are disabled whenever the dependent value is set back to `undefined`.

### DIFF
--- a/src/page-modules/contact/components/input/description-modal.tsx
+++ b/src/page-modules/contact/components/input/description-modal.tsx
@@ -43,7 +43,7 @@ const DescriptionModal = ({
       document.removeEventListener('keydown', handleEscapeOrClickOutside);
       document.removeEventListener('mousedown', handleEscapeOrClickOutside);
     };
-  }, []);
+  });
 
   if (!isModalOpen) return null;
 
--- a/src/page-modules/contact/components/input/description-modal.tsx
+++ b/src/page-modules/contact/components/input/description-modal.tsx
@@ -43,7 +43,7 @@ const DescriptionModal = ({
       document.removeEventListener('keydown', handleEscapeOrClickOutside);
       document.removeEventListener('mousedown', handleEscapeOrClickOutside);
     };
-  });
+  }, []);
 
   if (!isModalOpen) return null;
 
--- a/src/page-modules/contact/components/input/searchable-select.tsx
+++ b/src/page-modules/contact/components/input/searchable-select.tsx
@@ -1,0 +1,147 @@
+import { useEffect, useState } from 'react';
+import { MonoIcon } from '@atb/components/icon';
+import ErrorMessage from './error-message';
+import { andIf } from '@atb/utils/css';
+import style from './input.module.css';
+import {
+  Key,
+  Button,
+  ComboBox,
+  Input,
+  Label,
+  ListBox,
+  ListBoxItem,
+  Popover,
+  Group,
+} from 'react-aria-components';
+
+export type Option<T> = { id: string; name: string; value: T };
+export type SearchableSelectProps<T> = {
+  label: string;
+  value?: T;
+  placeholder: string;
+  isDisabled?: boolean;
+  options: Option<T>[];
+  error?: string;
+  onChange: (value?: T) => void;
+};
+
+export default function SearchableSelect<T>({
+  label,
+  value,
+  placeholder,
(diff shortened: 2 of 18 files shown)


src/page-modules/contact/components/input/description-modal.tsx
```diff
@@ -43,7 +43,7 @@ const DescriptionModal = ({
       document.removeEventListener('keydown', handleEscapeOrClickOutside);
       document.removeEventListener('mousedown', handleEscapeOrClickOutside);
     };
-  }, []);
+  });
```
mortennordseth, line 46: Why is this changed and why remove the empty dependency list?

src/page-modules/contact/utils.ts
```diff
@@ -75,6 +77,20 @@ export const setLineAndResetStops = (context: any, line: Line) => {
...
+};
+
+export const getStopOptions = (
+  stops: Line['quays'],
+): Option<Line['quays'][0]>[] => {
+  return stops.map((stop) => ({
+    id: stop.id,
+    name: stop.name,
+    value: stop,
+  }));
```
mortennordseth, line 95: These utility functions are very specific to the new search component, so I wouldn't place it in this file but rather have it closer to the component itself

src/page-modules/contact/components/input/description-modal.tsx
jonasbrunvoll: I received a warning when the deployment ran `yarn tsc` during deployment. I don't think it makes any difference if I remove  the empty dependency list or not, but I can undo the change to keep order.

src/page-modules/contact/utils.ts
jonasbrunvoll: Maybe create a folder names `searchable-selector` inside `components` containing a `utils.ts` with the functions?

src/page-modules/contact/components/input/description-modal.tsx
mortennordseth: We use empty dependency arrays for useEffects if we want the useEffect to run once (on initial render). Else, you want to add the dependencies to the array so that it runs when they changes (i.e. `handleEscapeOrClickOutside` in this case). I guess the warning you get is from es lint and not typescript, but you can disable it using `// eslint-disable-line react-hooks/exhaustive-deps` if you want to use an empty dependency list

src/page-modules/contact/components/input/searchable-select.tsx
```diff
@@ -0,0 +1,147 @@
...
+              className={style.searchable_select__listBoxItem}
+            >
+              {({ isSelected, isFocused }) => (
+                <div
+                  className={`${style.searchable_select__item} ${
+                    isFocused ? style.searchable_select__highlight : ''
+                  } ${isSelected ? style.searchable_select__selected : ''}`}
+                >
+                  <span className={style.searchable_select__checkmark}>
+                    {value && isSelected ? '✔' : '\u00A0'}
```
mortennordseth, line 133: Could we use the `Confirm.svg` instead here? https://github.com/AtB-AS/design-system/blob/main/packages/assets/files/common/mono/actions/Confirm.svg

jonasbrunvoll: Sure, we can use `Confirm.svg` instead. Will update.

mortennordseth (approved):

PR #399, opened 2024-10-18
chore: update components import

jonasbrunvoll:
This PR does the following:

- Rename `components/input` to `components/inputs`. (The reason for most of the file deletions and creations).
- Rename `file.tsx` to `file-input.tsx`. Component is named `FileInput`.
- Rename component from `RadioInput` to `Radio`.
- Ensures that all components insides  `components/inputs` are exported as default.  
- Update the component imports inside `page-modules/contact`.


mortennordseth:
Not all of these components are inputs, so naming it "form" or something similar would probably be better (as in form components)

PR #422, opened 2024-11-11
chore: update input label property type

jonasbrunvoll:
Update input label property type from `TranslationString` to `string`.


adriansberg (commented):
Is this change needed?

mortennordseth (approved):
I think this is a good change and more in line with the rest of the application.

PR #425, opened 2024-11-13
feat: extend input component with disabled property

jonasbrunvoll:
Extend input component with disabled property

Skjermbilde 2024-11-13 kl  14 38 23
Skjermbilde 2024-11-13 kl  14 38 32

### DIFF
--- a/src/page-modules/contact/components/form/form.module.css
+++ b/src/page-modules/contact/components/form/form.module.css
@@ -18,6 +18,10 @@
   gap: var(--spacings-small);
 }
 
+.label_disabled {
+  color: var(--text-colors-disabled);
+}
+
 .input {
   background: transparent;
   color: var(--text-colors-primary);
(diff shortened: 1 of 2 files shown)


mortennordseth:
Ref. discussion at the office. Need to take a look at the color choice here. Verify with sketches from the webshop, designers at AtB, etc.

src/page-modules/contact/components/form/form.module.css
```diff
@@ -18,6 +18,10 @@
   gap: var(--spacings-small);
 }
 
+.label_disabled {
+  color: var(--text-colors-disabled);
+}
+
```
mortennordseth, line 24: Could this also be done with `: disabled` selector as you have done with cursor?

jonasbrunvoll: I dont think so. Is that possible when the <label> is outside <input/>?

mortennordseth: Oh, it's the label.

mortennordseth (approved):

PR #432, opened 2024-11-14
feat: set the time of the incident manually

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/19522

### Background
Currently, the time is automatically set when reporting an incident. This results in several users sloppily specifying the correct time, which is again critical information. 

Time of an incident should be mandatory to provided when needed. The user must specify the time himself.

#### Illustrations

<details>
Skjermbilde 2024-11-14 kl  12 13 10

</details>

### Proposed solution

- [x] Add error message property to TimeSelector. 
- [x] Set plannedDepartureTime to `undefined` in the state machines.

### DIFF
--- a/src/page-modules/contact/means-of-transport/forms/delayForm.tsx
+++ b/src/page-modules/contact/means-of-transport/forms/delayForm.tsx
@@ -152,6 +152,10 @@ export const DelayForm = ({ state, send }: DelayFormProps) => {
               value: time,
             })
           }
+          errorMessage={
+            state.context?.errorMessages['plannedDepartureTime']?.[0] ||
+            undefined
+          }
         />
       </SectionCard>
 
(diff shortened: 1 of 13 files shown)


src/page-modules/contact/means-of-transport/forms/delayForm.tsx
```diff
@@ -152,6 +152,10 @@ export const DelayForm = ({ state, send }: DelayFormProps) => {
               value: time,
             })
           }
+          errorMessage={
+            state.context?.errorMessages['plannedDepartureTime']?.[0] ||
+            undefined
```
mortennordseth, line 157: Do you need `|| undefined` here? 🤔

mortennordseth (approved):

PR #439, opened 2024-11-18
fix: add condition to avoid runtime error

jonasbrunvoll:
A runtime error appears when updating the time using the `TimeSelector`. 

### How to replicate:
- First, enter a time. For example: `11:00`.
- Second, attempt updating the time på for `:00` by tapping the the delete button. 

### Illustration
<details>
<Summary>screenshots</Summary>
Skjermbilde 2024-11-18 kl  15 23 42

</details>

### DIFF
--- a/src/page-modules/contact/components/form/time-selector.tsx
+++ b/src/page-modules/contact/components/form/time-selector.tsx
@@ -30,7 +30,7 @@ export default function TimeSelector({
       <TimeField
         value={parsedValue}
         onChange={(change) => {
-          if (change === null) return;
+          if (!change) return;
           onChange(change.toString());
         }}
         hourCycle={24}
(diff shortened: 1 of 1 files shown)


src/page-modules/contact/components/form/time-selector.tsx
```diff
@@ -29,7 +29,10 @@ export default function TimeSelector({
       <Label>{t(label)}</Label>
       <TimeField
         value={parsedValue}
-        onChange={(change) => onChange(change.toString())}
+        onChange={(change) => {
+          if (change === null) return;
```
mortennordseth, line 33: Does this only apply to `null` values?

```suggestion
          if (!change) return;
```

jonasbrunvoll: Good point. This should apply for both `null` and `undefined`.

mortennordseth (approved):

PR #443, opened 2024-11-20
fix: ensure that only situations within valid periods are displayed

jonasbrunvoll:
Currently, situations are displayed outside their validity period.  

Situations should only be displayed whitin their validity period. 

### Illustration
<details>
<Summary>Before</Summary>

bilde (2)

</details>

<details>
<Summary>After</Summary>
Skjermbilde 2024-11-20 kl  10 06 57

</details>

### Acceptance criteria
- Only situation within valid periods are displayed.


mortennordseth:
As discussed: check if this should be done for the assistant page too and verify that this is the same behavior as in the app

mortennordseth (approved):

PR #454, opened 2024-11-25
fix: avoid opening new tab when clicking on link to contact form

jonasbrunvoll:
Avoid opening a new TAB when clicking on the link in the footer to the new contact form.

### DIFF
--- a/src/layouts/shared/footer.tsx
+++ b/src/layouts/shared/footer.tsx
@@ -17,7 +17,7 @@ type SomeLink = {
   title: string;
 };
 
-const { urls, fylkeskommune } = getOrgData();
+const { orgId, urls, fylkeskommune } = getOrgData();
 
 export default function Footer({ withoutSettings = false }: FooterProps) {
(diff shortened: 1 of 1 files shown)


mortennordseth:
Could we check if the url is external instead of making this org-specific?

src/layouts/shared/footer.tsx
```diff
@@ -247,3 +251,13 @@ function LanguageSelections() {
...
   );
 }
+
+function isExternalUrl(url: string) {
+  try {
+    const currentHostname = window.location.hostname;
+    const urlHostname = new URL(url).hostname;
+    return urlHostname !== currentHostname;
+  } catch {
+    return false;
```
adriansberg, line 261: Is this in a try..catch in case `window` isn't defined?

adriansberg (approved):

src/layouts/shared/footer.tsx
jonasbrunvoll: The try..catch is there to ensure that the function does not crash in case an invalid URL is provided.

PR #460, opened 2024-12-05
fix: farte org data and vkt logo

mortennordseth:
- Upgrade assets and theme packages
- Updated org data
- New logo


orgs/farte.json
```diff
@@ -14,25 +14,24 @@
...
-      "en-US": "https://www.farte.no/"
+      "default": "https://www.farte.no/kundeservice/",
+      "en-US": "https://www.farte.no/english/"
     },
     "homePageUrl": {
       "name": "farte.no",
       "href": "https://www.farte.no/"
     },
     "ticketsUrl": {
-      "en-US": "https://www.farte.no/",
```
jonasbrunvoll, line 33: Is there no English page for tickets? Is that why en-us is being removed completely?

mortennordseth: Correct

jonasbrunvoll: Ok👍

jonasbrunvoll (approved):

PR #463, opened 2024-12-11
fix: PostCSS processing order issue for widget

mortennordseth:
Updated the generate-widget-version script to do the PostCSS processing twice to ensure the @atb-as/token plugin is executed on all of the CSS. It's currently not being applied to CSS imported with `composes`. 

There are some issues related to determining postcss plugin order in vite, but this will do for now so that we can release the planner-web for Farte and VKT.


jonasbrunvoll (approved):

tft-abel (approved):
Interesting issue this. Seems like you found a good workaround manually running PostCSS once more. We might find a more sophisticated fix in the future. Looks good to me!

PR #469, opened 2025-01-07
fix: fix cors for reverse endpoint

adriansberg:
Reused the same CORS origins that are used for auto-complete. This might be the reason for CORS error for geolocation.


jonasbrunvoll (approved):
If it works in autocomplete.ts, I´ll guess it works here🤷‍♂️

PR #471, opened 2025-01-08
feat: force Farte light theme

mathiazom:
Adds an optional `forceTheme` option to `OrgData`, which disables dark theme toggling and forces the specified theme.

This is useful for organizations without a dark theme (or light theme), like Farte.

Without `"forceTheme"` in `farte.json`:

image

With `"forceTheme": "light"` in `farte.json`:

image


jonasbrunvoll (approved):
Looks good!

PR #473, opened 2025-01-13
feat: quay public code prefix

mathiazom:
VKT requested a prefix for quay public codes, similar to entur.no

> entur.no
>
> image

#### Before
image

#### After
image

### DIFF
--- a/src/page-modules/assistant/details/utils.ts
+++ b/src/page-modules/assistant/details/utils.ts
@@ -1,19 +1,23 @@
 import { fromCETToLocalTime } from '@atb/utils/date';
 import { parseTripQueryString } from '../server/journey-planner';
+import { TranslateFunction } from '@atb/translations';
+import { Assistant } from '@atb/translations/pages';
 
-export function formatQuayName(quayName?: string, publicCode?: string | null) {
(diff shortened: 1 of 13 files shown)


src/page-modules/assistant/details/utils.ts
```diff
@@ -1,19 +1,23 @@
...
 import { parseTripQueryString } from '../server/journey-planner';
+import { TranslateFunction } from '@atb/translations';
+import dictionary from '@atb/translations/dictionary';
 
-export function formatQuayName(quayName?: string, publicCode?: string | null) {
+export function formatQuayName(t: TranslateFunction, quayName?: string, publicCode?: string | null) {
   if (!quayName) return;
   if (!publicCode) return quayName;
-  return `${quayName} ${publicCode}`;
+  const prefix = t(dictionary.quayPublicCodePrefix);
```
adriansberg, line 9: - Should the prefix be capitalized or not?
- Should there be a comma before the prefix?

I do not have any clear answer to these, but we should maybe consider them at least.

adriansberg (approved):
LGTM

PR #479, opened 2025-01-17
feat: more global messages contexts

mathiazom:
image
image

---

image
image

---

image
image

---

image
image

### DIFF
--- a/src/components/message-box/message-box.module.css
+++ b/src/components/message-box/message-box.module.css
@@ -6,6 +6,10 @@
 .borderRadius {
   border-radius: token('border.radius.regular');
 }
+.subtle {
+  background-color: token('color.background.neutral.0.background') !important;
+  border: none;
+}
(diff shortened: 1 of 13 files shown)


src/components/message-box/message-box.module.css
```diff
@@ -6,6 +6,10 @@
 .borderRadius {
   border-radius: token('border.radius.regular');
 }
+.subtle {
+  background-color: token('color.background.neutral.0.background') !important;
```
adriansberg, line 10: Necessary to use `!important` here?

adriansberg (approved):
Looking good!

src/components/message-box/message-box.module.css
mathiazom: Nope, that was from before I added `backgroundColor: subtle ? undefined : ...` 🙃

PR #481, opened 2025-01-21
feat: time-specific and dismissable global messages

mathiazom:
Global messages will now respect the `isDismissable`, `startDate` and `endDate` if present.

Implementation is heavily inspired by webshop2 

image

### DIFF
--- a/src/components/message-box/index.tsx
+++ b/src/components/message-box/index.tsx
@@ -21,6 +21,7 @@ export type MessageBoxProps = {
   message: string;
   noStatusIcon?: boolean;
   onClick?: () => void;
+  onDismiss?: () => void;
   borderRadius?: boolean;
   subtle?: boolean;
 };
@@ -32,6 +33,7 @@ export const MessageBox = ({
   textId,
   title,
   onClick,
+  onDismiss,
   borderRadius = true,
   subtle = false
 }: MessageBoxProps) => {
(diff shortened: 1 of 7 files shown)


src/components/message-box/index.tsx
```diff
@@ -92,6 +83,19 @@ export const MessageBox = ({
...
         )}
       </div>
+      {onDismiss && (
+        <button
+          type="button"
+          onClick={onDismiss}
+          className={style.message__close}
+        >
+          <MonoIcon
+            icon={'actions/Close'}
```
mikaelbr, line 93:
```suggestion
            icon="actions/Close"
```

src/modules/global-messages/context.tsx
```diff
@@ -10,33 +11,99 @@ import { collection, onSnapshot, query, where } from '@firebase/firestore';
...
+  );
+
+  const isWithinTimeRange = useCallback(
+    (globalMessage: GlobalMessageType) => {
+      const startDate = globalMessage.startDate;
+      const endDate = globalMessage.endDate;
+      return (!startDate || startDate.getTime() <= now) && (!endDate || endDate.getTime() >= now);
+    },
+    [now],
+  );
```
mikaelbr, line 58: Hmmmmmm. Is this the best way to solve this? 🤔 This would create new functions for everywhere for each second but in most cases not be used. Would it be simpler to have an useEffect on time and remove it from the message list at all? 🤔 and have `isPastEndDate` and `isWithinTimeRange` as functions outside of the scope at all? Doing for instance:

```
function isWithinTimeRange (globalMessage: GlobalMessageType) {
      const now = new Date();
      const startDate = globalMessage.startDate;
      const endDate = globalMessage.endDate;
      return (!startDate || startDate.getTime() <= now) && (!endDate || endDate.getTime() >= now);
    }
```

Would change flow from being "create new function every time date changes so we can refer to date" to "refer to date when invoking the function". Seems to be unnecessary complicated here 🤔

mathiazom: Started to look into this, but opted to just reuse the implementation from webshop2. I agree that it seemed a little too complicated 🤔

mathiazom: Better now? 😇

mikaelbr (approved):
👍

PR #482, opened 2025-01-24
fix: disallow robots in non-production

mathiazom:
Updated `robots.txt` generator to use `Disallow` for non-`production` environments, i.e. `staging` and `dev`.

Mainly to avoid this situation:

image


mikaelbr (approved):
Good!

adriansberg (approved):

PR #484, opened 2025-01-27
fix: Fixed too strict zod-schema for global messages

strandlie:
Don't know if the old implementation is on purpose, but it seems a little odd so here is a suggestion. 

The old implementation would not accept a globalMessage unless **all** of the values in `context` existed in `GlobalMessageContextEnum`. Examples of where this is not true are `app-assistant` and `app-ticketing`. 

Added a little flexibility where values in `GlobalMessageContextEnum` are checked and cast as such, but other values are discarded. Also added some documentation on this.

### DIFF
--- a/src/modules/global-messages/types.ts
+++ b/src/modules/global-messages/types.ts
@@ -8,6 +8,13 @@ const messageModeSchema = z.union([
   z.literal('error'),
 ]);
 
+/**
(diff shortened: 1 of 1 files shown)


mathiazom (approved):

rosvik (approved):
Makes sense!

PR #485, opened 2025-02-04
feat: Improved the header to reflect the new design

strandlie:
Closes AtB-AS/kundevendt#19529


src/pages/departures/[[...id]].tsx
```diff
@@ -50,6 +54,7 @@ const DeparturesPage: NextPage<DeparturesPageProps> = (props) => {
   return (
     <DefaultLayout {...props} title={getDepartureTitle(props)}>
       <DeparturesLayout {...props}>
+        <GlobalMessages context={GlobalMessageContextEnum.plannerWeb} />
         <DeparturesRouting {...props} />
```
Comment (rosvik) [inline src/pages/departures/[[...id]].tsx:58]: Does this mean that global messages will be shown only after results have loaded? Could it be part of `DeparturesLayout` instead (and `AssistantLayout` for trip search)?

rosvik (commented):
Some issues with padding on this departure page:

Screenshot 2025-02-05 at 10 12 14

src/pages/departures/[[...id]].tsx
Comment (strandlie) [reply src/pages/departures/[[...id]].tsx:58]: Yes, and that was intentional to get the same behavior as for the `appTripResults` GlobalMessage [in the app](https://github.com/AtB-AS/mittatb-app/blob/master/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx#L421-L440). 

But now that I think about it, these cases are a little different because in the app we have the `app-assistant` context which is always shown, but if we make this change we have nothing like that for the web-planner. I will try your suggestion.

Comment (strandlie) [reply src/pages/departures/[[...id]].tsx]: Fixed now. It's better like this :)

PR #498, opened 2025-03-18
feat: add border around search dropdown

rosvik:
Adds a tiny border around the search dropdown, which was missing in https://github.com/AtB-AS/kundevendt/issues/20062

Screenshot 2025-03-18 at 10 24 43


hanne-at-atb (approved):
🐗

PR #499, opened 2025-03-18
chore: Upgrade tpw to new colors and buttons

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/19971

image


strandlie:
All the new css like

```
.button--primary,
.button--primary:visited {
  color: token('color.foreground.inverse.primary');
  background-color: token('color.interactive.0.default.background');
}
```

are added because they are new in the design system?

jorelosorio:
> All the new css like
> 
> ```
> .button--primary,
> .button--primary:visited {
>   color: token('color.foreground.inverse.primary');
>   background-color: token('color.interactive.0.default.background');
> }
> ```
> 
> are added because they are new in the design system?

Yes, they are referenced as primary, and secondary buttons.

src/layouts/shared/footer.tsx
```diff
@@ -165,7 +165,7 @@ export default function Footer({ withoutSettings = false }: FooterProps) {
               {someLinks.map((link) => (
                 <ButtonLink
                   key={link.href}
-                  mode={isDarkMode ? 'interactive_2' : 'interactive_1'}
+                  mode={isDarkMode ? 'interactive_2' : 'secondary'}
```
strandlie, line 168: Will `interactive_2`still work, or should this be changed to something like `primary` as well?

src/page-modules/assistant/layout.tsx
```diff
@@ -271,8 +273,9 @@ function AssistantLayout({ children, tripQuery }: AssistantLayoutProps) {
             mode={
               orgId === 'fram'
                 ? 'interactive_0--bordered-light-outline'
```
strandlie, line 275: Same as the other comment: should `interactive_0--bordered-light-outline`

strandlie (approved):

strandlie:
Looks good! I just added some questions because I don't know how the new and old styles of referencing colors in the design system works together yet :)

src/layouts/shared/footer.tsx
jorelosorio: Still works.

jorelosorio: But, I think at some point we need to make sure we have only primary, and secondary, right now there are a lot of types.

jorelosorio: image

PR #502, opened 2025-03-20
Preparations for TPW for AtB

strandlie:
Work towards AtB-AS/kundevendt#20014

This PR does two things:

- Adds CORS handling for prod-environments of the widget for AtB
- Adds a e.preventDefault() for a-elements specific, instead of closest a-element in js-tablist, as the old implementation did not catch all cases


reidzeibel (approved):
LGTM! 🚀

PR #503, opened 2025-03-22
fix: search field adjustments

Oleninkaa:
### Closes https://github.com/AtB-AS/kundevendt/issues/19868
I'm considering the appearance of the switch button `searchInputButton`. Should we hide it on hover, or should it look like I have it? Also, I understand that there may be some odd and unnecessary solutions in my code regarding styling—especially with the borders and outlines of the input items 🥲😁

**Reiseplanlegger:**

Skjermbilde 2025-03-22 kl  14 39 06 <br>

**Avganger:**

Skjermbilde 2025-03-22 kl  14 39 13|

| **'Fra'** selected | **'Til'** selected  |
|--------|--------|
| Skjermbilde 2025-03-22 kl  14 44 07 | Skjermbilde 2025-03-22 kl  14 44 24 | 

****'Fra'** selected at 'Avganger':**

Skjermbilde 2025-03-22 kl  14 44 38


src/page-modules/assistant/assistant.module.css
```diff
@@ -27,8 +27,17 @@
   flex-direction: column;
 }
 
-.input > :last-child {
-  margin-top: 0.375rem;
+.inputGroup {
+  position: relative;
+  display: grid;
+  grid-template-rows: 1fr 1fr;
+  background-color: token('color.foreground.dynamic.disabled');
```
strandlie, line 34: Should this be border-color, not background-color? It looks that way in Figma to me

src/page-modules/assistant/assistant.module.css
```diff
@@ -27,8 +27,17 @@
...
+  border-radius: token('border.radius.regular');
+}
+
+.inputGroup > div {
+  margin: 0 calc(-1 * token('border.width.slim'))
```
strandlie, line 39: I don't really understand what is going on here. Maybe you could explain? We can do it in person when we are both in the office, if that is easier :)

src/page-modules/assistant/assistant.module.css
```diff
@@ -80,13 +89,23 @@
...
   align-items: center;
+  position: absolute;
+  top: 50%;
+  right: token('spacing.medium');
+  transform: translate(0, -50%);
+  z-index: 100;
+}
+
+.container:focus-within .searchInputButton {
+  border-color: token('color.interactive.2.outline.background');
```
strandlie, line 108: I find it a little strange that the border of only the searchInputButton changes color when the search-inputGroup gets focus. Maybe the border of the whole container could change color instead? Ask @ckbilstad if you are unsure :)

strandlie:
Also remember to make the changes in `src/widget/widget.ts`, as that is not updated by these changes is what is shown on the front pages of the OMS-partners.

You can see the result on `localhost:3000/widget`

src/page-modules/assistant/assistant.module.css
Oleninkaa: Sorry, I should have mentioned this. To avoid double borders, I discovered this solution on stack overflow. Therefore, we use negative top and right margins. I tried using a different solution now, which is just a box-shadow, and it also works well! So is box-shadow a better option than border and negative margin?

Skjermbilde 2025-03-26 kl  10 18 28
Skjermbilde 2025-03-26 kl  10 18 17

**Here is the second way of doing this:**
Skjermbilde 2025-03-26 kl  10 53 58

src/page-modules/assistant/assistant.module.css
Oleninkaa: The appearance was approved by Christian, so everything is fine 😊

src/page-modules/assistant/assistant.module.css
strandlie: Well, personally I would prefer the `box-shadow`, but its up to you!

PR #511, opened 2025-03-26
ci: use GAR instead of GCR

rosvik:
Shamelessly stolen from https://github.com/AtB-AS/webshop2/pull/568

@mjenssen Is there something else that needs to be done to migrate planner-web to GCR?

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,7 +10,7 @@ jobs:
   build:
     uses: atb-as/workflows/.github/workflows/cluster-docker-build-tag-push.yaml@v2
     with:
-      image: gcr.io/atb-mobility-platform/planner-web
+      image: europe-west1-docker.pkg.dev/amp-artifacts/amp/planner-web
       build_args: |
         ATB_NEXT_PUBLIC_MAPBOX_DEFAULT_LAT=63.43457
         ATB_NEXT_PUBLIC_MAPBOX_DEFAULT_LNG=10.39844


mjenssen:
> Shamelessly stolen from [AtB-AS/webshop2#568](https://github.com/AtB-AS/webshop2/pull/568)
> 
> @mjenssen Is there something else that needs to be done to migrate planner-web to GCR?

This PR shouldn't be necessary at all. The workflow does this for you, see below. webshop2 is unique in that it is a custom setup for everything, while planner-web uses the common workflow that everything else also uses.

https://github.com/AtB-AS/workflows/blob/91546c9b68f29a6dced6d918ea7b4461a6f0b4a9/.github/workflows/cluster-docker-build-tag-push.yaml#L125

.github/workflows/docker.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   build:
     uses: atb-as/workflows/.github/workflows/cluster-docker-build-tag-push.yaml@v2
     with:
-      image: gcr.io/atb-mobility-platform/planner-web
+      image: europe-west1-docker.pkg.dev/amp-artifacts/amp/planner-web
```
rosvik, line 13: @mjenssen Aha, so we only set the image name then?

```suggestion
      image: planner-web
```

mjenssen: No, it should be like it was without this PR

rosvik: Then why are the builds failing? 🤔 https://github.com/AtB-AS/planner-web/actions/runs/14081026165/job/39433812339

mjenssen: This is because the common one was _also_ pushing to GCR (as mentioned in Slack earlier). This has been fixed in the common workflow now. The next action run should work

PR #513, opened 2025-04-02
feat: add checkbox for showing only stop places on trip search

reidzeibel:
Part of https://github.com/AtB-AS/kundevendt/issues/20334

One of the most-requested feature on from the kundeservice team feedback. This will add a checkbox that is similar to the one found in the app, where the search autocomplete will only show stop places if the checkbox is checked.

This is a very rough implementation and lots of improvement might be found in the CSS. (I'm super noob in CSS ✌🏼 ).

- We already have checkbox component from the `contact form` which FRAM uses.
- I extracted that checkbox component (and error-message component) to generic use component.
- Added the checkbox to the search box (top of the list) and modified the css part so it allows expanding the clickable area.
- Modified the API implementation to show only stop places.

https://github.com/user-attachments/assets/8500eaa7-f630-4990-b0c6-e02b4049efb4

### Acceptance Criteria
- [ ] Checkbox appears inside the list when user types something into the search bar
- [ ] Checkbox doesn't show when there are nothing in the search bar
- [ ] Checking the checkbox shows only stop places
- [ ] Unchecking the checkbox shows all results
- [ ] Checkbox can be selected using arrow keys on keyboard
- [ ] While selected, the checkbox can be toggled by using the enter key on keyboard
- [ ] If an item other than the checkbox are selected, the enter key should close the list and use the selected value as the selected item.

### DIFF
--- a/src/components/error-message/error-message.module.css
+++ b/src/components/error-message/error-message.module.css
@@ -1,0 +1,6 @@
+/* error message */
+
+.errorMessage {
+  display: flex;
+  gap: token('spacing.small');
+}
--- a/src/components/checkbox/checkbox.module.css
+++ b/src/components/checkbox/checkbox.module.css
@@ -11,6 +11,7 @@
   padding: token('spacing.xSmall');
   border-radius: token('border.radius.small');
 }
+
 .checkbox:not(.checkbox--readonly) {
   background-color: token('color.interactive.2.default.background');
   color: token('color.interactive.2.default.foreground.primary');
@@ -28,6 +29,29 @@
   color: token('color.interactive.2.disabled.foreground.primary');
 }
 .checkbox:not(.checkbox--readonly):focus-within {
+  outline: 0;
+  box-shadow: inset 0 0 0 token('border.width.slim')
+    token('color.interactive.2.outline.background');
+}
+
+.checkbox:not(.checkbox--readonly):is(.checkbox--transparent) {
+  background-color: transparent;
+  color: token('color.interactive.0.default.foreground.primary');
+}
+.checkbox:not(.checkbox--readonly):is(.checkbox--transparent):hover {
+  background-color: transparent;
+  color: token('color.interactive.0.hover.foreground.primary');
+}
+.checkbox:not(.checkbox--readonly):is(.checkbox--transparent):active {
+  background-color: transparent;
+  color: token('color.interactive.0.default.foreground.primary');
+}
+.checkbox:not(.checkbox--readonly):is(.checkbox--transparent):disabled {
+  background-color: transparent;
+  color: token('color.interactive.0.disabled.foreground.primary');
+}
+.checkbox:not(.checkbox--readonly):is(.checkbox--transparent):focus-within {
+  background-color: transparent;
   outline: 0;
   box-shadow: inset 0 0 0 token('border.width.slim')
     token('color.interactive.2.outline.background');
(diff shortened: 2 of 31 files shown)


src/components/error-message/error-message.module.css
```diff
@@ -0,0 +1,6 @@
+/* error message */
```
strandlie, line 1: Unnecessary comment. I see it was copied from another file, but here it can be deleted

strandlie (commented):
I can imagine that this is something a lot of users are asking for!

3 thoughts: 
- Have you discussed with a designer whether this should be placed within the filter-segment? I would argue it fits better there.
- I think this should be stored (localStorage is sufficient) so that the user does not have to change it each time. I would think that is done in the app today?
- We should also have the same choice in the Departures-tab

reidzeibel:
@strandlie 

> I can imagine that this is something a lot of users are asking for!
> 
> 3 thoughts:
> 
> * Have you discussed with a designer whether this should be placed within the filter-segment? I would argue it fits better there.
> 
> * I think this should be stored (localStorage is sufficient) so that the user does not have to change it each time. I would think that is done in the app today?
> 
> * We should also have the same choice in the Departures-tab

I'll try to go through the points:

- Discuss with designer? No, this was intended to be a quick-win, didn't want to spend days on this, and if there are release nearby, this can be merged first and can be followed up later with improvements. Also, on the app, the checkbox appears right below the search box, if we put the checkbox inside the filter, it would add 1 extra action and they would have to direct their focus to another part of the page, which I think would reduce the practicality of the function.
- Local storage seems to be a great idea, I can add that!
- I intentionally left it out of the departures tab because if you put any address there, the bus stop option will show anyway. I'll add it though.

strandlie:
> @strandlie 
> 
> 
> 
> > I can imagine that this is something a lot of users are asking for!
> 
> > 
> 
> > 3 thoughts:
> 
> > 
> 
> > * Have you discussed with a designer whether this should be placed within the filter-segment? I would argue it fits better there.
> 
> > 
> 
> > * I think this should be stored (localStorage is sufficient) so that the user does not have to change it each time. I would think that is done in the app today?
> 
> > 
> 
> > * We should also have the same choice in the Departures-tab
> 
> 
> 
> I'll try to go through the points:
> 
> 
> 
> - Discuss with designer? No, this was intended to be a quick-win, didn't want to spend days on this, and if there are release nearby, this can be merged first and can be followed up later with improvements. Also, on the app, the checkbox appears right below the search box, if we put the checkbox inside the filter, it would add 1 extra action and they would have to direct their focus to another part of the page, which I think would reduce the practicality of the function.


I would at least discuss it with a designer to make sure we don't mess up the UX by adding it here. Don't need sketches and time consuming stuff. 

And another point is that if we add localStorage to it, it wouldn't matter as much if the option is moved to the filter section, as it wouldn't need to be used as often.

reidzeibel:
Discussed with @Mariblaas on Friday, and will decide on where to put the checkbox after Mari decides where should we put it. Local storage will be added regardless.

reidzeibel:
Added localstorage, updated video preview, the checkbox will now show on the list item when the user types anything to the search bar

reidzeibel:
@strandlie I have refactored the code and added some accessibility improvement to the checkbox, let me know what you think https://github.com/AtB-AS/planner-web/pull/513/commits/34a8aa4434b3396353be991b44eee27b06e95269

reidzeibel:
Added some acceptance criteria for testers.

PR #507, opened 2025-03-25
fix: Colors and buttons fixes

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/19971

PR #514, opened 2025-04-03
fix: Oms colors and theme fixes

jorelosorio:
Requires https://github.com/AtB-AS/planner-web/pull/507 to be merged first


src/layouts/shared/page-header.tsx
```diff
@@ -54,7 +50,14 @@ export default function PageHeader() {
         <ButtonLink
           href={urls.homePageUrl.href}
           title={t(CommonText.Layout.homeLink(urls.homePageUrl.name))}
-          icon={{ right: <MonoIcon icon="navigation/ExternalLink" /> }}
+          icon={{
+            right: (
+              <MonoIcon icon="navigation/ExternalLink" overrideMode={'dark'} />
```
strandlie, line 55: Should this always be dark mode? For all partners?

strandlie (approved):
Looks good! Only one question

reidzeibel:
@jorelosorio seems like there's a conflict here

src/layouts/shared/page-header.tsx
jorelosorio: Yes for the new color configurations, now this is fixed with the merge of the previous PR.

PR #512, opened 2025-03-28
Simplify types and remove some zod

strandlie:
Contributes to AtB-AS/kundevendt#19334

During development of the issue above, the typing system with GraphQL-queries, documents, zod-schemas and validation was found to be time-consuming to work with and get right. At the same time the AtB-team has positive experience with not validating GraphQL-return types but only using the generated types directly from the app and BFF. 

For that reason we are proposing to removing zod-validation on GraphQL-responses, and using the generated types directly, but that impacts a lot of files so I don't expect anyone to look through every file since this PR is massive.

Running the app locally and testing known cases is more a more useful way to do code review. As far as I can see the app works perfectly and no different than before, but I might have missed some cases. 

## Test input
- This affects a lot of code. Useful to do some thorough testing to see if things still works as espected

PR #515, opened 2025-04-03
Fix issues from types-PR

strandlie:
Fixes feedback from AtB-AS/planner-web#512

- Adds missing cursor-information, such that "Load more" now actually loads more trips.
- Fixes types of Position/Coordinates such that map now works as intended

### DIFF
--- a/src/components/map/use-map-legs.ts
+++ b/src/components/map/use-map-legs.ts
@@ -116,14 +116,14 @@ export const useMapLegs = (
   }, [mapRef, mapLegs, addToMap, addStartEndText]);
 };
 
-const createRouteFeature = (points: PositionType[]): AnySourceData => ({
+const createRouteFeature = (positions: PositionType[]): AnySourceData => ({
   type: 'geojson',
   data: {
     type: 'Feature',
     properties: {},
     geometry: {
       type: 'LineString',
-      coordinates: points.map((pos) => [pos.lon, pos.lat]), // Reversed for some reason?
+      coordinates: positions.map((pos) => [pos.lon, pos.lat]), // [longitude, latitude] because GeoJSON
     },
   },
 });
@@ -161,8 +161,8 @@ const createRouteLayer = (
 };
 
 const createStartEndCircle = (
-  startPoint: PositionType,
-  endPoint: PositionType,
+  startPosition: PositionType,
+  endPosition: PositionType,
 ): AnySourceData => ({
   type: 'geojson',
   data: {
@@ -170,11 +170,11 @@ const createStartEndCircle = (
     geometries: [
       {
         type: 'Point',
-        coordinates: [startPoint.lon, startPoint.lat],
+        coordinates: [startPosition.lon, startPosition.lat], // [longitude, latitude] because GeoJSON
       },
       {
         type: 'Point',
-        coordinates: [endPoint.lon, endPoint.lat],
+        coordinates: [endPosition.lon, endPosition.lat], // [longitude, latitude] because GeoJSON
       },
     ],
   },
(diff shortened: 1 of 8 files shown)


src/components/map/use-map-legs.ts
```diff
@@ -191,14 +190,14 @@ const createStartEndLayer = (id: number | string, color: string): AnyLayer => ({
...
-const createStartEndTextPoint = (point: Position): AnySourceData => ({
+const createStartEndTextPoint = (point: PositionType): AnySourceData => ({
   type: 'geojson',
   data: {
     type: 'Feature',
     properties: {},
     geometry: {
       type: 'Point',
-      coordinates: [point.lat, point.lon],
+      coordinates: [point.lon, point.lat],
```
jorelosorio, line 200: Is a bit confusing to see a coordinate (Lon, Lat) be called as point, position and coordiantes at the same place, we might consider call them Point only. Might not be part of this PR but as a suggestion for future implementations refactoring.

src/components/map/utils.ts
```diff
@@ -46,28 +47,29 @@ export const mapToMapLegs = (
...
-  const fromCoordinates: [number, number] = [
-    fromStopPlace?.latitude ?? 0,
-    fromStopPlace?.longitude ?? 0,
-  ];
-  const toCoordinates: [number, number] | undefined = toStopPlace
-    ? [toStopPlace?.latitude ?? 0, toStopPlace?.longitude ?? 0]
-    : undefined;
+  const fromCoordinates: PositionType = {
+    lat: fromStopPlace?.latitude ?? defaultPosition.lat,
+    lon: fromStopPlace?.longitude ?? defaultPosition.lon,
```
jorelosorio, line 58: Great! No more in the ocean if fails.

src/components/map/use-map-legs.ts
reidzeibel: what I don't like in the current TPW is that in many places, coordinate is `[lon, lat]`, instead of `[lat, lon]`, I thought it was a convention to have `lat` first then `lon`, but this is just my rant.... 😮‍💨

src/components/map/utils.ts
```diff
@@ -46,28 +47,29 @@ export const mapToMapLegs = (
...
+  };
 
+  const toCoordinates: PositionType | undefined = toStopPlace
+    ? { lat: toStopPlace?.latitude ?? 0, lon: toStopPlace?.longitude ?? 0 }
```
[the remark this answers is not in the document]
reidzeibel: If we want to have default values, we should probably provide a point for every OMS partners that uses TPW, so that they can have the default values based on their location. Can be done in another PR.

reidzeibel (approved):
Looks good, can merge if Jorge's comments are addressed 👍🏼

src/components/map/use-map-legs.ts
strandlie: I 100% agree with both of you, and I can give a little context. 

Before AtB-AS/planner-web#512 it was much worse. Most coordinates was just [number, number], and you had to work hard to figure out which was latitude and which was longitude. I have tried to fix that by applying the PositionType everywhere. I would also argue that "position" is the correct abstraction, because the type actually represents a physical position - while coordinates are "more abstract", since coordinates can be coordinates to anything. 

But to what @reidzeibel comments, most of the functions in this "use-map-legs.ts"-file are mapping-functions from the domain-model in the app to GeoJSON-format which is used for the map. And GeoJSON has a convention of [longitude, latitude] while arguably the convention everywhere else is [latitude, longitude]. I can add a comment that explains this. 

So to conclude, I think that now you do not have to worry about the order of latitude, longitude anywhere else than in this file. Everywhere else is just `{ lat: number, lon: number}: PositionType`. And this file makes the conversion clear.

strandlie: But I can change from `point` to `position` if that helps @jorelosorio

PR #522, opened 2025-04-22
Improve search section

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/19332

Requires https://github.com/AtB-AS/planner-web/pull/514 to be merged

image


public/widget/IYFwRkA/2.36.0/planner-web.css
```diff
@@ -1 +1 @@
...
...cing-small)}.widget-module__listItem:not(:last-child){border-bottom:1px solid var(--color-border-primary-background)}.widget-module__listItem[aria-selected=true],.widget-module__listItem:hover{background-color:var(--color-interactive-2-active-background);font-weight:700}.widget-module__itemIcon{padding:var(--spacing-small)}.widget-module__itemIcon img{display:block;height:1.25rem}.widget-module__itemLocality{margin-left:var(--spacing-small)}.widget-module__popupContainer{left:0;list-style-position:inside;padding:0;margin:0;background-color:var(--color-background-neutral-0-background)}.widget-module__messageBox{background:var(--color-status-error-primary-background);color:var(--color-status-error-primary-foreground-primary);display:flex;padding:var(--spacing-medium);border-radius:var(--border-radius-regular)}.widget-module__messageBox[hidden]{display:none}.widget-module__inheritFont,.widget-module__inheritFont *{font-family:inherit!important}.widget-module__singleColumnLayout .widget-module__main{grid-template-columns:1fr}.widget-module__singleColumnLayout .widget-module__search_container{max-width:unset}.widget-module__singleColumnLayout .widget-module__selector_options{width:100%}
```
strandlie, line 1: The widget should be stable and backwards compatible so we should never change existing versions of the widget. Instead, bump the version number in package-json and run the yarn generate-widget-version such that a new version is generated and the existing versions are kept untouched, and we don't suddenly break the front page of any of our partners. 

I would say this is a minor version bump, since it should not be required to change anything around the widget where it is used (atb.no, frammr.no etc)

strandlie (changes requested):

public/widget/IYFwRkA/2.36.0/planner-web.css
jorelosorio: It might look like I made changes for the widget, but this pr depends on the colours, because figma for those new changes uses the new color names.

strandlie: I don't quite understand. Do you mean that when the colors-PR is merged these changes to the widget will disappear from this PR?

jorelosorio: We had a meeting and clarified this.

strandlie (approved):

PR #525, opened 2025-04-25
fix: Added unique ID, and removed unnecessary braces

strandlie:
Small missed fix for AtB-AS/kundevendt#19334


src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -88,8 +89,8 @@ export default function TripPattern({
   return (
     <div className={style.tripPatternContainer}>
       <motion.div
-        id="assistant-details-accordion"
-        role={'region'}
+        id={id}
```
mikaelbr, line 92: It is very common practise to use a named suffix (like `-details-region`) to make it easier to refer to and see it's intention.

Also need to refer to this from `aria-controls` on buttonProps for footer button.

src/translations/pages/assistant.ts
```diff
@@ -108,7 +108,7 @@ const AssistantInternal = {
         unknownPlace: _('ukjent', 'unknown', 'ukjend'),
       },
       quayPublicCodePrefix: _('', '', ''),
-      details: _('Gå til detaljer', 'Go to details', 'Gå til detaljar'),
+      details: _('Flere detaljer', 'More details', 'Fleire detaljar'),
```
adriansberg, line 111: Changing the wording here might make it less understandable (less context) for users that are using screen readers. Just a thought.

src/page-modules/assistant/trip/trip-pattern/trip-pattern.module.css
```diff
@@ -155,6 +154,17 @@
...
 
+.accordionBody {
+  padding: token('spacing.medium');
+}
+
+.accordionFooter {
+  display: flex;
+  justify-content: flex-end;
+  padding: token('spacing.small') token('spacing.medium');
+
```
adriansberg, line 165:
```suggestion
```

src/page-modules/assistant/trip/trip-pattern/index.tsx
```diff
@@ -263,19 +264,24 @@ export default function TripPattern({
...
+            <div className={style.accordionBody}>
+              <AssistantDetailsBody tripPattern={tripPattern} />
+            </div>
+            <div className={style.accordionFooter}>
+              <ButtonLink
+                href={`/assistant/${tripPattern.compressedQuery}?filter=${router.query.filter}`}
+                onClick={() => {
+                  setIsDetailsButtonClicked(true);
+                }}
+                state={isDetailsButtonClicked ? 'loading' : undefined}
```
adriansberg, line 276: Is this loading state strictly necessary? Should we instead try to reduce the loading time of the details page and just let the browser's native loading style indicate that something is loading?

src/translations/pages/assistant.ts
strandlie: Good point. Any thoughts about that @Mariblaas ?

src/page-modules/assistant/trip/trip-pattern/index.tsx
strandlie: Will fix. The footer-button is now moved [to here](https://github.com/AtB-AS/planner-web/pull/525/files#diff-6ed44a116e280ef09327293949e9e57fca748b834b065ad0f8cce4123c2e02e8L229) so that it does not move when clicked. But I see that `aria-controls` refers to the wrong ID there, so I will fix that as well.

strandlie: Fixed

src/page-modules/assistant/trip/trip-pattern/index.tsx
strandlie: With how the details page currently works I sometimes wondered if the button had actually been clicked. See video: 
https://github.com/user-attachments/assets/bd3a8e13-3c2d-4ca7-9a92-20272c273620

But yes, I agree that reducing the loading time of the details page is a better solution. I can remove this and create an issue for that.

adriansberg (approved):
Looking good! The textual changes doesn't need to be a blocker for this PR.

src/page-modules/assistant/trip/trip-pattern/index.tsx
strandlie: Yes I agree. It is removed now.

PR #526, opened 2025-04-25
update: reorganises the contact from validator

jonasbrunvoll:
Overall, this PR reorganises how the logic used to validate the input properties from the contact form. More specifically, the following happens:

- [x] The old set of rules used to validate the contact form properties is divided into separate sets of for each property (example `rulesFirstName `). This is done to achieve an better overview of the rules applying to each separate form property, and to make it easier to apply new rules.  
- [x]  Previously when validation, the validator looped trough all rules regardless of which values ​​were to be validated. Now,  the validator only lookup rules related to the provided form properties.


adriansberg (approved):
LGTM

PR #531, opened 2025-04-28
Widget in its more natural habitat

marius-at-atb:
This makes it possible to access a version of the widget a bit more "in context".
Made as part of becoming familiar with the widget ~~not sure if we should merge this?~~ To be merged due to [_popular demand_](https://github.com/AtB-AS/planner-web/pull/531#issuecomment-2835374418).
Going to `http://localhost:3000/widget/fullscreen` will default to the most recent version of the widget, or a specific version can be selected, such as e.g. `http://localhost:3000/widget/fullscreen/2.0.0`

The implementation is mostly based on the existing `src/pages/widget/index.tsx`.

Skjermbilde 2025-04-28 kl  15 59 57

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ NEXT_PUBLIC_PLANNER_ORG_ID=atb yarn build:widget
 
 This will place asset inside `public/widget/<VERSION>` which will be reachable
 through `http://localhost:3000/widget/` or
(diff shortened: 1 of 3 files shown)


reidzeibel (approved):
🚀

strandlie (approved):
Looking good!

PR #532, opened 2025-04-28
feat: improve HTML structure in contact form for better accessibility and semantics

jonasbrunvoll:
Replace SectionCard component with fieldset and legende to improve accessibility and semantics. Also, update to only display privacy and terms text in pages containing a form to submit.


reidzeibel:
Should we test this using VoiceOver or similar tools?

jonasbrunvoll:
> Should we test this using VoiceOver or similar tools?

I'm not an accessibility expert, but I tested the form using Apple's VoiceOver tool. When navigating into a fieldset, the text inside the <legend> tag is now announced,informing the user which section of the form they're in.  I don't believe additional VoiceOver testing is necessary

strandlie (approved):
The change to which pages has privacy seems a little strange to me, but I assume that is a request from someone :)

PR #534, opened 2025-04-30
feat: reduce maxSlack in viaTripQuery

mathiazom:
Unexpected behavior has been reported concerning the via search. Since the max slack has been set to 9 hours, a high number of suggestions will be returned for the same second departure (see image). The reason behind the 9 hours seems to be lost (and no significant use-case identified by AtB or FRAM), so reducing to 2 hours for now.

image


reidzeibel (approved):
Nice, also I noticed this is only for queries with `via`, which we don't have yet in the app.

tormoseng:
✅

PR #535, opened 2025-04-30
Use autocomplete instead of reverse

strandlie:
Closes AtB-AS/kundevendt#20033

A small change, but it is a breaking change for users of the widget since this requires `fromName` to be included in the query when the widget redirects, as seen in `widget.ts` here. The change have been coordinated with all the OMS-partners and will be implemented at roughly the same time. 

For this reason I also bump the major version.


reidzeibel (approved):
Looks great, hopefully this one works well!

PR #536, opened 2025-04-30
fix: crash in form validation due to unexpected input type

jonasbrunvoll:
This PR fixes a bug that caused the contact form to crash during validation.

The issue occurred because several form properties `transportMode` (`TransportModeType`), `line` (`Line`), `fromStop` and `toStop` (`Line['quays'][0]`), and `attachments` (`File[]´) —are not of type `string`. However, the validation function `isNotEmptyOrUndefined` was only designed to handle `string | undefined` types, leading to runtime errors when non-string values were passed.

To resolve this:

The original isNotEmptyOrUndefined function has been split into two type-safe functions:
- isDefined() – checks if any value is not undefined
- isNonEmptyString() – checks if a value is a non-empty string (ignoring whitespace)

This ensures proper validation without type errors and improves clarity by separating concerns.

### DIFF
--- a/src/page-modules/contact/validation/validationRules.ts
+++ b/src/page-modules/contact/validation/validationRules.ts
@@ -1,9 +1,15 @@
 import { PageText, TranslatedString } from '@atb/translations';
+import { Line } from '..';
+import { TransportModeType } from '../types';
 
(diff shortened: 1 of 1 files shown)


reidzeibel (approved):
🚀

PR #540, opened 2025-05-02
feat: add id property to select component

jonasbrunvoll:
Add id property to select component. This is done to include the select component in the automatic focus and scroll on submit, as described in https://github.com/AtB-AS/planner-web/pull/539.


strandlie (approved):
Very nice!

PR #554, opened 2025-05-08
feat(widget): trip search style changes and widget from-to a swap button implementation

Oleninkaa:
Closes https://github.com/AtB-AS/kundevendt/issues/19868

Skjermbilde 2025-05-08 kl  13 10 53

Widget:
- a swap button is implemented
- 'my posisjon' button is now in the toggle list

Skjermbilde 2025-05-08 kl  13 11 20

Note:
Changed a little search test because we 

> remove ", Trondheim" eg. after input (location names should still appear in the dropdown during input)


reidzeibel:
@strandlie have the OMS updated their widget yet? if not, this PR might worth checking before updating the widget.

src/components/search/search.tsx
```diff
@@ -292,7 +292,5 @@ function geocoderFeatureToString(
     // Location has been selected, but it hasn't been resolved to a feature yet
     return '';
   }
-  return feature
-    ? `${feature.name}${feature.locality ? ', ' + feature.locality : ''}`
-    : '';
+  return feature ? `${feature.name}` : '';
```
adriansberg, line 295:
```suggestion
  return feature?.name ?? '';
```

or 

```
  return feature ? feature.name : '';
```

src/components/search/search.module.css
```diff
@@ -2,21 +2,39 @@
...
+  border-bottom-left-radius: 0;
+}
+
+.container:not(:only-of-type):nth-child(2) {
+  border-top-right-radius: 0;
+  border-top-left-radius: 0;
+}
+
+.container:not(:only-child):first-child:focus-within .menu {
+  top: calc(100%);
```
adriansberg, line 31: Is `calc` needed here?

strandlie:
> @strandlie have the OMS updated their widget yet? if not, this PR might worth checking before updating the widget.

Yes, they have already updated their widget, but they will need to update again when the colors and changes are updated so they will get it in time

strandlie (approved):
Very good job :)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@atb-as/planner-web",
-  "version": "3.0.1",
+  "version": "3.0.2",
```
strandlie, line 3: I suggest bumping this to 3.1.0 (and then running `yarn generate-widget-version` again to get the correct version on the widget) since we are introducing swap for the widget now, and that is a new feature

PR #556, opened 2025-05-08
fix: rounded top border for destination input

mathiazom:
(This error was hidden by the new colors in `main` since the background and input where both white)

### Before

image

### After

image


jonasbrunvoll (approved):
🚀

PR #558, opened 2025-05-09
feat: update link to contact form

jonasbrunvoll:


mortennordseth (changes requested):
We should instead check if `process.env.NEXT_PUBLIC_CONTACT_API_URL` is defined. If it is, use `/contact`, if not, use the config url. See the function `shouldShowContactPage` in `src/page-modules/contact/utils.ts`

jonasbrunvoll:
> We should instead check if `process.env.NEXT_PUBLIC_CONTACT_API_URL` is defined. If it is, use `/contact`, if not, use the config url. See the function `shouldShowContactPage` in `src/page-modules/contact/utils.ts`

Makes sense. Updated now👍

mortennordseth (approved):

PR #565, opened 2025-05-12
refactor: use mapbox data for default coords

adriansberg:
Since the default latitude and longitude is a fairly static confiuration I think we can keep it as environment variables, and use those in the case we need the defaults instead of having configuration both as envirionment variables and in Firestore.

I have a PR ready to remove the configuration from Firestore if we decide that this is the way to go.

closes https://github.com/AtB-AS/kundevendt/issues/20583


strandlie:
Good solution! The only problem is that I don't think the envvars that the mapbox-data is based on is set in all environments (Fram, Svipper etc). Did you check that?

adriansberg:
@strandlie they are all defined in `.github/workflows/docker.yml`.

reidzeibel:
Would this make sense to be applied to the app as well or not? 🤔

adriansberg:
@reidzeibel I guess the geolocation is quite different on the app, based on permissions and what not. Not sure how it's done for the app either to be honest.

strandlie:
> @strandlie they are all defined in `.github/workflows/docker.yml`.

Would you look at that! Nice

PR #567, opened 2025-05-13
feat: Time Picker

marius-at-atb:
Resolves https://github.com/AtB-AS/kundevendt/issues/20322

The native time picker is not supported by Safari or Firefox on desktop.
This implements a custom time picker component which
- works everywhere
- has consistent style with the calendar picker right next to it

On touch devices, the size is 1.5x in order to make it easier to tap.

Desktop (chrome):

Touch (iOS Safari):

Note: the location of the popover will move to where there is space on the screen, so that it stays visible.


reidzeibel:
@marius-at-atb the time selector popup doesn't seem to be intuitive to use with screenreaders:

- You cannot go into the popup selection intuitively (unlike using combobox, for example)
- You cannot pick/change selection on the minutes (can't seem to navigate to the right part of the selector using screenreader)
- You cannot select the number once you have it highlighted.

marius-at-atb:
@reidzeibel Good catch!
Solved with [excludeFromTabOrder](https://github.com/AtB-AS/planner-web/pull/567/commits/e1df393e6ec746b7ccbc161f8799c5e05266427e), as the existing time input already supports arrow keys in that manner.

reidzeibel (approved):
Nice one for sure, but we have to fix this in the future if needed, feels wrong to skip it from tab order.

strandlie (approved):

PR #575, opened 2025-05-14
refactor: simplify trip query fetching logic and enhance location retrieval

adriansberg:
There was a bug in the logic of the feature retrieval based on the query parameters. The ID, e.g. `fromId`, in the query was never used. The name and coordinates where used to perform an autocomplete search, then the first result was always used. This resulted in cases where the wrong feature was used in, mainly, the from feature (as can be read in the issue details: https://github.com/AtB-AS/kundevendt/issues/20621). 

I have now refactored the logic to take the ID into account. The logic is now like this: 

1. If the name query parameter is present, perform an autocomplete search with only the name.
2. Based on the result from the autocomplete search: 
  a. If the id query parameter is present, find the correct feature based on it.
  b. If the latitude/longitude query parameters are present, find the correct feature based on them.
  c. If none of the above, take the first result.

Note: this does not solve any of the loading quirks like the loading indicator disappearing when the first search ends up empty.

closes https://github.com/AtB-AS/kundevendt/issues/20621

### DIFF
--- a/src/page-modules/assistant/trip-query-fetcher.ts
+++ b/src/page-modules/assistant/trip-query-fetcher.ts
@@ -12,92 +12,135 @@ export async function fetchFromToTripQuery(
   client: AssistantClient,
 ): Promise<FromToTripQuery> {
   const tripQuery = parseTripQuery(query);
-  const cursor = tripQuery?.cursor;
-  const transportModeFilter = parseFilterQuery(tripQuery?.filter);
-  const lineFilter = parseFilterQuery(tripQuery?.lineFilter);
+
+  if (!tripQuery) {
+    return {
+      from: null,
+      to: null,
+      via: null,
+      transportModeFilter: [],
+      searchTime: { mode: 'now' },
+      cursor: null,
+      lineFilter: [],
+    };
+  }
+
+  const transportModeFilter = parseFilterQuery(tripQuery.filter);
+  const lineFilter = parseFilterQuery(tripQuery.lineFilter);
   const searchTime = parseSearchTimeQuery(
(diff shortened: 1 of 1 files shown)


src/page-modules/assistant/trip-query-fetcher.ts
```diff
@@ -12,92 +12,146 @@ export async function fetchFromToTripQuery(
...
+  }
+
+  // If no match by ID, try to find by coordinates
+  if (hasCoordinates(tripQuery, locationType)) {
+    const latKey = `${locationType}Lat` as const;
+    const lonKey = `${locationType}Lon` as const;
+    feature = result.find(
+      (feature) =>
+        feature.geometry.coordinates[0] === tripQuery[lonKey] &&
+        feature.geometry.coordinates[1] === tripQuery[latKey],
```
strandlie, line 97: Here you look for an exact match of the coordinates, and if you don't find that just return the first result. Why not pick the closest result if you don't get an exact match?

strandlie (approved):
Amazing stuff 🎉

src/page-modules/assistant/trip-query-fetcher.ts
adriansberg: Good point, we can do that instead!

adriansberg: Implemented a simple number distance check. Didn't see the need for e.g. a Haversine formula to be used, not that crucial.

reidzeibel (approved):
Thank you! Didn't think of this when working on the feature!

PR #577, opened 2025-05-15
feat: Auto-move time to early morning on date change away from today

marius-at-atb:
It was requested by @Sebstorvik / customer service to auto-select an early time when selecting a different date than today.
This PR does that, however does not change it when you have already selected another date, because then you probably have already set it yourself..

Acceptance Criteria:
- [ ] When switching from today's date to a date in the future, the time should be reset to 06:00.
- [ ] The time should not be reset for any other date changes.
- [ ] The trip search should use the updated time on the next search


strandlie (approved):

tormoseng:
✅

PR #579, opened 2025-05-16
refactor: remove unused gql queries

adriansberg:
> [!NOTE]
> I changed it to use the via trip with details query, since that similar change was done for trip as well. This will more than likely be removed when we start using the BFF.

The main reason for changing this was that I noticed the via trip query did not have any limitations on how many trip patterns that should be fetched. That resulted in 100 (assume that's the default) trip patterns being fetched.

I also noticed that the `viaTrip` query is deprecated and the regular `trip` with `viaStop` should be used.


strandlie (approved):
Very nice! I think this was forgotten after the accordion was implemented.

PR #580, opened 2025-05-21
fix: don't revalidate trip pattern fetching

adriansberg:
```ts
        revalidateOnFocus: false,
        revalidateIfStale: false,
        revalidateOnReconnect: false,
```

Added these options to `useSWRInfinite` to _not_ automatically revalidate the data that has been fetched. Automatic revalidation might cause confusion for the user if it's not intended.


reidzeibel (approved):
Letsgo!

tormoseng:
✅  Looks good!

PR #581, opened 2025-05-21
fix: search filter new styles

Oleninkaa:
Closes https://github.com/AtB-AS/kundevendt/issues/20180 for **TPW**:
- new styles
- a different component is being used for the transportation icons
- '**AtB Bestill**' is not visible for OMS 

Notice that we:
- remove '**Other**' from the filter
- add '**AtB Bestill**' to the filter
Skjermbilde 2025-05-21 kl  14 54 58

Skjermbilde 2025-05-21 kl  14 55 23


src/modules/transport-mode/filter/filter.module.css
```diff
@@ -38,7 +52,8 @@
   display: flex;
   gap: token('spacing.small');
   align-items: center;
-  height: 1.75rem;
+  border: 1px solid token('color.background.neutral.3.background');
+  height: 2.7rem;
```
adriansberg, line 56: This makes the transport mode boxes to be 43.2px tall – in the design they look like they should be 36px. `2.25rem` might look like is closer to that.

src/modules/transport-mode/filter/filter.tsx
```diff
@@ -33,77 +34,88 @@ export default function TransportModeFilter({
...
-          );
-        })}
+              <span>{t(ComponentText.TransportModeFilter.all)}</span>
+            </label>
+          </div>
+        </li>
+
+        {data
+          .filter((option) =>
+            isFlexibleTransportEnabled
```
adriansberg, line 78: Aren't the filters controlled by what's in Firestore, which means that we shouldn't need to do a check here?

src/modules/transport-mode/filter/filter.module.css
```diff
@@ -19,16 +19,30 @@
...
-    flex-direction: row;
+  .filter {
+    grid-template-columns: repeat(2, 1fr);
+  }
+  .all label {
+    max-width: 100%;
   }
 }
 
+.transportMode div {
```
strandlie, line 39: This is probably a matter of style and preference, but I find this style of writing css a little hard to follow. Do we really want **all** children of an element with `transportMode`-class to have full width? What if we add more divs within transport mode? Seems like an receipe for hard-to-find bugs to me. Same with `.transportMode label` below. 

The style in the rest of the TPW is to create a css selector `transportModeContainer`, add `width: 100%` to that and add that to the div we are targeting. That pattern is also not perfect, but at least then we know that rule only applies to that specific div.

strandlie:
I see that this introduces AtB-bestill for TPW. But will AtB-bestill tripPatterns actually appear in the travel search? I think I heard something about there being done a lot of work in the app to enable that. Maybe @rosvik knows?

rosvik:
@strandlie Right, it's not something we support in TPW as fare as I'm aware. Try searching e.g. from [Steinkjer Stasjon to Frosta skole](https://reise.atb.no/?filter=bus%2Ctram%2Crail%2Cspeedboat%2Cferry%2Cflybus%2Cother&searchMode=now&fromId=NSR%3AStopPlace%3A714&fromName=Steinkjer+stasjon&fromLon=11.494591&fromLat=64.012677&fromLayer=venue&toId=NSR%3AStopPlace%3A44672&toName=Frosta+skole&toLon=10.737343&toLat=63.587328&toLayer=venue), and compare with the app. So we probably shouldn't show filter for it.

reidzeibel (approved):
Looks ok with all the review comment resolved and updated on the code

PR #582, opened 2025-05-21
refactor: use atb-bff for geocoder

adriansberg:
This PR replaces the http calls to Entur directly by calling the atb-bff. This is only for the geocoder events, i.e. autocomplete and reversing of position.

closes https://github.com/AtB-AS/kundevendt/issues/20682


src/widget/widget.ts
```diff
@@ -1020,7 +1020,7 @@ function createTripQueryForDeparture(
 }
 
 async function autocomplete(urlBase: string, q: string) {
-  const url = `${urlBase}api/departures/autocomplete?q=${q}`;
+  const url = `${urlBase}api/bff/autocomplete?q=${q}`;
```
adriansberg, line 1023: This is probably unfortunate for the widget. I don't know how we should handle this, @strandlie?

src/page-modules/bff/client/geocoder/index.ts
```diff
@@ -15,15 +16,18 @@ export function useAutocomplete(
...
-    : '';
-
-  const layerQuery = `&onlyStopPlaces=${onlyStopPlaces}`;
+  const queryObj = {
+    query: debouncedQuery,
+    lon: autocompleteFocusPoint?.geometry.coordinates[0],
+    lat: autocompleteFocusPoint?.geometry.coordinates[1],
+    layers: onlyStopPlaces ? ['venue'] : ['venue', 'address'],
+    limit: 10,
+    multiModal: 'parent', // TODO: This was set to 'child' previously
```
adriansberg, line 25: Since the app uses `parent` I would assume that's the correct one to use?

src/modules/api-server/index.ts
```diff
@@ -25,19 +25,41 @@ export {
   createWithExternalClientDecoratorForHttpHandlers,
 } from './external-client';
 
+// Helper types for composing client factories
```
strandlie, line 28: This brings up the question whether we want to have code that is complex to read and maintain but achieves the goal, or if that is a sign we should try to find another solution that is easier to understand. I tend to fall on the latter side. 

Any simpler ways to achieve the same thing?

src/page-modules/bff/client/geocoder/index.ts
strandlie: Maybe add that info to the comment

src/page-modules/bff/client/geocoder/index.ts
```diff
@@ -35,7 +39,7 @@ export function useAutocomplete(
 
 export async function reverse(coords: GeolocationCoordinates) {
   const result = await fetch(
-    `/api/departures/reverse?lat=${coords.latitude}&lon=${coords.longitude}`,
+    `/api//reverse?lat=${coords.latitude}&lon=${coords.longitude}`,
```
strandlie, line 42: Is the double slash intentional?

src/page-modules/bff/server/geocoder/index.ts
```diff
@@ -0,0 +1,79 @@
...
+): BffGeocoderApi {
+  return {
+    async autocomplete(query, options) {
+      const url = '/v1/geocoder/features';
+
+      const queryObject = {
+        query,
+        ...options,
+        limit: 10,
+        multiModal: 'parent', // TODO: child was used previously
```
strandlie, line 34: Same as the other multiModal

src/widget/widget.ts
strandlie: Ah, yes. We would either have to keep the URL to the endpoints the same, or keep both endpoints until all widgets have been upgraded to this version. 

I don't really see why we would want to change away from `departures`?

adriansberg: We probably don't. I'll move it back.

src/modules/api-server/index.ts
adriansberg: > Any simpler ways to achieve the same thing?

I reverted these changes since we ended up needing the flexibility. I think the original code is manageable to understand, and that we can keep it like that for now.

strandlie (approved):

PR #583, opened 2025-05-22
fix: use same search config values as the app

adriansberg:
The app uses 5 for the amount of search attempts and 8 for the wanted number of trip patterns to show. It also uses 8 as the amount it wants when loading more, but that is not included as of yet in TPW.

## 📋 Acceptance criteria

- [ ] Trip search behaves more like the app


tormoseng:
✅  Looks good. Checked both max 5 requests and 8 trip patterns.

PR #584, opened 2025-05-23
refactor: use immutable swr hook

adriansberg:
And also clean up unused imports.

`useSWRImmutable` is the equivalent of using `useSWR` and passing the revalidate options with all of them set to `false`.

## 📋 Acceptance criteria

- [ ] No API calls are performed when refocusing (e.g. switching tabs) TPW

### DIFF
--- a/src/page-modules/assistant/client/journey-planner/index.ts
+++ b/src/page-modules/assistant/client/journey-planner/index.ts
@@ -1,4 +1,3 @@
-import useSWR from 'swr';
 import {
   NonTransitTripData,
   TripQuery,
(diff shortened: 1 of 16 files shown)


src/page-modules/assistant/client/journey-planner/index.ts
```diff
@@ -99,11 +99,11 @@ export function useTripPatterns(
...
-  const { data, error, isLoading } = useSWR<NonTransitTripApiReturnType>(
-    `/api/assistant/non-transit-trip?${queryString}`,
-    swrFetcher,
-    {},
-  );
+  const { data, error, isLoading } =
+    useSWRImmutable<NonTransitTripApiReturnType>(
+      `/api/assistant/non-transit-trip?${queryString}`,
+      swrFetcher,
+    );
```
strandlie, line 106: You say that it is equivalent to passing all options with false, but is that the same as passing an empty object, which seems like it was done previously?

strandlie (approved):

src/page-modules/assistant/client/journey-planner/index.ts
adriansberg: Passing an empty options object is essentially the same as not passing anything. The options themselves were not passed.

tormoseng:
✅

PR #603, opened 2025-08-01
fix: Retain via and line number when navigating back

strandlie:
Closes https://github.com/AtB-AS/kundevendt/issues/20239

### DIFF
--- a/src/page-modules/assistant/details-body/index.tsx
+++ b/src/page-modules/assistant/details-body/index.tsx
@@ -17,8 +17,6 @@ import {
   ExtendedLegType,
   ExtendedTripPatternWithDetailsType,
 } from '@atb/page-modules/assistant';
-import { tripQueryStringToQueryParams } from '@atb/page-modules/assistant/details/utils.ts';
-import { useRouter } from 'next/router';
 
 type DetailsBodyProps = {
   tripPattern: ExtendedTripPatternWithDetailsType;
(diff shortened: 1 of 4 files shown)


src/page-modules/assistant/details-body/index.tsx
```diff
@@ -36,14 +35,6 @@ export function AssistantDetailsBody({ tripPattern }: DetailsBodyProps) {
     language,
   );
 
-  const tripSearchParams = router.query.id
```
strandlie, line 39: This did not do anything. Just wasn't marked as unused, since it used itself in a cycle. Not used outside of this deleted block.

src/page-modules/assistant/details/utils.ts
```diff
@@ -34,54 +32,3 @@ export function formatLineName(
   const name = frontText ?? lineName ?? '';
   return publicCode ? `${publicCode} ${name}` : name;
 }
-
-export function tripQueryStringToQueryParams(
```
strandlie, line 38: Deleting this means that the original searchTime is no longer retained when searching `now`, but it just stays as `now`. That means some tripPatterns may disappear when navigating back, but is that really a problem? I found it confusing that it suddenly had a time, and also this meant different behavior between the in-page back button and the browser back button

strandlie: Is that really an issue though? If you search now and share the link that link is really only relevant **right now**, and if you search in the future then the `searchTime` is included

PR #619, opened 2025-08-14
chore: upgrades dependencies (2025-08-14)

mikaelbr:
I have done a manual regression test for FRAM. Don't expect there to be much difference between different PTAs. Doesn't seem to be any issues introduced. Most changes are on reusable components in regards to new ways to handle refs in React 19, and some implementation details on hooks, which both has cross cutting concerns, meaning I think issues with them would present them selfs pretty quickly.

Major changes:
- General typescript strictness
- New ref use (react 19)
- Lint strictness causing issue with missing hook dependency and for some reason a hook was marked as async.
- Next 15 changes in cookies API (async)


strandlie (approved):
Tested OK for me as well!

PR #627, opened 2025-08-22
docs: Added info about release to README

strandlie:


README.md
```diff
@@ -38,6 +38,44 @@ yarn dev
...
+````
+
+
+## Changing organization
+To easily change organization, you can the following command. It assumes the new  `NEXT_PUBLIC_PLANNER_ORG_ID` 
+is set in your terminal environment as explained above.
+```bash
+# Clean assets from old org and create new ones
+NEXT_PUBLIC_PLANNER_ORG_ID=<orgID> yarn clean:install
+```
```
rosvik, line 55: I find this a bit confusing 🤔 

I guess the `set -a && source .env.local && set +a` will load all env variables to the current bash/zsh shell. Are any other than the `NEXT_PUBLIC_PLANNER_ORG_ID` variable matter for the yarn scripts? If not, maybe running `export NEXT_PUBLIC_PLANNER_ORG_ID=<orgID>` is easier.

> It assumes the new  `NEXT_PUBLIC_PLANNER_ORG_ID` is set in your terminal environment as explained above.

Then I guess we shouldn't include `NEXT_PUBLIC_PLANNER_ORG_ID=<orgID>` in the following command.

---

Or can we just remove the need for setting `NEXT_PUBLIC_PLANNER_ORG_ID` manually? And always read from .env.local? 👀

PR #634, opened 2025-09-05
fix: prevent duplicate departures when loading more in departures

rosvik:
The `set.has()` check used when loading more departures was unreliable, and caused the issue described in https://github.com/AtB-AS/kundevendt/issues/21567. Better to check against a custom key that should define unique estimated calls.

closes https://github.com/AtB-AS/kundevendt/issues/21567


adriansberg (approved):
Makes sense :shipit:

PR #704, opened 2026-03-26
fix: round trip details arrival times up

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/23035

Two changes here in essence + some updated tests:

- Reverts this change: https://github.com/AtB-AS/planner-web/pull/600
- Rounds arrival times up in TripSection to match trip details


reidzeibel (approved):
Yay!

PR #705, opened 2026-03-27
feat: add missing box-shadow for accessibility

reidzeibel:
related to https://github.com/AtB-AS/kundevendt/issues/22865

fulfills WCAG rule 2.4.7 : Synlig fokus by adding box-shadow on the focused items when navigating with keyboard. Most of the buttons already has it, just need to add the same box-shadow on button-primary, button-secondary, and search

### DIFF
--- a/src/components/button/button.module.css
+++ b/src/components/button/button.module.css
@@ -92,6 +92,11 @@
 .button--primary.button--disabled {
   opacity: 0.2;
 }
+.button--primary:focus {
+  outline: 0;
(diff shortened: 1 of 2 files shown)


rosvik (approved):
Nice! Should this (and other a11y improvements you've made) be copied over to the webshop as well?

reidzeibel:
@rosvik Soon™️

Will do a separate lookup for webshop, I think there are a lot more components/forms in the webshop that it isn't as simple as it is in TPW 👍🏼

PR #707, opened 2026-03-27
fix: add missing alt on icons, add fallback alt on page-header

reidzeibel:
related to https://github.com/AtB-AS/kundevendt/issues/22865

Fulfills WCAG 1.1.1 so now there are no `` without `alt`, also adds fallback `alt` to the logo in case `fylkeskommune` or `fylkeskommune.name` is undefined.

### DIFF
--- a/src/components/icon/color-icon.tsx
+++ b/src/components/icon/color-icon.tsx
@@ -26,7 +26,7 @@ export function ColorIcon({ icon, size = 'normal', ...props }: ColorIconProps) {
   if (!currentIcon.darkable) {
     const totalPath = `/assets/${currentIcon.relative}`;
     // eslint-disable-next-line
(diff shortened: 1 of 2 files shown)


rosvik (approved):
I wasn't sure if this was correct, so looked it up. Adding it here for context:

> However, a decorative image that isn't discussed by the surrounding text but still has some relevance can be included in a page using the [img](https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element) element. Such images are decorative, but still form part of the content. In these cases, the [alt](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-alt) attribute must be present but its value must be the empty string.

https://html.spec.whatwg.org/multipage/images.html#a-purely-decorative-image-that-doesn't-add-any-information

PR #715, opened 2026-04-28
chore: update default transfer slack FRAM to 2 minutes (120 seconds)

reidzeibel:
Fix oopsie from https://github.com/AtB-AS/planner-web/pull/714


rosvik (approved):
😅

PR #718, opened 2026-06-03
feat: add date pagination component to departures

rosvik:
Adds a date-pagination to departures that works similarly to the app. This also meant I had to update how the departures list is loaded, and not load departures for the next day unless start time is changed.

https://github.com/user-attachments/assets/ba0c357a-d2a6-454a-8632-47b54a634554

closes https://github.com/AtB-AS/kundevendt/issues/23884


reidzeibel (approved):
🚀

PR #721, opened 2026-06-03
feat: expose tariffZones for `quay` object in the GraphQL query

reidzeibel:
Tariff zones will be useful in the future when we want to show what ticket zones the user requires.

Also helps with the reisesøk til billetkjøp project


reidzeibel:
@rosvik sure 👍🏻

PR #726, opened 2026-06-08
fix: refresh now-time stamp on each travel search navigation

gorandalum:


reidzeibel (approved):
I believe I had this bug earlier today, nice!

PR #729, opened 2026-06-09
fix: Transport mode filter centering

gorandalum:
Fixes this comment:
https://github.com/AtB-AS/planner-web/pull/724#issuecomment-4629110581

This was fixed by just adding top margins to icon and text. I tried to look into a more dynamic fix which automatically centered in relation to the transport icon height, but the resulting code changes was quite complex and not worth the effort.

**Before/After**


reidzeibel (approved):
Simple fix 👍🏼

PR #737, opened 2026-06-11
feat: show affected stop places in situation message box

gorandalum:
Lists the affected stops in the situation message modal, deduplicated
and truncated to "8 + N other stops" when there are more than 10.


rosvik (approved):
LGTM 🚀

